PR #10, opened 2017-04-06
[SecurityBundle] Proposed a default config for test environment

javiereguiluz:
Just a proposal about something that people usually have problems with.


javiereguiluz:
Closing in favor of the duplicated, but more modern, #131 issue.

PR #13, opened 2017-04-07
Moved the "dev" security config to dev/security.yaml

javiereguiluz:
This is not a real PR. This is just to ask your opinion about this:

1. I don't like the `firewalls.dev` config in the security.yaml used in production.
2. We can't move the entire "dev" config to another file because of this limitation: https://github.com/symfony/symfony/issues/22308
3. So the only solution is to define a `dev: null` in prod and add the rest of config in dev.

But it's a bit ugly, isn't it?


javiereguiluz:
Closing because this is not a solution but a "hack" to solve something that the Security component doesn't want us to solve.

PR #30, opened 2017-04-21
Recommend specifying the controller itself for EasyAdmin

Pierstoval:
I think it's a better idea to expose the controller itself directly.

It would encourage people to add their own controller by just changing the name of the bundle if they want to override it.

Also, I would suggest to change `master` for a lower requirement (1.14 or 1.15 maybe, there are some without much BC breaks, and they'd still share the same recipe) and add `version_aliases`, but haven't done it in the PR because `master` might have been added for a specific reason 🙂

### DIFF
--- a/javiereguiluz/easyadmin-bundle/master/etc/routing/easy_admin.yaml
+++ b/javiereguiluz/easyadmin-bundle/master/etc/routing/easy_admin.yaml
@@ -1,4 +1,4 @@
 easy_admin_bundle:
-    resource: "@EasyAdminBundle/Controller"
+    resource: "@EasyAdminBundle/Controller/AdminController.php"
     prefix: /admin
     type: annotation


fabpot:
pings @javiereguiluz

javiereguiluz:
This proposal makes sense to me! I've updated the official docs accordingly: https://github.com/javiereguiluz/EasyAdminBundle/pull/1604  Thanks @Pierstoval.

Regarding the question about the versions, I don't really know yet what to do. I'd say, let's keep it as is for now.

fabpot:
@javiereguiluz Thanks (check your emails, I need you :))

fabpot:
Thank you @Pierstoval.

Pierstoval:
> Regarding the question about the versions, I don't really know yet what to do. I'd say, let's keep it as is for now.

@fabpot is there anything regarding the "stability" of a Flex recipe like there is for Composer packages stability? (stable, beta, dev, etc.)
If yes, then we should put EasyAdmin under its latest version to be 100% sure, and remove master.
If no, then, whatever works 😄

fabpot:
There is no stability for a recipe, it can only be stable (the underlying package can have a stability, but not recipes). Also, keep in mind that it works only for deps > Symfony 3.3. So, we can only support bundles that work with Symfony 3.3+.

Pierstoval:
Ok, so EasyAdmin will have to maintain a new branch for Symfony `^3.3|^4.0`, as well as almost all other bundles 😕

stof:
@Pierstoval you can have a branch supporting 2.8+, as long as 3.3 works.
what is bad is a bundle being incompatible with Symfony 3.3 (as the recipe will be useless for people using 3.2 and older anyway and they will need to use manual configuration)

Pierstoval:
Yep, this can be checked via composer version constraints so this is no issue actually

PR #31, opened 2017-04-21
Fix PHPUnit xsd version

Pierstoval:
The base supported version is ~4.7~ 6.1 as specified in the ~directory~ version aliases


sstok:
Note that PHPUnit doesn't actually use this uri for validating, it's mainly for IDE auto suggest 😛

Pierstoval:
Yep, but still, _consistency_, you know 😜

fabpot:
What about doing it the other way and bump the version to the max in the XSD? That way, you have autocompletion for everything supported.

Pierstoval:
You're right @fabpot, updated the PR

fabpot:
Thank you @Pierstoval.

PR #32, opened 2017-04-27
Add LexikJWTAuthenticationBundle

chalasr:
Alternative to https://github.com/symfony/recipes-contrib/pull/5, I think it would be more appropriated here as this bundle has no competitor regarding the need it solves. 
Repository: https://github.com/lexik/LexikJWTAuthenticationBundle

### DIFF
--- a/lexik/jwt-authentication-bundle/2.3/Makefile
+++ b/lexik/jwt-authentication-bundle/2.3/Makefile
@@ -1,0 +1,8 @@
+generate-jwt-keys:
+ifeq (, $(shell which openssl))
+$(error "Unable to generate keys (needs OpenSSL)")
+endif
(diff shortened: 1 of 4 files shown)


lexik/jwt-authentication-bundle/2.3/Makefile
```diff
@@ -0,0 +1,5 @@
+generate-rsa-keys:
+		mkdir -p etc/jwt
```
sstok: Don't what the best practice is, but one `\t` should be enough?

chalasr: indeed, fixed

fabpot:
Thank you @chalasr.

PR #34, opened 2017-04-27
Readme tweaks

fabpot:


README.rst
```diff
@@ -37,6 +35,12 @@ The following example shows the real directory structure of some Symfony recipes
 All the ``manifest.json`` file contents are optional and they are divided into
 options and configurators.
 
+.. note::
+
+    Don't create a recipe for Symfony bundles if the only configuration in the
+    manifest is the registration of the bundle for all environments, as this is
+    done automatically.
```
stof: you should mention that this is done automatically *when there is no recipe*.

If there is a recipe to configure extra things, it must still register the bundle

fabpot: That would make the note heavy. I think people will realize that.

PR #36, opened 2017-04-27
Add current validation performed by the server

fabpot:


README.rst
```diff
@@ -279,6 +279,25 @@ colors`_ are supported too:
...
+
+* YAML files suffix must be `.yaml`, not `.yml`;
+* YAML files must be valid;
+* JSON files must be valid;
+* Aliases are only supported in the main repository, not the contrib one;
+* Aliases must not be already defined by another package;
+* The versions are valid;
+* Version defined in the directory must be the oldest amongst those defined in `version_aliases`;
+* The package must exist on Packagist;
+* The package must have at least one version on Packagist;
```
stof: do you mean `1 stable version`, `1 released version` (i.e. counting alpha, beta and RC versions) or are you also counting dev branches ?
If dev versions are counted, this is almost impossible to fail this validation.

README.rst
```diff
@@ -279,6 +279,25 @@ colors`_ are supported too:
...
+
+When submitting a recipe, several checks are automatically executed to validate
+the recipe:
+
+* YAML files suffix must be `.yaml`, not `.yml`;
+* YAML files must be valid;
+* JSON files must be valid;
+* Aliases are only supported in the main repository, not the contrib one;
+* Aliases must not be already defined by another package;
+* The versions are valid;
```
javiereguiluz: "The versions are valid" -> "All the versions listed in the manifest exist as package versions on Packagist" ?

javiereguiluz:
Proposal for another check: the GitHub repo has "symfony", "bundle" and "symfony-bundle" topics, as official recommended (https://symfony.com/blog/standardizing-the-github-topics-for-symfony-repositories).

javiereguiluz:
Another proposal: XLIFF files must be valid (it's easy to make mistakes and break everything ... I did that recently on the Symfony Demo app: https://github.com/symfony/symfony-demo/pull/549)

javiereguiluz:
Another proposal: service XML files must be valid (we have an XSD for them after all).

stof:
I don't think adding XLIFF files in the recipe makes much sense, so I'm not sure this makes sense to validate them.

And if you talk about validating the content of the bundle itself, this would be wrong: it would not forbid you to break your own bundle (as there is nothing triggering a recipe PR in this case), but then your own broken bundle would forbid other people to submit their own recipes (as the recipe of your own bundle, which is already merged, would then become invalid for rules validating content in another repo).

javiereguiluz:
Another proposal: bundles must be compatible with Symfony.

Not simple but the idea is: Symfony Flex server creates an empty Symfony project and then installs the bundle. The check is to run "bin/console --version" and verify that no errors happen. This check is not perfect, but it rules out a lot of incompatibilities.

javiereguiluz:
@stof yes, I was suggesting to check some of the bundle contents. Using Flex to install a bundle and find that your app is broken is not a good experience (of course it's not caused by Flex ... but if we can use Flex to detect those issues...)

README.rst
fabpot: I know, but as I'm making the check anyway, I documented here.

README.rst
fabpot: Just that they are valid, updating it

fabpot:
@javiereguiluz I think those kind of checks are interesting but should be part of a checker that people can install on their bundle repository directly.

Pierstoval:
> @javiereguiluz I think those kind of checks are interesting but should be part of a checker that people can install on their bundle repository directly.

Nothing prevents symfony contributors to create such checker and then integrate it into Flex validation server 😉

stof:
@Pierstoval it should not be in the flex validation server, as it would have to run when pushing in the bundle repo, not in the flex recipes repo

PR #37, opened 2017-04-28
Document the special %generate(secret)% env var value

javiereguiluz:
Some recipes are starting to use this feature (see https://github.com/symfony/recipes-contrib/pull/5/files) so let's document it.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -203,8 +203,8 @@ Don't remove or modify these separators.
 
 .. tip::
 
-    Use ``%generate(secret)%`` as the value of any environment variable to ask
-    Symfony to replace it with a cryptographically secure random value of 16 bytes.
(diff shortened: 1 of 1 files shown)


Pierstoval:
It is a Flex feature or a Symfony feature?

stof:
it is a Flex feature. It will replace it by a random value when editing the .env file

javiereguiluz:
@Pierstoval in #34 Fabien removed most references to the "Flex" name, so I thought it was a good idea to not mention it. But I've reworded to not mention neither Symfony nor Flex.

Pierstoval:
Yeah, because there's a difference between Symfony (the framework) and Symfony Flex (the package auto-configurator), and the server behind all of this that does the validation 😆 Lots of new cool stuff!

fabpot:
Thank you @javiereguiluz.

PR #38, opened 2017-04-28
[doctrine-bundle] Change the default DSN

dunglas:
| Q | A |
| --- | --- |
| License | MIT |

* Use a non-root user and a password to ease the integration with the official Docker MySQL image
* Change the default database name to `app` to be consistent with the name of new environment variables.

### DIFF
--- a/doctrine/doctrine-bundle/1.6/manifest.json
+++ b/doctrine/doctrine-bundle/1.6/manifest.json
@@ -10,6 +10,6 @@
         "#1": "Format described at http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url",
         "#2": "For a sqlite database, use: \"sqlite:///%kernel.project_dir%/var/data.db\"",
         "#3": "Set \"serverVersion\" to your server version to avoid edge-case exceptions and extra database calls",
-        "DATABASE_URL": "\"mysql://root@127.0.0.1:3306/symfony?charset=utf8mb4&serverVersion=5.7\""
+        "DATABASE_URL": "\"mysql://db_user:db_password@127.0.0.1:3306/db_name?charset=utf8mb4&serverVersion=5.7\""
     }
 }


doctrine/doctrine-bundle/1.6/manifest.json
```diff
@@ -10,6 +10,6 @@
         "#1": "Format described at http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url",
         "#2": "For a sqlite database, use: \"sqlite:///%kernel.project_dir%/var/data.db\"",
         "#3": "Set \"serverVersion\" to your server version to avoid edge-case exceptions and extra database calls",
-        "DATABASE_URL": "mysql://root@127.0.0.1:3306/symfony?charset=utf8mb4&serverVersion=5.7"
+        "DATABASE_URL": "mysql://app:app@127.0.0.1:3306/app?charset=utf8mb4&serverVersion=5.7"
```
fabpot: This is a default for dev env. Don't you think that "most" people are using MySQL locally where the username is root with no password?

javiereguiluz: I agree. By the way, if everything is `app`, how can I know which one is the user, the password and the db?

dunglas:
@fabpot I think most teams tend to use Docker or similar solutions (Vagrant, Homestead...) for dev environment.

@javiereguiluz it's a typical URL, it is a basic developper knowledge ;)

doctrine/doctrine-bundle/1.6/manifest.json
Pierstoval: I also think it's a good practice to propose users a non-root user name and a non-empty password. It encourages best practices at first, and next, it helps when dealing with PaaS where environment can define very long & complex database names automatically (like JawsDB heroku add-on for example).

Moreover, it explicitly adds verbosity to the full database url that will be parsed by the DBAL.

I'd go further by replacing it with more verbose names:

`mysql://db_user:db_password@127.0.0.1:3306/app?charset=utf8mb4&serverVersion=5.7`

Pierstoval:
> @fabpot I think most teams tend to use Docker or similar solutions (Vagrant, Homestead...) for dev environment.

Do you have any statistics about this? I personally know tons of developers that don't use docker at all, could it be for dev or prod.

doctrine/doctrine-bundle/1.6/manifest.json
fabpot: Using explicit placeholder as shown by @Pierstoval would be better than using app three times.

dkarlovi:
I think this change might be motivated by #64. :)

dkarlovi:
_Asks_

> Do you have any statistics about this?

_Uses_

> tons of developers

_as a counter-metric._

NHF @Pierstoval :)

doctrine/doctrine-bundle/1.6/manifest.json
teohhanhui: Can I propose that we rename `DATABASE_URL` to `DATABASE_DSN`?

http://php.net/manual/en/pdo.construct.php
http://php.net/manual/en/ref.pdo-mysql.connection.php
http://php.net/manual/en/ref.pdo-pgsql.connection.php
http://php.net/manual/en/ref.pdo-sqlite.connection.php

teohhanhui: But I'm not sure what's the benefit of using a DSN, when we could just pass these separately. (I'm sure there are some cases where it's necessary / desirable, but do we really consider this a best practice?)

See https://code.djangoproject.com/ticket/1810

teohhanhui: A real-world example:

`pgsql://%env(DB_USER)%:%env(DB_PASSWORD)%@%env(DB_HOST)%:%env(DB_PORT)%/%env(DB_DBNAME)%?charset=UTF8&serverVersion=9.6`

While I appreciate that using a URL / DSN provides the greatest flexibility, it's actually not useful in most cases because as the comment at the top of `etc/container.yaml` rightly points out:

```yaml
# Put parameters here that don't need to change on each machine where the app is deployed
# https://symfony.com/doc/current/best_practices/configuration.html#application-related-configuration
```

A typical user wouldn't use e.g. a different type and version of DBMS in different deployment environments.

(For testing, this is already covered by setting `APP_ENV=test` and overriding in `etc/packages/test/doctrine.yaml`.)

dunglas: @teohhanhui having a variable called `DATABASE_URL` and containing a DSN is a fairly common convention popularized by Rails.

It's used by:

* Rails (of course): http://edgeguides.rubyonrails.org/configuring.html#configuring-a-database
* Django Environ: https://django-environ.readthedocs.io/en/latest/#how-to-use
* Heroku: https://devcenter.heroku.com/articles/heroku-postgresql#create-a-new-database
* Cloud Foundry: https://docs.cloudfoundry.org/buildpacks/ruby/ruby-service-bindings.html#configuring-non-rails-applications

I'm to keep this as is.

dunglas:
I switched to placeholders proposed by @Pierstoval. I'm not very happy with it (I would like to not have to edit it when running a Docker environment).

teohhanhui:
Set the environment variable from the `ENTRYPOINT` script.

dunglas:
What do you think about `myql://symfony:symfony@127.0.0.1:3306/app` instead? It's explicit and cleaner.

Pierstoval:
> What do you think about `myql://symfony:symfony@127.0.0.1:3306/app` instead? It's explicit and cleaner.

I'm fund of explicit names like `db_name` and `db_user`, which seem (to me) much clearer than `symfony:symfony`. For newcomers, it's not explicit that `symfony:symfony` corresponds to `user:password` at all, and that `app` is the `database_name`.

> Asks
> 
> > Do you have any statistics about this?
> 
> Uses
> > tons of developers
> 
> as a counter-metric.
> 
> NHF @Pierstoval :)

Yeah, rumor versus anecdote, I think I completely failed my epistemic pitch, I confess 🤣 Maybe if I quote them it can add value? (yet it adds another failure: testimony 😆)

But still, I'd like to have statistics about docker usage in dev and/or prod, because a poll with 250 participants on Twitter is not representative of the thousands of PHP developers that may be interested with this feature :)

dkarlovi:
> I'd like to have statistics about docker usage in dev and/or prod

I don't think that'd be possible. Docker encourages (at least it did) to use private image registries for any non-OSS work and Gitlab ships with that by default for quite some time, so bulk of usage would be out of sight.

What I do know is people using Docker in dev because it's just vastly faster. For example, working with Sylius or any other heavier stack usually means you've up to 5x faster on Docker (on Linux, mind you, don't know about other platforms) with all debugging enabled, etc. I'm working on an API Platform based app and all my requests are within 150ms (with debug), you can't get anywhere near that on Vagrant (in say, VirtualBox). It's quite the productivity boost.

fabpot:
Done in ba3a9be

PR #39, opened 2017-04-28
Add DoctrineCacheBundle (non-standard bundle name)

dunglas:
The bundle class name doesn't follow Symfony standards, so it cannot be autodetected.

### DIFF
--- a/doctrine/doctrine-cache-bundle/1.0/manifest.json
+++ b/doctrine/doctrine-cache-bundle/1.0/manifest.json
@@ -1,0 +1,5 @@
+{
+    "bundles": {
+        "Doctrine\\Bundle\\DoctrineCacheBundle\\DoctrineCacheBundle": ["all"]
+    }
+}


Pierstoval:
Why do you say its name is not standard? 🤔

stof:
the magic detection would look for `DoctrineDoctrineCacheBundle`

Pierstoval:
Oh, then non-standard bundle names should be detectable in other ways.

We have PSR namespaces + access to the package itself, so it should be possible to detect a bundle even in a heavy way (I mean search bundle file/class in the repo) to make sure non-standard names are detected.

dunglas:
The Doctrine bundle looks valid according to our current best practice: http://symfony.com/doc/current/bundles/best_practices.html

I'll patch the detection code to at least support both conventions.

stof:
@Pierstoval I already suggested supporting the case where the bundle class name matches the last part of the segment only. Currently, Flex supports this only for the Symfony namespace (and does not allow prepending it). We can still improve it

dunglas:
I was wrong, this name isn't valid (it should be `Doctrine\Bundle\CacheBundle\DoctrineCacheBundle`.

fabpot:
fixed in fabpot/flex@d4260d1

Pierstoval:
Awesome! 👍

PR #45, opened 2017-05-03
Fixed useless calls

HeahDude:


fabpot:
Good catch, thanks @HeahDude.

Issue #49, opened 2017-05-03
[framework-bundle] Don't advice to install web-server-bundle when already installed


sstok:
I installed the web-server-bundle but I'm still getting:

```
$ make serve
Server listening on http://127.0.0.1:8000
Quit the server with CTRL-C.
Run composer require symfony/web-server-bundle for a better web server
php -S 127.0.0.1:8000 -t web
PHP 7.1.3 Development Server started at Wed May  3 14:25:01 2017
Listening on http://127.0.0.1:8000
```

It seems I need to run `bin/console server:start` instead but this is not explained.
Also no post-install is shown to tell me I need to use something else?

stof:
@fabpot looks like the symfony/web-server-bundle manifest does not overwrite `make serve`

sstok:
I don't think it's possible to overwrite this 🤔 but it should be possible to detect if this command is installed and then show a different message.

fabpot:
PR welcomed :) I think `make serve` should then run `server:run`.

sstok:
What's the best way to detect if this command is installed?
server:status gives error code 1 when the server is not running, but when the command is not installed it also gives this error 🤔

fabpot:
`./bin/console list server | grep server:run`?

sstok:
Ah yes 😋

sstok:
OK, Makefiles are something I really need to learn 😑

This doesn't work, and the error `/bin/sh: -c: line 1: syntax error: unexpected end of file` is anything but helpful.

```bash
serve:
	@if [ -f bin/console ] && [ bin/console list server | grep server:run ]; then
		bin/console server:start
		@echo "Quit the server with bin/console server:stop."
	@else
		@echo "\033[32;49mServer listening on http://127.0.0.1:8000\033[39m"
		@echo "Quit the server with CTRL-C."
		@echo "Run \033[32mcomposer require symfony/web-server-bundle\033[39m for a better web server"
		php -S 127.0.0.1:8000 -t web
	@if
```

sstok:
OK, I'm starting to loose my nerve here... 💀 Make is doing a great at make**ing** me become really angry, I give-up!! WHY CAN'T YOU DEFINE A SINGLE IF STATEMENT?!!

stof:
I'll try to look at this, as I want to play with the Makefile to solve #50 anyway.

fabpot:
fixed a while ago

PR #54, opened 2017-05-04
[FrameworkBundle] improve Makefile serve target

sstok:
Related to #49, I don't know if this the best way to do it. But as the Makefile is composed I don't think it's possible to use advanced Makefile features 😟

@greg0ire suggested something like this: 

```makefile
.PHONY: serve

CONSOLE=bin/console

$(CONSOLE):
    @echo 'Woops! The console does not exist'
    @exit

serve: bin/console
    bin/console list server|grep server:start
    bin/console server:start
```

But don't know if you can have multiple targets with the same name but different conditions?

### DIFF
--- a/symfony/framework-bundle/3.3/Makefile
+++ b/symfony/framework-bundle/3.3/Makefile
@@ -6,9 +6,23 @@ cache-warmup: cache-clear
 	@test -f bin/console && bin/console cache:warmup || echo "cannot warmup the cache (needs symfony/console)"
 .PHONY: cache-warmup
 
-serve:
-	@echo -e "\033[32;49mServer listening on http://127.0.0.1:8000\033[39m"
-	@echo "Quit the server with CTRL-C."
-	@echo -e "Run \033[32mcomposer require symfony/web-server-bundle\033[39m for a better web server"
+CONSOLE=bin/console
+sf_console:
+	@test -f $(CONSOLE) || printf "Run \033[32mcomposer require cli\033[39m to install the Symfony console.\n"
+	@exit
+
+serve_as_sf: sf_console
+	@test -f $(CONSOLE) && $(CONSOLE)|grep server:start > /dev/null || ${MAKE} serve_as_php
+	@$(CONSOLE) server:start || exit 0
+
+	@printf "Quit the server with \033[32;49mbin/console server:stop.\033[39m\n"
+
+serve_as_php:
+	@printf "\033[32;49mServer listening on http://127.0.0.1:8000\033[39m\n";
+	@printf "Quit the server with CTRL-C.\n"
+	@printf "Run \033[32mcomposer require symfony/web-server-bundle\033[39m for a better web server\n"
 	php -S 127.0.0.1:8000 -t web
-.PHONY: serve
+
+serve:
+	@${MAKE} serve_as_sf
+.PHONY: sf_console serve serve_as_sf serve_as_php


symfony/framework-bundle/3.3/Makefile
```diff
@@ -7,8 +7,13 @@ cache-warmup: cache-clear
 .PHONY: cache-warmup
 
 serve:
-	@echo -e "\033[32;49mServer listening on http://127.0.0.1:8000\033[39m"
-	@echo "Quit the server with CTRL-C."
-	@echo -e "Run \033[32mcomposer require symfony/web-server-bundle\033[39m for a better web server"
-	php -S 127.0.0.1:8000 -t web
+	@if [ -a bin/console ] && [[ `bin/console list server | grep server:start` ]]; then \
+		bin/console server:start; \
+		echo -e "Quit the server with \033[32;49mbin/console server:stop.\033[39m"; \
```
stof: isn't it already displayed by the server:start command itself ?

sstok: No, `[OK] Server listening on http://127.0.0.1:8000`

stof: ah, it is only explained in the command help indeed, not when running the command

symfony/framework-bundle/3.3/Makefile
```diff
@@ -6,9 +6,22 @@ cache-warmup: cache-clear
...
+serve_as_php:
+	@echo "\033[32;49mServer listening on http://127.0.0.1:8000\033[39m";
 	@echo "Quit the server with CTRL-C."
-	@echo -e "Run \033[32mcomposer require symfony/web-server-bundle\033[39m for a better web server"
+	@echo "Run \033[32mcomposer require symfony/web-server-bundle\033[39m for a better web server"
 	php -S 127.0.0.1:8000 -t web
+
+serve:
+	@${MAKE} serve_as_sf || ${MAKE} serve_as_php
 .PHONY: serve
```
greg0ire: You have 2 .PHONY declarations now :P

sstok:
I made a different version with suggestions from @greg0ire 👍

*I really need to learn how to fully utilize Makefile.*

symfony/framework-bundle/3.3/Makefile
```diff
@@ -6,14 +6,22 @@ cache-warmup: cache-clear
 	@test -f bin/console && bin/console cache:warmup || echo "cannot warmup the cache (needs symfony/console)"
 .PHONY: cache-warmup
 
+CONSOLE=bin/console
+$(CONSOLE):
+	@exit
+
+serve_as_sf: $(CONSOLE)
+	@$(CONSOLE)|grep server:start > /dev/null
+	@$(CONSOLE) server:start || exit 0
```
stof: looks wrong to me. If there is a failure in the command, we want to report it in the main exit code, to let the external system know that it failed

symfony/framework-bundle/3.3/Makefile
```diff
@@ -6,9 +6,22 @@ cache-warmup: cache-clear
...
+
+	@echo "Quit the server with \033[32;49mbin/console server:stop.\033[39m"
+
+serve_as_php:
+	@echo "\033[32;49mServer listening on http://127.0.0.1:8000\033[39m";
 	@echo "Quit the server with CTRL-C."
-	@echo -e "Run \033[32mcomposer require symfony/web-server-bundle\033[39m for a better web server"
+	@echo "Run \033[32mcomposer require symfony/web-server-bundle\033[39m for a better web server"
```
stof: removing `-e` breaks the support of colors AFAIK (you are reverting #48)

symfony/framework-bundle/3.3/Makefile
sstok: For serve? Note that multiple .PHONY's is perfectly valid, and this was already used. The Makefiles are combined together so you don't have one absolute header section.

greg0ire: TIL. But you should add `serve_as_sf` and `serve_as_php` there too

sstok: Done, also found a way to suppress the make error when the command doesn't exist 😄

symfony/framework-bundle/3.3/Makefile
sstok: No, because when I use this is shows `-e Run composer require symfony/web-server-bundle for a better web server`

greg0ire: Strange bug true.

greg0ire: My colleague @L-P : "use printf, it's more portable, you don't know what echo really is here (shell built-in, make built-in, whatever)

greg0ire: (no flag is required for `printf`, BTW)

greg0ire: Let's back that up with [a very solid source](https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo)

nicolas-grekas: :+1: for printf also, that's really the portable way

sstok: Done

symfony/framework-bundle/3.3/Makefile
```diff
@@ -6,9 +6,22 @@ cache-warmup: cache-clear
 	@test -f bin/console && bin/console cache:warmup || echo "cannot warmup the cache (needs symfony/console)"
 .PHONY: cache-warmup
 
-serve:
-	@echo -e "\033[32;49mServer listening on http://127.0.0.1:8000\033[39m"
-	@echo "Quit the server with CTRL-C."
-	@echo -e "Run \033[32mcomposer require symfony/web-server-bundle\033[39m for a better web server"
+CONSOLE=bin/console
+$(CONSOLE):
```
greg0ire: maybe : `@printf "%s could not be found. You may use an alternate binary by specifying it like this : make CONSOLE=path/to/console some_target" "$(CONSOLE)"`

sstok: Done. Slightly different.

symfony/framework-bundle/3.3/Makefile
```diff
@@ -6,9 +6,24 @@ cache-warmup: cache-clear
...
+	@printf "\033[31;49m%s could not be found.\033[39m Run \033[32mcomposer require console\033[39m to install the console.\n" "$(CONSOLE)"
+	@printf "You may use an alternate binary by using: make CONSOLE=path/to/console some_target\n\n"
```
fabpot: I would remove this second line as I don't see when one would need this.

fabpot:
I've just tested and it works quite well. Well done!

Here is the output when console and the web server are not installed:

```
bin/console could not be found. Run composer require console to install the console.
You may use an alternate binary by using: make CONSOLE=path/to/console some_target

/bin/sh: bin/console: No such file or directory
Server listening on http://127.0.0.1:8000
Quit the server with CTRL-C.
Run composer require symfony/web-server-bundle for a better web server
php -S 127.0.0.1:8000 -t web
PHP 7.1.2 Development Server started at Fri May  5 09:17:46 2017
Listening on http://127.0.0.1:8000
Document root is /private/tmp/foobar/web
Press Ctrl-C to quit.
```

Is it possible to remove the `/bin/sh: bin/console: No such file or directory` part?

I think we should also use shortcuts when possible:

`Run composer req cli to install the Symfony console.`
`Run composer req webserver for a better web server`

fabpot:
Forgot to mention something: the `bin/console could not be found.` message should probably not be in red as the command works without it; so this is not an error to not have the console installed.

sstok:
I'm away for today, so feel free to push new commits ;)

sstok:
> Is it possible to remove the /bin/sh: bin/console: No such file or directory part?

No idea 😐 it's possible suppress errors in a rule but not when the rule itself itself it a target or something. I can make the console a rule but then you still need to check if console exists in each rule, is this acceptable?

```makefile
CONSOLE=bin/console
sf_console:
	@test -f $(CONSOLE) || printf "%s could not be found. Run \033[32mcomposer require cli\033[39m to install the Symfony console.\n" "$(CONSOLE)"
	@exit

serve_as_sf: sf_console
	@test -f $(CONSOLE) && $(CONSOLE)|grep server:start > /dev/null || ${MAKE} serve_as_php
	@$(CONSOLE) server:start || exit 0

	@printf "Quit the server with \033[32;49mbin/console server:stop.\033[39m\n"

serve_as_php:
	@printf "\033[32;49mServer listening on http://127.0.0.1:8000\033[39m\n";
	@printf "Quit the server with CTRL-C.\n"
	@printf "Run \033[32mcomposer require symfony/web-server-bundle\033[39m for a better web server\n"
	php -S 127.0.0.1:8000 -t web

serve:
	@${MAKE} serve_as_sf
.PHONY: sf_console serve serve_as_sf serve_as_php
```

fabpot:
@sstok your proposer changes look good to me.

sstok:
OK, all should be good now.

sstok:
OK, should be good now 👍

fabpot:
Thank you @sstok.

symfony/framework-bundle/3.3/Makefile
```diff
@@ -6,9 +6,23 @@ cache-warmup: cache-clear
...
-	@echo "Quit the server with CTRL-C."
-	@echo -e "Run \033[32mcomposer require symfony/web-server-bundle\033[39m for a better web server"
+CONSOLE=bin/console
+sf_console:
+	@test -f $(CONSOLE) || printf "Run \033[32mcomposer require cli\033[39m to install the Symfony console.\n"
+	@exit
+
+serve_as_sf: sf_console
+	@test -f $(CONSOLE) && $(CONSOLE)|grep server:start > /dev/null || ${MAKE} serve_as_php
+	@$(CONSOLE) server:start || exit 0
```
stof: exit with 1 when the server fails to start looks wrong to me

PR #55, opened 2017-05-04
Tweaked a help note about routing

javiereguiluz:
If you want to use annotations in your Flex app, you must execute this:

```
$ composer req sensio/framework-extra-bundle:^3.0
```

That's why I propose to reword the original help note to make it as close as possible to that command ... and to avoid confusing people with the Doctrine mention.

PS: it'd be cool if the package was named `symfony/annotations` instead of `sensio/framework-extra-bundle` 😉

### DIFF
--- a/symfony/framework-bundle/3.3/etc/routing.yaml
+++ b/symfony/framework-bundle/3.3/etc/routing.yaml
@@ -2,7 +2,7 @@
 #    path: /
 #    defaults: { _controller: 'App\Controller\DefaultController::index' }
 
-# Depends on sensio/framework-extra-bundle:^3.0 and doctrine/annotations
+# requires installing the 'sensio/framework-extra-bundle:^3.0' package
 #controllers:
 #    resource: ../src/Controller/
 #    type: annotation


symfony/framework-bundle/3.3/etc/routing.yaml
```diff
@@ -2,7 +2,7 @@
 #    path: /
 #    defaults: { _controller: 'App\Controller\DefaultController::index' }
 
-# Depends on sensio/framework-extra-bundle:^3.0 and doctrine/annotations
+# requires installing the 'sensio/framework-extra-bundle:^3.0' package
```
fabpot: Don't we also need `doctrine/annotations`? I think it's required as well

fabpot:
What about creating a `symfony/annotations` pack instead?

symfony/framework-bundle/3.3/etc/routing.yaml
javiereguiluz: I didn't include doctrine/annotations and it worked for me: FrameworkExtraBudnle requires DoctrineCommon which requires DoctrineAnnotations

xabbuh: Do you really need the FrameworkExtraBundle? Isn't the Doctrine Annotations package enough?

javiereguiluz: The @Route, @Method, @Cache, etc. annotations are defined in the FwExtraBundle, not in the Doctrine thing.

javiereguiluz:
Closing as no longer needed because we now have an annotations pack. Thanks!

PR #56, opened 2017-05-04
Proposed to remove a confusing help note about Doctrine

javiereguiluz:
I originally proposed to add this help note:

```
# Set "serverVersion" to your server version to avoid edge-case exceptions
and extra database calls
```

Now I propose to remove it because it's confusing:

* It doesn't truly explain the consequences: "edge cases" and "database calls" sounds vague.
* The proposed solution ("set the serverVersion") is confusing: which server version? how do I get that? how do I set up that option?

I think this option requires a lengthy explanation to be really understood, so maybe we should leave that for Symfony Docs.

### DIFF
--- a/doctrine/doctrine-bundle/1.6/manifest.json
+++ b/doctrine/doctrine-bundle/1.6/manifest.json
@@ -9,7 +9,6 @@
     "env": {
         "#1": "Format described at http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url",
         "#2": "For a sqlite database, use: \"sqlite:///%kernel.project_dir%/var/data.db\"",
-        "#3": "Set \"serverVersion\" to your server version to avoid edge-case exceptions and extra database calls",
         "DATABASE_URL": "mysql://root@127.0.0.1:3306/symfony?charset=utf8mb4&serverVersion=5.7"
     }
 }


Pierstoval:
Then leave that to Symfony docs and send a link to the docs.

The server version is important because else Doctrine performs a query to check server version and adapt container compilation depending on the server version. 

Check [Doctrine/DBAL/Platforms](https://github.com/doctrine/dbal/tree/master/lib/Doctrine/DBAL/Platforms) directory and see the different classes that can depend on the platform's version.

weaverryan:
Let's include just one URL where the whole string - including serverVersion is explained... likely a symfony docs page as you said.

And if we don't have it already, what about a post install message that says to update the settings in this file?

craigh:
I believe the need for this is removed in the next version of Doctrine?
refs https://github.com/doctrine/dbal/issues/990 and https://github.com/doctrine/dbal/pull/2671

weaverryan:
Maybe :). Is the extra query to determine the version only run once when the container is being built? If so, paas will have a problem - they'll build the container before the db is ready. Or, is the extra query on every request? That's kind of lame, but more of a perf optimization. Or, does Doctrine run the query once at runtime and then cache it after? If that's the case, I don't see any reason to keep this.

If someone has a few minutes to run down those answers, that would help (I'm on my phone!)

Pierstoval:
The extra query is run only on container compilation to determine which driver/platform will be used by Doctrine IIRC

weaverryan:
@Pierstoval If it's run on container compilation, then the `serverVersion` *is* probably needed... as you may be building the cache when you don't yet have access to the database. platform.sh is like this and I'm guessing that Heroku is as well. What I mean is: without the `serverVersion`, if the query is made at compile time, deploys would fail if you don't have database access at the time of building your container.

xabbuh:
The need for the option should be solved in Doctrine DBAL (see the linked issues above).

teohhanhui:
No, the changes in DBAL does not remove the need for the `serverVersion` parameter.

javiereguiluz:
Closing because this was reworded as part of #258.

PR #58, opened 2017-05-07
Load container config, instead of importing it

ro0NL:
`import()` is an implementation detail; see https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Config/Loader/GlobFileLoader.php#L26

Fixed ticket: https://github.com/symfony/symfony/pull/22638


fabpot:
Thank you @ro0NL.

PR #60, opened 2017-05-08
[LexikJWTAuthenticationBundle] Add aliases

ogizanagi:
ping @chalasr

### DIFF
--- a/lexik/jwt-authentication-bundle/2.3/manifest.json
+++ b/lexik/jwt-authentication-bundle/2.3/manifest.json
@@ -5,6 +5,7 @@
     "copy-from-recipe": {
         "etc/": "%ETC_DIR%"
     },
+    "aliases": ["jwt-auth"],
     "env": {
         "JWT_PRIVATE_KEY_PATH": "%ETC_DIR%/jwt/private.pem",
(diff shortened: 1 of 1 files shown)


lexik/jwt-authentication-bundle/2.3/manifest.json
```diff
@@ -5,6 +5,7 @@
     "copy-from-recipe": {
         "etc/": "%ETC_DIR%"
     },
+    "aliases": ["jwt-auth"],
```
javiereguiluz: If it's decided that this bundle is "the JWT solution for Symfony", then we could even go for adding `jwt` too as an alias.

ogizanagi: I wanted at first, but the bundle is a solution for setting a jwt authentication, not for consuming external jwt tokens for instance (even if it requires libraries that will probably do natively).

chalasr: Indeed, the bundle is written on top of libraries like [lcobucci/jwt](https://github.com/lcobucci/jwt), it is only about authentication while jwt can be used in different contexts. 👍 for jwt-auth

fabpot:
Thank you @ogizanagi.

PR #62, opened 2017-05-11
Doctrine Migrations

Ricknox:
This recipe adds doctrine migrations to the official Symfony recipes repository.


doctrine/doctrine-migrations-bundle/1.2/etc/packages/migrations.yaml
```diff
@@ -0,0 +1,5 @@
+doctrine_migrations:
+    dir_name: '%kernel.project_dir%/DoctrineMigrations'
+    namespace: App\Migrations
+    table_name: migration_versions
+    name: Application Migrations
```
stof: please remove the table_name and the name from this config, as requested in #57. We don't need to configure them explicitly in the project.

I would even remove the explicit configuration of the namespace here, keeping the default one. I don't see any reason to change it here (one could argue that you want to make the namespace fit your app, but your config will *not* make it compatible with PSR-4 anyway so this is a bad argument. And when using advanced feature to organize files per date instead of having a single folder with hundred of migrations, you *cannot* make code inside this folder fit PSR-4 at all, so I would not even try it).
This would then leave only the `dir_name` config (which is the one we really need due to the new Flex structure)

Ricknox: I agree! Changed in 5bace2591cc8241f1ac5e4310364487963a0a117.

stof:
@fabpot there was some discussion in the past about forcing bundles to use their DI alias for the naming of the YAML config file they add in `packages` (which is not the case here, due to `migrations.yaml` vs `doctrine_migrations`). Is it something we want to enforce ?

doctrine/doctrine-migrations-bundle/1.2/etc/packages/migrations.yaml
```diff
@@ -0,0 +1,2 @@
+doctrine_migrations:
+    dir_name: '%kernel.project_dir%/DoctrineMigrations'
```
fabpot: All directory names are using the underscore notation, having a `DoctrineMigrations` directory at the root of the project looks weird. What about something like `'%kernel.project_dir%/etc/migrations'` instead?

fabpot: I would also add `etc/migrations/.gitkeep` to the PR so that the directory is automatically created when installing the bundle.

stof: in #57, we said that migrations should live inside `src`, not at the root

stof: @fabpot do you think migration classes should go in `src` or `etc` ? Your previous comment suggests `etc`, but migrations are code.

fabpot: Indeed, that should be `src/`, not `etc/`.

Ricknox: √

Ricknox: @fabpot Should `src/migrations` not be `src/Migrations`. Because the Entity folder has a capital letter to in the directory. Let me know, I've added it with a Capital letter for now.

fabpot:
@stof is right, you rename `migrations.yaml` to `doctrine_migrations.yaml`

fabpot:
Thank you @Ricknox.

doctrine/doctrine-migrations-bundle/1.2/etc/packages/migrations.yaml
rybakit: Would it make more sense to use singular `Migration` instead of `Migrations` for the namespace?

Ricknox: @rybakit We only changed the default directory for now. If you want to customize the namespace or another setting of the bundle I suggest you take a look at the options at http://symfony.com/doc/current/bundles/DoctrineMigrationsBundle/index.html#configuration

/cc @fabpot @stof @javiereguiluz Might be a idea to add `"env": {}` to the manifest file with a link to `doc/current/bundles/DoctrineMigrationsBundle/`?

rybakit: @Ricknox Sorry, I meant if singular `src/Migration` as the default directory would be more appropriate here (as the migrations are php classes, and most of the libraries use singular names in their namespace to describe the domain, fe https://github.com/symfony/recipes/blob/master/doctrine/doctrine-bundle/1.6/etc/packages/doctrine.yaml#L12)

PR #66, opened 2017-05-14
Add a Github PR template

fabpot:


javiereguiluz:
What's the purpose of this template? If "contrib" accepts any license, we wouldn't need it. And if "contrib" doesn't accept all licenses, shouldn't we instead add a message explaining which licenses do we accept? Thanks!

fabpot:
Indeed, you're right, I will add the template and only keep the comment.

PR #70, opened 2017-05-20
Remove version_aliases configuration setting

fabpot:
`version_aliases` was a bad idea as it incurs too much manual maintenance as we need to remember to add new minor and major versions. That's fragile and we will probably forget to do it promptly. That's even worse for contributed recipes.

So, instead (and as already documented in the README, so not part of the diff), the version of the recipe if the minimum version that the recipe supports. So, if there is only one recipe for version "3.3", the recipe will be applied for all versions >= 3.3 (including 3.x, 4.x, ...).

If a version introduces changes that has an impact on the recipe , then a new directory with the version when the break appeared should be created with the new recipe, which makes total sense.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -63,25 +63,6 @@ more aliases to make it easier to install:
 
 Developers can now install this dependency with ``composer require acme-log``.
 
-``version_aliases`` option
-~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-This option lists all the additional dependency versions (using the ``x.y``
-format) that work with this very same recipe. This avoids duplicating recipes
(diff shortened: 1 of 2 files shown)


fabpot:
Note that a recipe for version 3.3 will be applied when installing version 4.0. I think that's the best thing to do as not applying the recipe when installing 4.0 is worse than trying to apply the recipe for 3.3. Proof if PHPUnit for which the recipe works for version 4, 5, and 6.

And if something break, fixing it is as easy as creating a new recipe. That also avoid duplicated recipes with no changes.

PR #71, opened 2017-05-21
Removed extra slash for sqlite example

mykbas:
Cannot create sqlite database, removing extra slash makes it work as expected.


mykbas:
smth strange, in doctrine.yaml, it works:
url: 'sqlite:///%kernel.project_dir%/var/data.db'

With dbal: url: '%env(DATABASE_URL)%', it doesn't work when set in .env:
DATABASE_URL='sqlite:///%kernel.project_dir%/var/data.db'

PR #73, opened 2017-05-23
[lexik/jwt-authentication-bundle] Fix configured key path

chalasr:
Only using `%ETC_DIR%` leads to

> Private key "etc/jwt/private.pem" does not exist or is not readable.

This fixes it by prefixing paths by `%kernel.project_dir%/` in the bundle config.


javiereguiluz:
Should we add a comment here https://github.com/symfony/recipes/blob/master/lexik/jwt-authentication-bundle/2.3/manifest.json#L10 explaining that the paths should be relative to the project dir?

chalasr:
@javiereguiluz That'd be hard since [comments aren't supported in json](https://stackoverflow.com/questions/244777/can-comments-be-used-in-json) :)

javiereguiluz:
See https://github.com/symfony/recipes/blob/master/doctrine/doctrine-bundle/1.6/manifest.json#L10

stof:
@chalasr  People don't read this JSON anyway. They read their .env.dist file

chalasr:
Of course, comment added!

fabpot:
Thank you @chalasr.

PR #74, opened 2017-05-23
[doctrine/doctrine-bundle] Escape DATABASE_URL env value

chalasr:
When requiring `doctrine/doctrine-bundle` (more often `orm-pack`), the following is added to the project `.env`:

```
DATABASE_URL=mysql://root@127.0.0.1:3306/symfony?charset=utf8mb4&serverVersion=5.7
```
The makefile seems to parse that properly (tested on OSX and debian).
But running `source .env` from the cli (for debugging the .env vars, as shown in [dotenv docs](http://symfony.com/doc/master/components/dotenv.html)) gives

> .env:15: parse error near `&'

This fixes it by adding quotes around the value.

### DIFF
--- a/doctrine/doctrine-bundle/1.6/manifest.json
+++ b/doctrine/doctrine-bundle/1.6/manifest.json
@@ -10,6 +10,6 @@
         "#1": "Format described at http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url",
         "#2": "For a sqlite database, use: \"sqlite:///%kernel.project_dir%/var/data.db\"",
         "#3": "Set \"serverVersion\" to your server version to avoid edge-case exceptions and extra database calls",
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @chalasr.

dunglas:
[Docker's `.env`](https://docs.docker.com/compose/env-file/) doesn't handle quotes in env files:

> There is no special handling of quotation marks (i.e. they will be part of the VAL, you have been warned ;) ).

And they don't plan to accept support for it: https://github.com/moby/moby/pull/28092#issuecomment-273873571

It means that the `.env` file generated by Flex doesn't work with Docker. It took me some time to figure it out.

chalasr:
See #98

greg0ire:
#98 was rejected and I got that issue today. I think a solution could be to escape the `&` with a backslash, seems to work in my shell.

greg0ire:
Escaping doesn't seem to work either. Not sure how to solve this...

PR #75, opened 2017-05-24
[Translation] Allow using "etc/translations" to store translations files

Pierstoval:
| Q             | A
| ------------- | ---
| License       | MIT

For now, translations are stored in Symfony's default directory which is `%kernel.root_dir%/Resources/translations`, resolved to `src/Resources/translations`.

I find it much more convenient to store it in `etc/translations`.

I'm not sure about some things though:

* The yaml file name, `translation.yaml`, as this config is part of the framework and we cannot "update" the `framework.yaml` file with this recipe 😕 Maybe should I rename it `symfony_translation.yaml` or `translator`, or even maybe `symfony_translator`?
* Adding `%locale%` parameter with `en` by default. As it's not mandatory to have a `%locale%` parameter in Symfony apps anymore, I thought it would be useful to setup at least this value and add a comment to tell the user to update it depending on his/her application.

What do you think?

### DIFF
--- a/symfony/translation/3.3/etc/packages/translation.yaml
+++ b/symfony/translation/3.3/etc/packages/translation.yaml
@@ -1,0 +1,11 @@
+parameters:
+    # Change this locale depending on your app's default language
+    locale: 'en'
+
+framework:
+    default_locale: '%locale%'
+    translator:
+        paths:
+            - '%kernel.project_dir%/etc/translations/'
+        fallbacks:
+            - '%locale%'
--- a/symfony/translation/3.3/etc/packages/translation.yaml
+++ b/symfony/translation/3.3/etc/packages/translation.yaml
@@ -1,11 +1,7 @@
-parameters:
-    # Change this locale depending on your app's default language
-    locale: 'en'
-
 framework:
     default_locale: '%locale%'
     translator:
         paths:
-            - '%kernel.project_dir%/etc/translations/'
+            - '%kernel.project_dir%/translations/'
         fallbacks:
             - '%locale%'
--- a/symfony/translation/3.3/manifest.json
+++ b/symfony/translation/3.3/manifest.json
@@ -1,6 +1,10 @@
 {
     "copy-from-recipe": {
-        "etc/": "%ETC_DIR%/"
+        "etc/": "%ETC_DIR%/",
+        "translations/": "translations/"
+    },
+    "container": {
+        "locale": "en"
     },
     "aliases": ["translator"]
 }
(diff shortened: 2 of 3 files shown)


pixocode:
:+1:

syrm:
+1

javiereguiluz:
I'd prefer to move htem to the `translations/` dir at the project root dir. Translations are not config, so they shouldn't go to etc/ (or config/).

symfony/translation/3.3/etc/packages/translation.yaml
```diff
@@ -0,0 +1,11 @@
+parameters:
+    # Change this locale depending on your app's default language
+    locale: 'en'
```
fabpot: This should be removed, no parameters should be defined under `etc/`. Instead,  you should use the `container` configurator to set this automatically in `container.yaml`

symfony/translation/3.3/etc/packages/translation.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+framework:
+    default_locale: '%locale%'
+    translator:
+        paths:
+            - '%kernel.project_dir%/etc/translations/'
```
fabpot: I would use `%kernel.project_dir%/translations/` instead as translations as as important as `templates/` for instance when you have a translated project.

Pierstoval:
@fabpot 
> I would use ``

I think the full answer didn't come out 😆 

I'll fix your reviews in a minute

symfony/translation/3.3/manifest.json
```diff
@@ -1,3 +1,10 @@
 {
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/",
+        "translations/": "translations/"
+    },
+    "container": {
```
Pierstoval: @fabpot it seems the bot is not yet aware of this option 😕

fabpot: I know, it was on purpose as I wanted to remove this feature as I didn't had any usage anymore (that's also why it was not documented).

Pierstoval: Now do you think it should be kept & documented? 😊

sstok:
Can't the locale be an ENV variable?

Pierstoval:
I'm not sure that it would be useful, as most of the time it's application-specific, not platform-specific. That's why we often put it in parameters instead of env vars

fabpot:
@Pierstoval Can you add the License header in the PR description (this is now enforced), thanks.

Pierstoval:
@fabpot Added the same contents as in [recipes-contrib PR template](https://github.com/symfony/recipes-contrib/blob/master/.github/PULL_REQUEST_TEMPLATE.md)

fabpot:
Thank you @Pierstoval.

PR #83, opened 2017-05-30
Rename .gitkeep files to .gitignore

teohhanhui:
| Q | A |
| --- | --- |
| License | MIT |

Let's not keep this legacy convention. *(pun intended)*

The **official** guidance is to use an empty `.gitignore` file.

See

https://git.wiki.kernel.org/index.php/GitFaq#Can_I_add_empty_directories.3F

and

https://git.wiki.kernel.org/index.php/SvnMigration#Empty_directories


fabpot:
Thank you @teohhanhui.

rogamoore:
Hi everyone,

I just ran into a problem with this change when deploying to CloudFoundry. Apparently they ignore ".gitignore" by default when pushing an app. See https://docs.cloudfoundry.org/devguide/deploy-apps/prepare-to-deploy.html#exclude
This causes some nice not-easy-to-debug errors - e.g. the src/Entity/ directory is missing and cache-warmup will fail.
I'm aware this is wrong behaviour on CFs side, just giving a note here for other people stumbling on that.

stof:
well, you should probably create your entities before deploying anyway (or disable the ORM and so there is no issue anymore)

rogamoore:
@stof true, but this can potentially affect any recipe
Personally I'm not really sure about the purpose of this change as I've never seen anyone using .gitignore to keep directories and I find the name misleading. Most projects use the (unofficial) .gitkeep convention.

PR #85, opened 2017-05-30
[lexik/jwt-authentication-bundle] Fix generate-jwt-keys target in Makefile

teohhanhui:
| Q | A |
| --- | --- |
| License | MIT |

The definition of the `generate-jwt-keys` target broke the `Makefile` when the `openssl` binary is not available.


fabpot:
Can you add the license header to make the bot happy?

fabpot:
Thank you @teohhanhui.

PR #86, opened 2017-05-30
[doctrine/doctrine-bundle] Provide better defaults

teohhanhui:
| Q             | A
| ------------- | ---
| License       | MIT

See https://github.com/symfony/recipes/pull/38#discussion_r119162734 for reasoning. Replaces #38.

### DIFF
--- a/doctrine/doctrine-bundle/1.6/etc/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/1.6/etc/packages/doctrine.yaml
@@ -1,6 +1,9 @@
 doctrine:
     dbal:
-        url: '%env(DATABASE_URL)%'
+        # Format described at http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url
+        # For a sqlite database, use: "sqlite:///%kernel.project_dir%/var/data.db"
+        # Set "serverVersion" to your server version to avoid edge-case exceptions and extra database calls
+        url: 'mysql://%env(DB_USER)%:%env(DB_PASSWORD)%@%env(DB_HOST)%:%env(DB_PORT)%/%env(DB_DBNAME)%?charset=utf8mb4&serverVersion=5.7'
     orm:
         auto_generate_proxy_classes: '%kernel.debug%'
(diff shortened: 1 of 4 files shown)


teohhanhui:
@jvasseur Please help me understand your downvote. I thought the point of Symfony Recipes is to provide good defaults. The user can effortlessly change it to use one environment variable for setting the URL, where there are special needs.

jvasseur:
I think that me should either use an url env variable with the url config key like before or one key per parameter with the corresponding env variable like in the standard edition. This "in-between" solution feels like trying to get a compromise without having any good benefit.

And I'm against using sqlite for test environment: there is a reason why we removed the db_driver parameter for the standard edition your application is probably made to work with a specific database type in mind, testing with another one seems like a bad idea.

javiereguiluz:
@teohhanhui in my opinion, the key here is considering the URL format a special need. This may be true in the past, but it doesn't look like that in the present and the future. The URL looks to be promoted by Doctrine docs as the default connection format and it's mandatory for any cloud-related thing. So the URL is the future instead of the individual config params.

The only problem that some people discussed was Docker, which seems to be uncapable of using these URLs and needs the params. However, I refuse to believe that there is not a solution to fix that in Docker.

teohhanhui:
> The only problem that some people discussed was Docker, which seems to be uncapable of using these URLs and needs the params. However, I refuse to believe that there is not a solution to fix that in Docker.

Not a problem for the Docker use case. It's easy to combine and set the environment variable from the `ENTRYPOINT` script.

I'm closing this then.

Pierstoval:
@jvasseur 
> And I'm against using sqlite for test environment: there is a reason why we removed the db_driver parameter for the standard edition your application is probably made to work with a specific database type in mind, testing with another one seems like a bad idea.

... Bad idea only when your application clearly depends on the DBMS. If it doesn't, Sqlite is a neat idea because resetting the database can be made in a few milliseconds instead of using doctrine's commands.
Was just my 2cents, because I also don't agree about _forcing_ the user to use Sqlite in test, even if I personally do it (I'm not an example, because I made everything possible to make my app not-dbms-dependent, but sometimes, business needs are different for other people).

PR #87, opened 2017-05-30
Change directory name to match directory style.

ricknox:
| Q             | A
| ------------- | ---
| License       | MIT

This pull request changes the Doctrine migrations directory name from `Migrations` to `Migration` so it is in sync with the other directories (Controller, Entity, Repository, etc. etc.).

### DIFF
--- a/doctrine/doctrine-migrations-bundle/1.2/etc/packages/doctrine_migrations.yaml
+++ b/doctrine/doctrine-migrations-bundle/1.2/etc/packages/doctrine_migrations.yaml
@@ -1,3 +1,2 @@
 doctrine_migrations:
     dir_name: '%kernel.project_dir%/src/Migration'
-    namespace: App\Migration


teohhanhui:
We should also add:

```yaml
doctrine_migrations:
    namespace: App\Migration
```

Ricknox:
@javiereguiluz @fabpot Whats the call on the namespace that @teohhanhui suggests? We left it out in the initial pull request of this package, should I add the namespace as well since everything in Flex is under the namespace `App/`?

javiereguiluz:
I don't use this bundle ... but it looks like the `namespace` is mandatory to make this work out-of-the-box. And `App\Migration` looks like the right namespace.

Ricknox:
@teohhanhui I have added it to the pull request and it now awaits approving from the core members. Thank you for your suggestion.

teohhanhui:
We'd want to exclude `App\Migration` from the project's autoloading.

stof:
There is indeed no reason to configure a PSR-4 compatible namespace here, as the loading of migrations is not based on autoloading at all

teohhanhui:
But I think it would improve the DX...

stof:
@teohhanhui DX would be lost again when using some bundle features organizing migrations in subfolder (as it would not be PSR-4 compatible anymore).
And making developers think it is fine to access other migration classes (by making autoloading work for them in simple cases) is a bad idea IMO, as this may actually break the usage of the library

teohhanhui:
In that case we should have this in `etc/migrations`?

stof:
Well, they're still code

teohhanhui:
If we want to keep them in `src`, I'm of the opinion that it should be excluded from autoloading.

Ricknox:
I do not understand whoever wants what. I think we need to hit a hammer to prevent discussion or a simple bundle as this.

stof:
@teohhanhui the only way to exclude them from the autoloading fully is to make them not follow PSR-4 (which is the case before this PR). If the path and namespace are following PSR-4, autoloading will be able to find the classes.

ricknox:
@stof I suggest to remove my commit 90395f83e3f2d932647d6919b79bad73dcd37657 and leave 3bb8f1b5ad0b913d350b79061aa29faace1b3602 intact so this pr can be merged for now.

ricknox:
Ready to review.

fabpot:
closing in favor of #172

PR #89, opened 2017-05-31
[lexik/jwt-authentication-bundle] gitignore JWT keys

### LINKED ISSUE #84: [lexik/jwt-authentication-bundle] Keys should be git-ignored
You certainly do not want to commit these to the repository.

teohhanhui:
| Q             | A
| ------------- | ---
| License       | MIT

Fixes #84

### DIFF
--- a/lexik/jwt-authentication-bundle/2.3/manifest.json
+++ b/lexik/jwt-authentication-bundle/2.3/manifest.json
@@ -3,13 +3,14 @@
         "Lexik\\Bundle\\JWTAuthenticationBundle\\LexikJWTAuthenticationBundle": ["all"]
     },
     "copy-from-recipe": {
-        "config/": "%CONFIG_DIR%"
+        "config/": "%CONFIG_DIR%",
+        "secrets/": "secrets"
     },
     "aliases": ["jwt-auth"],
     "env": {
         "#1": "Key paths should be relative to the project directory",
-        "JWT_PRIVATE_KEY_PATH": "%CONFIG_DIR%/jwt/private.pem",
-        "JWT_PUBLIC_KEY_PATH": "%CONFIG_DIR%/jwt/public.pem",
+        "JWT_PRIVATE_KEY_PATH": "secrets/jwt/private.pem",
+        "JWT_PUBLIC_KEY_PATH": "secrets/jwt/public.pem",
         "JWT_PASSPHRASE": "%generate(secret)%"
     }
 }
--- a/lexik/jwt-authentication-bundle/2.3/Makefile
+++ b/lexik/jwt-authentication-bundle/2.3/Makefile
@@ -3,7 +3,7 @@ generate-jwt-keys:
 ifndef OPENSSL_BIN
 	$(error "Unable to generate keys (needs OpenSSL)")
 endif
-	mkdir -p config/jwt
+	mkdir -p secrets/jwt
 	openssl genrsa -passout pass:${JWT_PASSPHRASE} -out ${JWT_PRIVATE_KEY_PATH} -aes256 4096
 	openssl rsa -passin pass:${JWT_PASSPHRASE} -pubout -in ${JWT_PRIVATE_KEY_PATH} -out ${JWT_PUBLIC_KEY_PATH}
 	@echo "\033[32mRSA key pair successfully generated\033[39m"
--- a/lexik/jwt-authentication-bundle/2.3/secrets/jwt/.gitignore
+++ b/lexik/jwt-authentication-bundle/2.3/secrets/jwt/.gitignore
@@ -1,0 +1,2 @@
+*
+!.gitignore


javiereguiluz:
Even if this proposal is "the right thing" to avoid committing sensitive data to the repo, I think the fix is not correct. The `etc/` dir (or `config/` if we finally rename it) should be committed entirely. Now that we're getting rid of the parameters.yml + parameters.yml.dist files, it's time to simplify things committing the etc/ dir entirely and putting all the sensitive credentials in the .env file.

teohhanhui:
> putting all the sensitive credentials in the .env file

From moby/moby#13490:

> * Environment Variables. Probably the most used, because it's part of the ["12 factor app"](http://12factor.net/). Environment variables are discouraged, because they are;
>    * Accessible by any proces in the container, thus easily "leaked"
>    * Preserved in intermediate layers of an image, and visible in docker inspect
>    * Shared with any container linked to the container

Docker now has a ["secrets"](https://docs.docker.com/engine/swarm/secrets/) feature (Swarm Mode only) which expose each piece of sensitive data as files under the `/run/secrets` directory.

dunglas:
Maybe can we just move the JWT keys in `var/jwt`?

sroze:
Yes, I think that this should definitely live in `var` if they are committed to the code repository.

sstok:
If the JWT keys are generated for each application separate (like secret), then 👍

chalasr:
Not sure about `var`. I use it myself and not so long ago someone told me "But var/ is about storage, temporary data, cache, logs... Removing keys just breaks the setup", and I think he is right.

dunglas:
But the keys should be different on each instance of an app. So they shouldn't be in `etc/`. Imo we should have a directory to store such sensitive informations (maybe `secret/`?).

Btw, I'm implementing a mechanism similar to `env()` but to read files containing secret data at runtime and inject them in the container as parameters (for instance, to use Docker secrets). It would be nice to have a directory to store such data (for those not using Docker).

teohhanhui:
I disagree about using `var`. Please see the relevant FHS section: http://www.pathname.com/fhs/pub/fhs-2.3.html#PURPOSE31

`etc` is the correct location.

teohhanhui:
Since `etc` has been renamed to `config`, I agree with @dunglas that we should put it under `secrets/jwt`.

Will update the PR.

dunglas:
I've almost finished secret support for the DI component. It just need some polish (probably when I'll be back from vacation).

lexik/jwt-authentication-bundle/2.3/manifest.json
```diff
@@ -3,13 +3,14 @@
...
+        "config/": "%CONFIG_DIR%",
+        "secrets/": "secrets"
     },
     "aliases": ["jwt-auth"],
     "env": {
         "#1": "Key paths should be relative to the project directory",
-        "JWT_PRIVATE_KEY_PATH": "%CONFIG_DIR%/jwt/private.pem",
-        "JWT_PUBLIC_KEY_PATH": "%CONFIG_DIR%/jwt/public.pem",
+        "JWT_PRIVATE_KEY_PATH": "secrets/jwt/private.pem",
+        "JWT_PUBLIC_KEY_PATH": "secrets/jwt/public.pem",
```
fabpot: Shoudn't it be `%PROJECT_DIR%/secrets/jwt/...`

teohhanhui: No, unless we want to support storing the key pair outside of the project directory?

teohhanhui: But if we go ahead with this, we should probably add a `secrets-dir` in Flex?

javiereguiluz: I agree with Fabien. This looks really strange. It should be `%CONFIG_DIR%/secrets/jwt/...` or `%PROJECT_DIR%/secrets/jwt/...`

teohhanhui: `%CONFIG_DIR%` resolves to `config`.
`%PROJECT_DIR%` resolves to the full path of the project directory.

What we might want to add is a `%SECRETS_DIR%` which resolves to `secrets`. But I still find it weird to see a `%SECRETS_DIR%` in an environment variable...

javiereguiluz: In Rails 5 they introduced full management of secrets. They are stored encrypted in a file called `secrets.yml`. The contents of the file are separated by environment:

```yaml
# secrets.yml

shared: # <-- this one means, "all environments"
    # ...

development:
    # ...

test:
    # ...

production:
    # ...
```

Maybe we can introduce something like that for Symfony 4.x.

teohhanhui: symfony/symfony#23901 introduced the `file` prefix.

dunglas: @javiereguiluz managing secret's encryption at this level looks like a bad idea to me. There are already plenty dedicated tools to do it like Docker and Kubernetes. IMO the best thing to to is to make it easy to use them in Symfony (and it's achieved with symfony/symfony#23901 ), not to reinvent the wheel.

I'm 👍 to introduce a new `secrets` directory and store things like JWT keys and Doctrine passwords in it (and retrieve them with symfony/symfony#23901).

sroze: The risk of the `secrets` directory is that people would be tempted to directly commit their credentials in the code repository... You use the example of Docker/Kubernetes and the secrets management is exactly meant to prevent people to add their secrets within the Docker image directly. As long as we give a very clear documentation that these secrets have to be development secrets and overwritten by production secrets while deploying production, then I'm 👍 with such `secrets` directory.

teohhanhui: Adding an empty directory with a `.gitignore` which ignores everything else should help.

dunglas: We can add this directory to the .gitignore file.

sroze:
Not convinced about that the DX with this directory in the `.gitignore` as it means you don't have a working development environment directly after cloning a project anymore. Even if our official recipes could generate development secrets, I'm not convinced about user secrets.

dunglas:
You can ignore the content of the directory while keeping it in the version system.

teohhanhui:
Secrets should never be checked in.

sroze:
> Secrets should never be checked in.

This is not true. Production secrets should not be checked in. Development secrets should be checked in, so as a developer, I don't have to pull secrets from anywhere else to have a development environment working.

teohhanhui:
> Development secrets should be checked in, so as a developer, I don't have to pull secrets from anywhere else to have a development environment working.

I disagree. But we can already see this is a never-ending discussion.

nicolas-grekas:
I'm closing as the PR is old, and Makefile have been removed.
Please resubmit if you think we still need to improve.

PR #92, opened 2017-06-05
Add the Behat recipe

sroze:
| Q             | A
| ------------- | ---
| License       | MIT

As Behat is part of the official Symfony documentation, I'm creating this PR in here. This recipe will create files already configured for Symfony 4 and will create a demo working scenario showing how to get access to Symfony's kernel and send a request to it.

### DIFF
--- a/behat/symfony2-extension/2.1/behat.yml.dist
+++ b/behat/symfony2-extension/2.1/behat.yml.dist
@@ -9,5 +9,4 @@ default:
         Behat\Symfony2Extension:
             kernel:
                 bootstrap: features/bootstrap/bootstrap.php
-                path: src/Kernel.php
                 class: App\Kernel
--- a/behat/symfony2-extension/2.1/behat.yml.dist
+++ b/behat/symfony2-extension/2.1/behat.yml.dist
@@ -1,0 +1,13 @@
+default:
+    suites:
+        default:
+            contexts:
+                - FeatureContext:
+                    kernel: '@kernel'
+
+    extensions:
+        Behat\Symfony2Extension:
+            kernel:
+                bootstrap: features/bootstrap/bootstrap.php
+                path: src/Kernel.php
+                class: App\Kernel
--- a/behat/symfony2-extension/2.1/features/bootstrap/FeatureContext.php
+++ b/behat/symfony2-extension/2.1/features/bootstrap/FeatureContext.php
@@ -1,0 +1,40 @@
+<?php
+
+use Behat\Behat\Context\Context;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\KernelInterface;
+
+class FeatureContext implements Context
+{
+    /**
+     * @var KernelInterface
+     */
+    private $kernel;
+
+    /**
+     * @var Response|null
+     */
+    private $response;
+
+    public function __construct(KernelInterface $kernel)
+    {
+        $this->kernel = $kernel;
+    }
+
+    /**
+     * @When this is a demo scenario that sends a request to :path
+     */
+    public function thisIsADemoScenarioThatSendsARequestTo(string $path)
+    {
+        $this->response = $this->kernel->handle(Request::create($path, 'GET'));
+    }
+
+    /**
+     * @Then the response should be received
+     */
+    public function theResponseShouldBeReceived()
+    {
+        assert($this->response !== null);
+    }
+}
--- a/behat/symfony2-extension/2.1/features/bootstrap/FeatureContext.php
+++ b/behat/symfony2-extension/2.1/features/bootstrap/FeatureContext.php
@@ -41,6 +41,8 @@ class FeatureContext implements Context
      */
     public function theResponseShouldBeReceived()
     {
-        assert($this->response !== null);
+        if ($this->response === null) {
+            throw new \RuntimeException('No response received');
+        }
     }
 }
--- a/behat/symfony2-extension/2.1/features/bootstrap/FeatureContext.php
+++ b/behat/symfony2-extension/2.1/features/bootstrap/FeatureContext.php
@@ -23,9 +23,9 @@ class FeatureContext implements Context
     }
 
     /**
-     * @When this is a demo scenario that sends a request to :path
+     * @When a demo scenario sends a request to :path
      */
-    public function thisIsADemoScenarioThatSendsARequestTo(string $path)
+    public function aDemoScenarioSendsARequestTo(string $path)
     {
         $this->response = $this->kernel->handle(Request::create($path, 'GET'));
     }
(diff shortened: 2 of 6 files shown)


behat/symfony2-extension/2.1/features/bootstrap/FeatureContext.php
```diff
@@ -0,0 +1,40 @@
...
+
+    public function __construct(KernelInterface $kernel)
+    {
+        $this->kernel = $kernel;
+    }
+
+    /**
+     * @When this is a demo scenario that sends a request to :path
+     */
+    public function thisIsADemoScenarioThatSendsARequestTo(string $path)
```
Pierstoval: Maybe `@When A demo scenario sends a Request`? WDYT?

behat/symfony2-extension/2.1/features/bootstrap/FeatureContext.php
```diff
@@ -0,0 +1,40 @@
...
+    {
+        $this->response = $this->kernel->handle(Request::create($path, 'GET'));
+    }
+
+    /**
+     * @Then the response should be received
```
Pierstoval: Maybe `@Then A response should be returned`? WDYT?

behat/symfony2-extension/2.1/features/demo.feature
```diff
@@ -0,0 +1,8 @@
+Feature:
+  In order to proves that the Behat Symfony extension is correctly installed
```
sstok: In order to prove the Behat Symfony extension is correctly installed. ??

behat/symfony2-extension/2.1/behat.yml.dist
```diff
@@ -0,0 +1,13 @@
...
+        default:
+            contexts:
+                - FeatureContext:
+                    kernel: '@kernel'
+
+    extensions:
+        Behat\Symfony2Extension:
+            kernel:
+                bootstrap: features/bootstrap/bootstrap.php
+                path: src/Kernel.php
```
stof: should be removed as the kernel is autoloadable.

sroze: Good point!

behat/symfony2-extension/2.1/features/demo.feature
sroze: Tell me more :)

behat/symfony2-extension/2.1/features/bootstrap/FeatureContext.php
sroze: Yes, I'm happy with that as well.

behat/symfony2-extension/2.1/features/bootstrap/FeatureContext.php
sroze: I would keep this one, as it's referring to the response of the request. Nah?

behat/symfony2-extension/2.1/features/demo.feature
xabbuh: s/proves/prove/ :)

behat/symfony2-extension/2.1/manifest.json
```diff
@@ -0,0 +1,7 @@
+{
+    "copy-from-recipe": {
+        "behat.yml.dist": "behat.yml.dist",
+        "features/": "features/"
+    },
+    "aliases": ["behat"]
```
Pierstoval: I suggest adding a `.gitignore` line for `behat.yml`

everzet:
Can we add a huge comment in both the feature file AND the context class saying that they are there only for demonstration reasons with maybe links and other pointers for good BDD resources?

sroze:
@everzet sure. Which link(s) would you suggest?

sroze:
This PR has been updated and all comments sorted. 

@fabpot what do you think? All good?

fabpot:
@sroze You need another Symfony core team approval to get it merged automatically :) ping @symfony/deciders

behat/symfony2-extension/2.1/features/bootstrap/bootstrap.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+use Symfony\Component\Dotenv\Dotenv;
+
+require __DIR__.'/../../vendor/autoload.php';
+
+// The check is to ensure we don't use .env in production
+if (!getenv('APP_ENV')) {
```
dunglas: Maybe can you check if the Dotenv class exists? For instance in my Docker skeleton the Dotenv component isn't installed because the loading of `.env` file is handled by Docker Compose itself: https://github.com/dunglas/symfony-docker-skeleton/blob/master/composer.json

sroze: In such example, it would mean that the `APP_ENV` environment variable is set and therefore `new Dotenv` is not called, right?

dunglas: Of course you're right :)

behat/symfony2-extension/2.1/features/bootstrap/bootstrap.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+use Symfony\Component\Dotenv\Dotenv;
+
+require __DIR__.'/../../vendor/autoload.php';
```
stof: this line is useless. Behat already loads the composer autoloading file much earlier than the loading of this file by the extension

sroze: @stof very good point. Double checked and removed the line.

behat/symfony2-extension/2.1/features/bootstrap/FeatureContext.php
```diff
@@ -0,0 +1,46 @@
...
+     */
+    public function theResponseShouldBeReceived()
+    {
+        assert($this->response !== null);
```
stof: `assert` is not a good implementation here. This may not always run, and it may not always throw an exception on failures (it does not throw by default).

Behat wants you to throw an exception on failure in step definitions.

sroze: AFAIK it throws an exception by default now, but I've changed it to manually throw a `RuntimeException`. That's actually better to explain to beginners 👍

sroze:
How much automated is this process then @fabpot? 😜

xabbuh:
The automatic merge is probably blocked by @Pierstoval's review.

fabpot:
@sroze It's completely automated (except when there is a bug in the algorithm :))

Pierstoval:
It could be nice if the bot could show the things that "stop" the PR to be mergeable, do you think it's possible @fabpot?

fabpot:
@Pierstoval That's already the case if you know where to look :) Open the report and have a look at the bottom of the HTML source. There is a comment with the auto-merge logs.

Pierstoval:
Report for this repo is unavailable as it's merged, I looked at report for #87, #89 and #38 and no information about merge logs:
image

xabbuh:
@Pierstoval It's in the HTML source:

bildschirmfoto 2017-06-24 um 13 23 17

Pierstoval:
Oh yeah, right, didn't see that. Why is it in HTML comments and not explicitly added as a reminder?

PR #96, opened 2017-06-10
Discard 403 error from production

adyassine:
As we have only one frontal we dont need to show 403 error on production

| Q             | A
| ------------- | ---
| License       | MIT


symfony/framework-bundle/3.3/web/index.php
```diff
@@ -23,6 +23,7 @@
     if (isset($_SERVER['HTTP_CLIENT_IP'])
         || isset($_SERVER['HTTP_X_FORWARDED_FOR'])
         || !(in_array(@$_SERVER['REMOTE_ADDR'], ['127.0.0.1', '::1']) || PHP_SAPI === 'cli-server')
+        && getenv('APP_ENV') !== 'prod'
```
xabbuh: This is only ever reached when the `APP_DEBUG` environment variable is set which should not be the case in production.

adyassine: Yes you're right !

PR #97, opened 2017-06-12
Routing configuration extraction

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT


javiereguiluz:
I've never cared or changed or configured this `router.strict_requirements` option. Is this something still used/needed? Because I need something like this for Twig extensions (e.g. generating the logout URL can throw an exception and that's bad) but I've never need it for generating routes.

sstok:
Strict requirements are about generating, when you generate an URL the parameters for that URL must match it's requirements, else you will never be able reach that URL.

https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Routing/Generator/UrlGenerator.php#L156-L181

The configuration name a bit unclear, and I actually had to check the source code to verify what I remembered :)

javiereguiluz:
@sstok yes ... but let me put you an example. Imagine that your route has a `page` param that requires integers and you pass `a` as the value. If this option ignores the requirements and lets you generate a URL like `....?page=a` that might briefly work in the template, but it will generate a big exception in the backend.

In summary ... this option looks to me like an illusion. It helps you believe that things will work ... but the generated URLs will probably be wrong and will generate exceptions short after.

fabpot:
That's a good discussion to have, but on symfony/symfony :) This was introduced by Jordi IIRC.

PR #98, opened 2017-06-13
[doctrine-bundle] Use long config for docker/source compatibility

chalasr:
| Q             | A
| ------------- | ---
| License       | MIT

https://github.com/symfony/recipes/pull/74#issuecomment-308262952

### DIFF
--- a/doctrine/doctrine-bundle/1.6/etc/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/1.6/etc/packages/doctrine.yaml
@@ -1,6 +1,11 @@
 doctrine:
     dbal:
-        url: '%env(DATABASE_URL)%'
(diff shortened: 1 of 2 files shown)


stof:
Well, this means that projects created by Flex are not working by default anymore on Heroku, Sensio Cloud and CloudFoundry (which expose DATABASE_URL).
so I'm -1 on this PR

chalasr:
Closing then, an issue should be created for https://github.com/symfony/recipes/pull/74#issuecomment-308262952

stof:
Well, AFAICT, the DotEnv component was created to support all (or at least most) features available when doing `source .env`, but Docker did not. I suggest opening an issue on Docker to see whether they are willing to support this. If no, it means people will have to choose between Docker support or `source` support.

PR #103, opened 2017-06-28
Support alternative bin dirs for the security checker

stof:
| Q             | A
| ------------- | ---
| License       | MIT

The auto-scripts entry now relies on the fact that Composer adds the bin-dir in the PATH instead of asking the Flex plugin to run a php-script.

Fixes https://github.com/symfony/flex/issues/109

### DIFF
--- a/sensiolabs/security-checker/4.0/manifest.json
+++ b/sensiolabs/security-checker/4.0/manifest.json
@@ -3,7 +3,7 @@
         "etc/": "%ETC_DIR%/"
     },
     "composer-scripts": {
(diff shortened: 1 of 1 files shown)


iammichiel:
> on the fact that Composer adds the bin-dir in the PATH

Do you have any information about this?

fabpot:
Thank you @stof.

stof:
@iammichiel before running scripts, composer prepends the bin-dir to the PATH env variables, so that all scripts in it can be called with just their name rather than passing the full path.

PR #105, opened 2017-06-29
Remove final attribute from the kernel class

iammichiel:
The Kernel is declared as `final` which breaks the cache:clear command included in the `framework-bundle`. 

I am not aware of all implication but is there any reason to keep this `final`? It seems the Kernel in the symfony standard edition on the 3.3 branch is not final. 

| Q             | A
| ------------- | ---
| License       | MIT

 Fixes symfony/flex#121

### DIFF
--- a/symfony/framework-bundle/3.3/src/Kernel.php
+++ b/symfony/framework-bundle/3.3/src/Kernel.php
@@ -11,7 +11,7 @@ use Symfony\Component\Routing\RouteCollectionBuilder;
 /**
  * @author Fabien Potencier <fabien@symfony.com>
  */
-final class Kernel extends BaseKernel
+class Kernel extends BaseKernel
 {
     use MicroKernelTrait;
 


stof:
the cache clearing command in the Makefile does not use the deprecated warmup feature of `cache:clear`, so I don't think it is affected by the issue

dunglas:
👎, using `final` class for this kind of classes is a best practice (https://ocramius.github.io/blog/when-to-declare-classes-final/) and as pointed out by @stof the option you try to use (`cache:clear` without `--no-warmup`) is deprecated.

stof:
However, we should ensure that a warning about the deprecated combination is display *before* the failure (and maybe add a check giving a better error message when the class is final)

fabpot:
Closing as the warning should be implemented in symfony/symfony. So, nothing we need to do here.

PR #106, opened 2017-07-03
Move etc/ to config/ and web/ to public/

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

Must be merged just after symfony/flex#122


README.rst
```diff
@@ -25,9 +25,9 @@ The following example shows the real directory structure of some Symfony recipes
                 manifest.json
         framework-bundle/
             3.3/
-                etc/
+                config/
                 src/
-                web/
+                public/
```
stof: should be moved before `src` to be consistent with the display of filesystem explorers (they generally sort folders by name)

PR #108, opened 2017-07-03
Added a recipe for Webpack Encore

javiereguiluz:
| Q             | A
| ------------- | ---
| License       | MIT

This fixes https://github.com/symfony/webpack-encore/issues/77 ... but we should discuss about:

* Aliases
* the initial webpack.config.js contents


symfony/webpack-encore/0.9/webpack.config.js
```diff
@@ -0,0 +1,23 @@
...
+
+Encore
+    // directory where all compiled assets will be stored
+    .setOutputPath('web/build/')
+    // what's the public path to this directory (relative to your project's document root dir)
+    .setPublicPath('/build')
+    // empty the outputPath dir before each build
+    .cleanupOutputBeforeBuild()
+    // allow sass/scss files to be processed
+    .enableSassLoader()
```
yceruto: then we need to add `node-sass` and `sass-loader` to devDependencies list by default too?

symfony/webpack-encore/0.9/webpack.config.js
```diff
@@ -0,0 +1,23 @@
+var Encore = require('@symfony/webpack-encore');
+
+Encore
+    // directory where all compiled assets will be stored
+    .setOutputPath('web/build/')
```
pyrech: Should now be `public/build/` \o/

stof:
Does it actually work ? There is no symfony/webpack-encore composer package AFAIK

symfony/webpack-encore/0.9/package.json
```diff
@@ -0,0 +1,11 @@
+{
+    "devDependencies": {
+        "@symfony/webpack-encore": "^0.9.1"
+    },
+    "scripts": {
+        "dev-server": "./node_modules/.bin/encore dev-server",
```
stof: should just be `encore dev-server`, relying on the fact that NPM/Yarn adds the `.bin` folder in the PATH when running scripts.
This would make the NPM script portable on Windows (where writing the path with `/` does not work)

javiereguiluz:
@stof you are right. It doesn't work. But there's a GitHub repo: https://github.com/symfony/webpack-encore  Maybe Flex could add support for this feature?

symfony/webpack-encore/0.9/webpack.config.js
```diff
@@ -0,0 +1,23 @@
...
+    // empty the outputPath dir before each build
+    .cleanupOutputBeforeBuild()
+    // allow sass/scss files to be processed
+    .enableSassLoader()
+    .enableSourceMaps(!Encore.isProduction())
+    // will output as web/build/app.js
+    // .addEntry('app', './assets/js/app.js')
+    // will output as web/build/global.css
+    // .addStyleEntry('global', './assets/css/app.scss')
+    // allow legacy applications to use $/jQuery as a global variable
```
stof: only in files processed by Webpack.

stof:
@javiereguiluz Flex is a composer plugin, adding automatic configuration for packages selected by the composer dependency solver. If there is no package, Composer will never select it, and so Flex will not be asked to configure it.

symfony/webpack-encore/0.9/webpack.config.js
dunglas: Sass requires Ruby, it would be a pain when creating Docker image. Can we use Less (or nothing) by default instead?

weaverryan: It uses `node-sass` - so it uses the C version of Sass. Still a problem @dunglas?

symfony/webpack-encore/0.9/webpack.config.js
weaverryan: Probably add some line breaks between each line and its description

fabpot:
Option B sounds good to me as well

symfony/webpack-encore/0.9/webpack.config.js
Pierstoval: There are some issues with node-sass regarding segfaults on many systems (7 issues about segfaults on [sass/node-sass](https://github.com/sass/node-sass/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20segmentation)) , why not recommend `less` instead of `sass` by default to avoid people arguing Encore not working with `sass` (when it's obviously a `node-sass` issue)?

dunglas: To push in this direction, Bootstrap changed its default implementation from SASS to LESS in version 3.

dunglas: Humm.. It looks like Bootstrap switched back to SASS for the upcoming v4, so I don't know :D

dunglas: Having fool Bootstrap compatibility (like the ability to import specific case) is a must have. So I finally join the SASS team.

Pierstoval: You're right, and Foundation and Materialize are using SASS too 😛 Seems that `node-sass` will benefit from this new recommendation from Symfony and Encore, let's hope the project will clean its current big issues with  👍

iammichiel: > To push in this direction, Bootstrap changed its default implementation from SASS to LESS in version 3.
Bootstrap has always used Less. They had a Sass-alternative package for the 3rd version. However as you say, they fully switched to the Sass framework starting at version 4. Anyway, Sass has won the game. Less will be around for legacy projects but any new project now should rely on Sass.

dunglas: Indeed: https://github.com/twbs/bootstrap/tree/v1.0.0/lib

dunglas: The SASS alternative exists since v1.2 (6 years ago 😱, we're getting old): https://github.com/twbs/bootstrap-sass/tree/v1.2.0

weaverryan:
@fabpot can you create the new repo? `symfony/webpack-encore-pack`? I can also put it under `weaverryan/...`

fabpot:
@weaverryan Done

javiereguiluz:
@weaverryan is it OK to put `"@symfony/webpack-encore": "^0.12.0"` in the default `package.json` provided by the recipe? If we update that version in the future, should we release a new recipe version? Thanks!

weaverryan:
@javiereguiluz I think we should version it independently - almost more like a version of the recipe. So I would use `^1.0.0` and we would tag the repo as 1.0. The ORM pack is also done this way: https://github.com/symfony/orm-pack/releases

fabpot:
Is this one ready? What needs to be done before merging? /cc @javiereguiluz @weaverryan

symfony/webpack-encore-pack/1.0/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "copy-from-recipe": {
+        "assets/": "%PROJECT_DIR%/",
```
fabpot: That does not work

fabpot:
Ok, I've released 1.0 of the related Composer package

symfony/webpack-encore/0.9/webpack.config.js
stof: @dunglas it was unofficial at that time

PR #109, opened 2017-07-04
Fix `make serve` with web-server-bundle

sstok:
| Q             | A
| ------------- | ---
| License       | MIT

When renaming the directory structure, this was forgotten.


fabpot:
Thank you @sstok.

PR #111, opened 2017-07-05
[framework-bundle] Set trusted proxies

teohhanhui:
| Q             | A
| ------------- | ---
| License       | MIT

Ideally, the user should not have to modify `index.php`. This removes one such case.

Space is used as separator, consistent with the `for` loop defined in POSIX shell: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09_04_03

### DIFF
--- a/symfony/framework-bundle/3.3/manifest.json
+++ b/symfony/framework-bundle/3.3/manifest.json
@@ -14,7 +14,10 @@
     "env": {
         "APP_ENV": "dev",
         "APP_DEBUG": "1",
-        "APP_SECRET": "%generate(secret)%"
+        "APP_SECRET": "%generate(secret)%",
+        "#1": "See https://symfony.com/doc/current/request/load_balancer_reverse_proxy.html",
+        "#TRUSTED_PROXIES": "0.0.0.0/0",
+        "#TRUSTED_HEADERS": "HEADER_FORWARDED"
     },
     "gitignore": [
         ".env",
--- a/symfony/framework-bundle/3.3/public/index.php
+++ b/symfony/framework-bundle/3.3/public/index.php
@@ -21,7 +21,9 @@ if (getenv('APP_DEBUG')) {
     Debug::enable();
 }
 
-// Request::setTrustedProxies(['0.0.0.0/0'], Request::HEADER_FORWARDED);
+if (($trustedProxies = getenv('TRUSTED_PROXIES')) && ($trustedHeaders = getenv('TRUSTED_HEADERS'))) {
+    Request::setTrustedProxies(explode(' ', $trustedProxies), constant(Request::class.'::'.$trustedHeaders));
+}
 
 $kernel = new Kernel(getenv('APP_ENV'), getenv('APP_DEBUG'));
 $request = Request::createFromGlobals();


teohhanhui:
This is my use case (Docker):

```sh
#!/bin/sh

if [ -z "$TRUSTED_PROXIES" ]; then
	export TRUSTED_PROXIES
	TRUSTED_PROXIES=$(ip -o -f inet addr show | awk '{print $4}' ORS=' ' | sed 's/ *$//g')
	[ -n "$TRUSTED_PROXIES" ] || { echo 'Could not determine TRUSTED_PROXIES'; exit 1; }
fi
```

symfony/framework-bundle/3.3/public/index.php
```diff
@@ -21,7 +21,11 @@
     Debug::enable();
 }
 
-// Request::setTrustedProxies(['0.0.0.0/0'], Request::HEADER_FORWARDED);
+$trustedProxies = explode(' ', getenv('TRUSTED_PROXIES') ?: '');
```
dunglas:
```php
if (($trustedProxies = getenv('TRUSTED_PROXIES')) && $trustedHeaders = getenv('TRUSTED_HEADERS') {
    Request::setTrustedProxies(explode(' ', $trustedProxies), constant(Request::class.'::'.$trustedHeaders));
}
```

WDYT about this one? It prevent a useless call to explode.

symfony/framework-bundle/3.3/manifest.json
```diff
@@ -14,7 +14,9 @@
     "env": {
         "APP_ENV": "dev",
         "APP_DEBUG": "1",
-        "APP_SECRET": "%generate(secret)%"
+        "APP_SECRET": "%generate(secret)%",
+        "#TRUSTED_PROXIES": "0.0.0.0/0",
```
dunglas: Maybe can you add:

    "#", "See https://symfony.com/doc/current/request/load_balancer_reverse_proxy.html",

symfony/framework-bundle/3.3/public/index.php
```diff
@@ -21,7 +21,9 @@
     Debug::enable();
 }
 
-// Request::setTrustedProxies(['0.0.0.0/0'], Request::HEADER_FORWARDED);
+if (($trustedProxies = getenv('TRUSTED_PROXIES')) && ($trustedHeaders = getenv('TRUSTED_HEADERS'))) {
```
dunglas: The second parenthesis around `$trustedHeaders = getenv('TRUSTED_HEADERS')` are useless.

teohhanhui: It makes it easier to read when the first and second part use the same amount of parentheses. There is no drawback anyway.

teohhanhui:
ping @fabpot

symfony/framework-bundle/3.3/public/index.php
Pierstoval: If you want readability, maybe assign the vars before like this is more readable?

```php
$trustedProxies = getenv('TRUSTED_PROXIES');
$trustedHeaders = getenv('TRUSTED_HEADERS');
if ($trustedProxies && $trustedHeaders) {
    // ...
}
```

I know it's more lines of code, but it's actually the most readable way IMO

teohhanhui: I'm not sure that really improves readability, and not sure if that's worth it...

teohhanhui:
Anything still blocking this?

teohhanhui:
ping @fabpot

symfony/framework-bundle/3.3/manifest.json
```diff
@@ -14,7 +14,10 @@
     "env": {
         "APP_ENV": "dev",
         "APP_DEBUG": "1",
-        "APP_SECRET": "%generate(secret)%"
+        "APP_SECRET": "%generate(secret)%",
+        "#1": "See https://symfony.com/doc/current/request/load_balancer_reverse_proxy.html",
+        "#TRUSTED_PROXIES": "0.0.0.0/0",
+        "#TRUSTED_HEADERS": "HEADER_FORWARDED"
```
dunglas: IMO, it should not be configurable using environment variables.

Anyway, it should default too `X_HEADER_FORWARDED`, almost all existing software (Varnish, Nginx) and Cloud Platform (Kubernetes) use this one for now.

symfony/framework-bundle/3.3/public/index.php
```diff
@@ -21,7 +21,9 @@
...
+    Request::setTrustedProxies(explode(' ', $trustedProxies), constant(Request::class.'::'.$trustedHeaders));
```
dunglas: I suggest using JSON instead. It's more standard, more reliable and in sync with `symfony/symfony#23901` (it has a `json` prefix but not a `csv` one).

nicolas-grekas:
Closing in favor of #182
Thank you for raising the idea @teohhanhui

PR #112, opened 2017-07-07
Fix default namespace for generated migration classes

Lctrs:
| Q             | A
| ------------- | ---
| License       | MIT

Currently, migration classes are generate with the default namespace `Application\Migrations`, which means they cannot be autoloaded since the default namespace of Symfony Flex is `App`. This PR fixes it.

### DIFF
--- a/doctrine/doctrine-migrations-bundle/1.2/config/packages/doctrine_migrations.yaml
+++ b/doctrine/doctrine-migrations-bundle/1.2/config/packages/doctrine_migrations.yaml
@@ -1,2 +1,3 @@
 doctrine_migrations:
     dir_name: '%kernel.project_dir%/src/Migrations'
+    namespace: 'App\Migrations'


stof:
Translation classes of doctrine/migrations are **not** meant to be autoloadable. Doctrine Migrations will not rely on the autoloader to load them (and making them autoloadable could even open the door to bugs)

Lctrs:
Oh okay. I made this because the autoloading feature of the container tried to autoload the migration classes.

Maybe it could be worth it to add a note asking to add the `Migrations` directory to the `exclude` key at https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/config/packages/app.yaml#L18 since it's not very obvious ?

stof:
Well, migrations must not be registered as service either

Lctrs:
Yeah. Good point. I guess it's just friday.

PR #113, opened 2017-07-12
[doctrine-bundle] Docker support

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

Requires https://github.com/dunglas/symfony-docker-skeleton/pull/5 and https://github.com/symfony/flex/pull/128.

### DIFF
--- a/doctrine/doctrine-bundle/1.6/manifest.json
+++ b/doctrine/doctrine-bundle/1.6/manifest.json
@@ -31,7 +31,7 @@
             "    # - ./docker/db/data:/var/lib/mysql:rw",
             "  # To expose the port on the host",
             "  # ports:",
-            "  #   - \"3306:3306\""
+            "  #   - \"3306\""
         ],
         "volumes": ["db-data:"]
     }
--- a/doctrine/doctrine-bundle/1.6/manifest.json
+++ b/doctrine/doctrine-bundle/1.6/manifest.json
@@ -22,7 +22,7 @@
     "docker-compose": {
         "docker-compose.yml": {
             "services": [
-                "db:",
+                "database:",
                 "  image: postgres:${POSTGRES_VERSION:-13}-alpine",
                 "  environment:",
                 "    POSTGRES_DB: ${POSTGRES_DB:-app}",
--- a/doctrine/doctrine-bundle/1.6/manifest.json
+++ b/doctrine/doctrine-bundle/1.6/manifest.json
@@ -10,7 +10,7 @@
         "#1": "Format described at https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url",
         "#2": "For an SQLite database, use: \"sqlite:///%kernel.project_dir%/var/data.db\"",
         "#3": "For a PostgreSQL database, use: \"postgresql://db_user:db_password@127.0.0.1:5432/db_name?serverVersion=11&charset=utf8\"",
-        "#4": "For the Docker integration, use: \"mysql://symfony:ChangeMe@db:3306/app?serverVersion=mariadb-10.5\"",
+        "#4": "For the Docker integration, use: \"mysql://symfony:ChangeMe@db:3306/app?serverVersion=mariadb-10.5.6\"",
         "#5": "IMPORTANT: You MUST configure your server version, either here or in config/packages/doctrine.yaml",
         "DATABASE_URL": "mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=5.7"
     },
@@ -18,7 +18,7 @@
     "docker-compose": {
         "services": [
             "db:",
-            "  image: mariadb:10.4",
+            "  image: mariadb:10.5.6",
             "  environment:",
             "    - MYSQL_DATABASE=app",
             "    # You should definitely change the password in production",
@@ -29,8 +29,9 @@
             "    - db-data:/var/lib/mysql:rw",
             "    # You may use a bind-mounted host directory instead, so that it is harder to accidentally remove the volume and lose all your data!",
             "    # - ./docker/db/data:/var/lib/mysql:rw",
-            "  ports:",
-            "    - \"3306\""
+            "  # To expose the port publicly",
+            "  # ports:",
+            "  #   - \"3306\""
         ],
         "volumes": ["db-data: {}"]
     }
--- a/doctrine/doctrine-bundle/1.6/manifest.json
+++ b/doctrine/doctrine-bundle/1.6/manifest.json
@@ -9,29 +9,32 @@
     "env": {
         "#1": "Format described at https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url",
         "#2": "For an SQLite database, use: \"sqlite:///%kernel.project_dir%/var/data.db\"",
-        "#3": "For a PostgreSQL database, use: \"postgresql://db_user:db_password@127.0.0.1:5432/db_name?serverVersion=11&charset=utf8\"",
-        "#4": "For the Docker integration, use: \"mysql://symfony:ChangeMe@db:3306/app?serverVersion=mariadb-10.5.6\"",
+        "#3": "For a MySQL database, use: \"mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=5.7\"",
         "#5": "IMPORTANT: You MUST configure your server version, either here or in config/packages/doctrine.yaml",
-        "DATABASE_URL": "mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=5.7"
+        "DATABASE_URL": "postgresql://db_user:db_password@127.0.0.1:5432/db_name?serverVersion=13&charset=utf8"
     },
-    "dockerfile": ["RUN docker-php-ext-install pdo_mysql"],
+    "dockerfile": [
+        "RUN apk add --no-cache --virtual .pgsql-deps postgresql-dev; \\",
+        "\tdocker-php-ext-install -j$(nproc) pdo_pgsql; \\",
+        "\tapk add --no-cache --virtual .pgsql-rundeps so:libpq.so.5; \\",
+        "\tapk del .pgsql-deps"
+    ],
     "docker-compose": {
         "services": [
             "db:",
-            "  image: mariadb:${MARIADB_VERSION:-10.5.6}",
+            "  image: postgres:${POSTGRES_VERSION:-13}-alpine",
             "  environment:",
-            "    MYSQL_DATABASE: ${MYSQL_DATABASE:-app}",
+            "    POSTGRES_DB: ${POSTGRES_DB:-app}",
             "    # You should definitely change the password in production",
-            "    MYSQL_PASSWORD: ${MYSQL_PASSWORD:-ChangeMe}",
-            "    MYSQL_RANDOM_ROOT_PASSWORD: \"true\"",
-            "    MYSQL_USER: ${MYSQL_USER:-symfony}",
+            "    POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-ChangeMe}",
+            "    POSTGRES_USER: ${POSTGRES_USER:-symfony}",
             "  volumes:",
-            "    - db-data:/var/lib/mysql:rw",
+            "    - db-data:/var/lib/postgresql/data:rw",
             "    # You may use a bind-mounted host directory instead, so that it is harder to accidentally remove the volume and lose all your data!",
-            "    # - ./docker/db/data:/var/lib/mysql:rw",
+            "    # - ./docker/db/data:/var/lib/postgresql/data:rw",
             "  # To expose the port on the host",
             "  # ports:",
-            "  #   - \"3306\""
+            "  #   - \"5432\""
         ],
         "volumes": ["db-data:"]
     }
--- a/doctrine/doctrine-bundle/1.6/manifest.json
+++ b/doctrine/doctrine-bundle/1.6/manifest.json
@@ -18,21 +18,21 @@
     "docker-compose": {
         "services": [
             "db:",
-            "  image: mariadb:10.5.6",
+            "  image: mariadb:${MARIADB_VERSION:-10.5.6}",
             "  environment:",
-            "    - MYSQL_DATABASE=app",
+            "    MYSQL_DATABASE: ${MYSQL_DATABASE:-app}",
             "    # You should definitely change the password in production",
-            "    - MYSQL_PASSWORD=ChangeMe",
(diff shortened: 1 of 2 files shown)


doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
```diff
@@ -8,12 +8,8 @@ parameters:
 doctrine:
     dbal:
         # configure these for your database server
-        driver: 'pdo_mysql'
-        server_version: '5.7'
-        charset: utf8mb4
-        default_table_options:
-            charset: utf8mb4
-            collate: utf8mb4_unicode_ci
+        driver: 'pdo_pgsql'
```
Pierstoval: Is there any good reason to twitch to pgsql for default projects? Mysql is still the most used for simple projects 😕

dunglas: Because Postgres is better,  has more features (JSON queries, advanced pagination and many many more...) and is as easy to use than MySQL.

OskarStark: And has better default values

maxhelias: A PR already exists about this. #411

fabpot: @dunglas  Can you revert this change?

dunglas: @fabpot you mean switching back to MySQL?

doctrine/doctrine-bundle/1.6/manifest.json
```diff
@@ -10,6 +10,29 @@
...
+            "    - POSTGRES_DB=symfony",
+            "    - POSTGRES_USER=symfony",
+            "    # You should definitely change the password in production",
+            "    - POSTGRES_PASSWORD=!ChangeMe!",
+            "  volumes:",
+            "    - db-data:/var/lib/postgresql/data:rw",
+            "    # You may use a bind-mounted host directory instead, so that it is harder to accidentally remove the volume and lose all your data!",
+            "    # - ./docker/db/data:/var/lib/postgresql/data:rw",
+            "  ports:",
+            "    - \"5432:5432\""
```
fabpot: Like for the Messenger recipe, I would let Docker choose a random port for localhost to allow running several projects in parallel.

fabpot:
Anyone willing to finish this one?

maxhelias:
@fabpot I can take care of it if it's OK for everyone

dunglas:
I'm on it @maxhelias

doctrine/doctrine-bundle/1.6/manifest.json
```diff
@@ -10,6 +10,26 @@
         "#1": "Format described at https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url",
         "#2": "For an SQLite database, use: \"sqlite:///%kernel.project_dir%/var/data.db\"",
         "#3": "Configure your db driver and server_version in config/packages/doctrine.yaml",
-        "DATABASE_URL": "mysql://db_user:db_password@127.0.0.1:3306/db_name"
+        "DATABASE_URL": "mysql://symfony:password@db:3306/symfony?charset=utf8mb4&serverVersion=5.7"
```
maxhelias: I'm not in favor to put the option of the database in the DSN. cf. https://github.com/symfony/recipes/pull/258#issuecomment-346142034

doctrine/doctrine-bundle/1.6/manifest.json
```diff
@@ -10,6 +10,26 @@
         "#1": "Format described at https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url",
         "#2": "For an SQLite database, use: \"sqlite:///%kernel.project_dir%/var/data.db\"",
         "#3": "Configure your db driver and server_version in config/packages/doctrine.yaml",
-        "DATABASE_URL": "mysql://db_user:db_password@127.0.0.1:3306/db_name"
+        "DATABASE_URL": "mysql://symfony:password@db:3306/symfony"
```
fabpot: `db` should be reverted to `127.0.0.1`.

dunglas: If we do that, the recipe will not work "out of the box" anymore

fabpot: If we change the name to `db`, the recipe will not work anymore when not using Docker :)

dunglas: But it’s already not working in this case isn’t it? I mean even with a local installation of MariaDB, the DB name, the login and the password have to be changed.

bocharsky-bw: Why not create a commented out env var specifically for Docker like:
```
"#4": "If you use Docker - uncomment the next line"
"#DATABASE_URL": "mysql://symfony:password@db:3306/symfony"
```

And leave the original one unchanged:
```
"DATABASE_URL": "mysql://db_user:db_password@127.0.0.1:3306/db_name"
```

dunglas: My main goal with the Docker support is to have something totally transparent for the end user, as we’ve done for API Platform.

The user should be able to download the Docker skeleton, run docker-compose up, run docker-compose exec composer require orm, and everything must work seamlessly, without having to tweak config files or even to know how Docker works in the beginning.

@fabpot could it be possible to set a different env var depending if Docker support is enabled or not? By introducing a new key in the manifest for instance?

maxhelias: Or we can just set in the environment variable a default value like `{db:-127.0.0.1:3306}`.

After that we will be able in the `EnvConfigurator` to replace if docker support is enabled or not.

```php 
preg_replace('#^{(\w+):-(\w+.+)}$#', $this->composer->getPackage()->getExtra()['symfony']['docker'] ? '${1}' : '${2}', $env);
```

WDYT ?

dunglas:
@fabpot this one is now ready to be merged.

I added a hack in https://github.com/dunglas/symfony-docker/pull/74 to ensure that everything will work just with a simple `composer req doctrine`.

maxhelias (approved):

doctrine/doctrine-bundle/1.6/manifest.json
```diff
@@ -10,7 +10,29 @@
...
+            "    - MYSQL_PASSWORD=ChangeMe",
+            "    - MYSQL_RANDOM_ROOT_PASSWORD=true",
+            "    - MYSQL_USER=symfony",
+            "  volumes:",
+            "    - db-data:/var/lib/mysql:rw",
+            "    # You may use a bind-mounted host directory instead, so that it is harder to accidentally remove the volume and lose all your data!",
+            "    # - ./docker/db/data:/var/lib/mysql:rw",
+            "  # To expose the port publicly",
+            "  # ports:",
+            "  #   - \"3306\""
```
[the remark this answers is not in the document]
maxhelias: It's hardcode in Flex : https://github.com/symfony/flex/blob/main/src/Configurator/DockerComposeConfigurator.php#L34-L35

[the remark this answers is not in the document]
Pierstoval: Well, it's probably a bit too much, but I'd recommend the creation of a special configurator for a `docker-compose.override.yml.dist` file, so maintainers can explicitly define what can be "publicly" exposed, or what kind of "machine-specific" configuration can be used. 

Exposing public ports is not the best, it can cause conflicts, or priority issues (like two apps listening to the same port), etc., and having them directly in the `docker-compose.yml` file, even commented, might invite users to put the public ports in it.

Having the `dist` file invites the user to create their own use-case when needed, instead of having it enabled by default.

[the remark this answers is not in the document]
dunglas: > Well, it's probably a bit too much, but I'd recommend the creation of a special configurator for a docker-compose.override.yml.dist file, so maintainers can explicitly define what can be "publicly" exposed, or what kind of "machine-specific" configuration can be used.

@Pierstoval haha, actually I'm already on it. But it's not a blocker for this one. We'll be able to update it to leverage the new capabilities of the configurator later.

> Exposing public ports is not the best, it can cause conflicts, or priority issues (like two apps listening to the same port), etc., and having them directly in the docker-compose.yml file, even commented, might invite users to put the public ports in it.

It will be not be necessary anymore with the new configurator, indeed.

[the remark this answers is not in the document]
Pierstoval: > actually I'm already on it

You're always in advance 🤣 👍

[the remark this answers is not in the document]
dunglas: And here we are: symfony/flex#705

chalasr (dismissed):

nicolas-grekas:
I'd vote +2 if this were using PG instead of mariadb ;)

dunglas:
I'm also totally for using Postgres. It was what I proposed when I opened this PR but it has been refused at this time (3 years ago, see comments above). IIRC @lyrixx was also in favor of using Postgres by default.

Using Postgres instead of MariaDB/MySQL, in addition to the usual benefits, also has the following ones:

* out of the box support for the PG NOTIFY transport for Messenger
* ability for the API Platform distribution, which already ships Postgres, to use this recipe too

WDYT @fabpot?

Pierstoval:
All the docs related to database handling with Symfony will have to change too, if default recipes start using Docker-Compose, especially if MySQL/MariaDB is changed for Postgres.

This might introduce a whole new good practice for newcomers :)

fabpot:
I'm all for psql as well as I'm using it exclusively.
Please, do NOT approve until I test it with Flex thoroughly (thank you).

doctrine/doctrine-bundle/1.6/manifest.json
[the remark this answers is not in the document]
fabpot: I don't understand the issue here. On the localhost, the port is going to be random, so there is no possible conflicts.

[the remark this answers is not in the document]
dunglas: If we put the port in `docker-compose.yml` instead of not in `docker-compose.override.yml` the port will be exposed publicly even in production, which can be a security issue.

dunglas:
I switched to Postgres in the last commit.

nicolas-grekas (dismissed):

doctrine/doctrine-bundle/1.6/manifest.json
```diff
@@ -9,8 +9,39 @@
...
+        "#3": "For a MySQL database, use: \"mysql://db_user:db_password@127.0.0.1:3306/db_name?serverVersion=5.7\"",
+        "#5": "IMPORTANT: You MUST configure your server version, either here or in config/packages/doctrine.yaml",
+        "DATABASE_URL": "postgresql://db_user:db_password@127.0.0.1:5432/db_name?serverVersion=13&charset=utf8"
+    },
+    "dockerfile": [
+        "RUN apk add --no-cache --virtual .pgsql-deps postgresql-dev; \\",
+        "\tdocker-php-ext-install -j$(nproc) pdo_pgsql; \\",
+        "\tapk add --no-cache --virtual .pgsql-rundeps so:libpq.so.5; \\",
+        "\tapk del .pgsql-deps"
+    ],
```
fabpot, line 21: This session is very opinionated.

doctrine/doctrine-bundle/1.6/manifest.json
```diff
@@ -9,8 +9,39 @@
...
+    "docker-compose": {
+        "docker-compose.yml": {
+            "services": [
+                "db:",
```
fabpot, line 25: should be database

fabpot (commented):

fabpot:

fabpot:

doctrine/doctrine-bundle/1.6/manifest.json
dunglas: What do you mean? Postgres-opinionated? That's the case, but there are no alternatives as far as I know.

fabpot (approved):

chalasr (approved):

doctrine/doctrine-bundle/1.6/manifest.json
```diff
@@ -9,8 +9,39 @@
...
+                "  volumes:",
+                "    - db-data:/var/lib/postgresql/data:rw",
+                "    # You may use a bind-mounted host directory instead, so that it is harder to accidentally remove the volume and lose all your data!",
+                "    # - ./docker/db/data:/var/lib/postgresql/data:rw"
+            ],
+            "volumes": ["db-data:"]
+        },
+        "docker-compose.override.yml": {
+            "services": [
+                "db:",
```
pyrech, line 41: Shouldn't this service be named `database` too?

chalasr: correct :) Up for a PR @pyrech?

pyrech: Yep :slightly_smiling_face:

pyrech: See https://github.com/symfony/recipes/pull/843 :wink:

fabpot: You're right, can you do a PR?

doctrine/doctrine-bundle/1.6/manifest.json
```diff
@@ -9,8 +9,39 @@
...
+        "RUN apk add --no-cache --virtual .pgsql-deps postgresql-dev; \\",
+        "\tdocker-php-ext-install -j$(nproc) pdo_pgsql; \\",
+        "\tapk add --no-cache --virtual .pgsql-rundeps so:libpq.so.5; \\",
+        "\tapk del .pgsql-deps"
+    ],
+    "docker-compose": {
+        "docker-compose.yml": {
+            "services": [
+                "database:",
+                "  image: postgres:${POSTGRES_VERSION:-13}-alpine",
```
alexander-schranz, line 26: @dunglas Not sure if I did already ask this 🙈 , but was there a specific reasons why this yaml files are instead of 4 spaces 2 spaces. And so are not similar to symfony config files?
Make it harder in handling .editorconfig files.

fabpot: 4 spaces are the way to go indeed.

alexander-schranz: Here we go: https://github.com/symfony/recipes/pull/1018 :)

PR #123, opened 2017-07-21
Move routing configuration to routes/ instead of routing/

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT


fabpot:
That makes it easier to make a difference from routing configuration (`config/packages/routing.yaml`) and routes configuration (`config/routes/*`)

PR #126, opened 2017-07-21
[API Platform] Update the API Platform config to use the new mapping paths

meyerbaptiste:
| Q             | A
| ------------- | ---
| License       | MIT

We recently changed the way to configure resources' paths in API Platform 2.1.0 beta 3, see https://github.com/api-platform/core/pull/1271.


dunglas:
The failure is a false positive.

fabpot:
Thank you @meyerbaptiste.

PR #130, opened 2017-07-28
Updated the MAILER_URL env var to avoid SwiftMailer issues

### LINKED ISSUE #95: Swiftmailer default URL
Regarding the [Swiftmailer manifest](https://github.com/symfony/recipes/blob/master/symfony/swiftmailer-bundle/2.5/manifest.json) the default value of the `MAILER_URL` parameter is `smtp://localhost:25?encryption=&auth_mode=`.

Which is an invalid url. I got the following exception : `The  encryption is not supported` (note the double space between The & encryption)

Regarding of [`Symfony\Bundle\SwiftmailerBundle\DependencyInjection\SwiftmailerTransportFactory`](https://github.com/symfony/swiftmailer-bundle/blob/master/DependencyInjection/SwiftmailerTransportFactory.php) I found that the URL query part is parsed with `parse_str` and the values are added one after the other without trying to convert it.

We may have 2 strategies here : 

- Modifiy the `resolveOptions` method of `SwiftmailerTransportFactory` to replace empty string with null
- Modify the value of the `MAILER_URL` parameter in the manifest, and remove these parameters

javiereguiluz:
| Q             | A
| ------------- | ---
| License       | MIT

This fixes #95.

### DIFF
--- a/symfony/swiftmailer-bundle/2.5/manifest.json
+++ b/symfony/swiftmailer-bundle/2.5/manifest.json
@@ -8,7 +8,7 @@
     "env": {
         "#1": "For Gmail as a transport, use: \"gmail://username:password@localhost\"",
         "#2": "To disable sending emails, use: \"null://localhost\"",
-        "MAILER_URL": "MAILER_URL=smtp://localhost?encryption=ssl&auth_mode=login&username=&password="
+        "MAILER_URL": "smtp://localhost?encryption=ssl&auth_mode=login&username=&password="
     },
     "aliases": ["mailer", "mail"]
 }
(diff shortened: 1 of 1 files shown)


symfony/swiftmailer-bundle/2.5/manifest.json
```diff
@@ -8,7 +8,7 @@
     "env": {
         "#1": "For Gmail as a transport, use: \"gmail://username:password@localhost\"",
         "#2": "To disable sending emails, use: \"null://localhost\"",
-        "MAILER_URL": "smtp://localhost:25?encryption=&auth_mode="
+        "MAILER_URL": "smtp://localhost?encryption=ssl&auth_mode=login&username=&password="
```
fabpot: I'd prefer to fix the underlying issue in the bundle.

javiereguiluz:
Closing in favor of https://github.com/symfony/swiftmailer-bundle/issues/186

PR #131, opened 2017-07-28
Proposed a security.yaml file for the "test" environment

javiereguiluz:
| Q             | A
| ------------- | ---
| License       | MIT

Just a proposal. This is a common practice in Symfony apps and it's was surprising not having it when upgrading the Symfony Demo to use Flex.

### DIFF
--- a/symfony/security-bundle/3.3/config/packages/security.yaml
+++ b/symfony/security-bundle/3.3/config/packages/security.yaml
@@ -6,7 +6,7 @@ security:
         dev:
             pattern: ^/(_(profiler|wdt)|css|images|js)/
             security: false
-        # if the name of the firewall changes, update it in the packages/test/security.yaml file too
+        # if you change the name of the firewall, update it in the packages/test/security.yaml file too
         main:
             anonymous: null
 
--- a/symfony/security-bundle/3.3/config/packages/security.yaml
+++ b/symfony/security-bundle/3.3/config/packages/security.yaml
@@ -6,6 +6,7 @@ security:
         dev:
             pattern: ^/(_(profiler|wdt)|css|images|js)/
             security: false
+        # if the name of the firewall changes, update it in the packages/test/security.yaml file too
         main:
             anonymous: null
 
--- a/symfony/security-bundle/3.3/config/packages/test/security.yaml
+++ b/symfony/security-bundle/3.3/config/packages/test/security.yaml
@@ -2,5 +2,6 @@
 # See https://symfony.com/doc/current/cookbook/testing/http_authentication.html
 security:
     firewalls:
+        # replace 'main' by the name of your own firewall
         main:
             http_basic: ~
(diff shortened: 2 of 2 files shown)


Pierstoval:
Isn't this behavior prohibited by [symfony/security-bundle's configuration](https://github.com/symfony/security-bundle/blob/master/DependencyInjection/MainConfiguration.php#L229) ?

javiereguiluz:
@Pierstoval are you sure? We document this feature: https://symfony.com/doc/current/testing/http_authentication.html and I've been using it for a long time in lots of projects.

Pierstoval:
Oh yeah, nevermind, it's all about new keys, not existing keys, it's ok 😛

stof:
yeah, SecurityBundle let's you reconfigure existing firewalls. What it forbids is adding new firewalls, because merging configuration would make it harder to control the order of firewalls in this case (reconfiguring an existing firewall does not move it in the list)

fabpot:
I don't think this is a good idea. What if there is no `main` firewall or if it is named differently?

javiereguiluz:
The main security.yaml recipe creates a "main" firewall (https://github.com/symfony/recipes/blob/master/symfony/security-bundle/3.3/config/packages/security.yaml#L9). I supposed we could use the same name for the "test" env. Besides, in the Symfony Docs we use the same "main" name everywhere, and my guess is that most people never changes this (only advanced developers ... but they will quickly realize that it must be changed in test/security.yaml too).

fabpot:
Can we at least add a comment like done in the docs?

fabpot:
I would also add a comment in the main security.yml to warn about changing the dev file when renaming the "main" firewall.

fabpot:
@javiereguiluz How do we move forward on this one?

javiereguiluz:
I've added the help comments you asked for. Before merging/rejecting, I'd ask our security experts: @weaverryan and @chalasr. Thanks!

symfony/security-bundle/3.3/config/packages/security.yaml
```diff
@@ -6,6 +6,7 @@ security:
         dev:
             pattern: ^/(_(profiler|wdt)|css|images|js)/
             security: false
+        # if the name of the firewall changes, update it in the packages/test/security.yaml file too
```
fabpot: `if you change the name of the firewall, update it ...`

weaverryan:
I'm not sure about this... this simplifies testing... except if you're testing your authentication itself (and then it's impossible). And if you're system uses some more interesting authentication mechanism (e.g. API tokens, or more simply, a Guard form login where you load User data from some SSO API), then it will fail: this only works if your normal authentication mechanism uses the user provider in the traditional way.

I'd like to give a proactive recommendation, but I'm short on time right this moment. I like the thinking though: how can we *truly* simplify / improve the testing experience?

fabpot:
Didn't see @weaverryan comment, reverting now as I agree with him.

Pierstoval:
@weaverryan if you want to test a Guard, you can still unit test it, that's the big advantage of having the whole authentication process at the same place. The problem comes, as you say it, for other built-in systems, like oauth or `form_login`.

PR #133, opened 2017-07-28
[framework-bundle][console] Introduce APP_DIR environment variable

voronkovich:
| Q             | A
| ------------- | ---
| License       | MIT

In the Symfony demo application (see https://github.com/symfony/symfony-demo/pull/617) we need to set an absolute path to a database, but we can't get it in the `.env` file. This PR introduces an environment variable which contains the absolute path to the application's folder (it works like the `%kernel.project_dir%` container parameter).

Example:
```sh
# .env
DATABASE_URL="sqlite:///${APP_DIR}/var/data/blog.sqlite"
```

### DIFF
--- a/symfony/console/3.3/bin/console
+++ b/symfony/console/3.3/bin/console
@@ -17,7 +17,9 @@ if (!class_exists(Application::class)) {
 }
 
 if (!getenv('APP_ENV')) {
-    (new Dotenv())->load(__DIR__.'/../.env');
+    $dotEnv = new Dotenv();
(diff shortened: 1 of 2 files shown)


javiereguiluz:
By the way, given that this env var is the project dir, shouldn't be called `PROJECT_DIR` instead of `APP_DIR`? (or `SYMFONY_PROJECT_DIR`)

voronkovich:
@fabpot, I think this problem can't be solved in the Kernel, because we can't instantiate the Kernel before loading the `.env` file (we need the `APP_ENV` and the `APP_DEBUG` variables).

PR #135, opened 2017-07-30
[DoctrineMigrationBundle] Add `namespace` value to doctrine_migrations.yaml

VolCh:
Now migrations generated in `Application\Migrations` namespace by default that cause error in autoloader. This PR fix the namespace.

| Q             | A
| ------------- | ---
| License       | MIT


BoShurik:
https://github.com/symfony/recipes/pull/112

PR #138, opened 2017-07-31
[console] Add an option allowing to ignore the .env file

voronkovich:
| Q             | A
| ------------- | ---
| License       | MIT

In some cases (`server:run`, `server:start`) it's needed to disable the `.env` file loading. See https://github.com/symfony/symfony/issues/23723


nicolas-grekas:
I don't see the need for this personally.
If you want to achieve that, you can still use:
`APP_ENV=dev ./bin/console server:run`

voronkovich:
@nicolas-grekas, using the `APP_ENV=dev ./bin/console server:run` causes the impossibility to refresh the environment variables when the `.env` file was changed. See https://github.com/symfony/symfony/issues/23723

PR #144, opened 2017-08-02
Revert the change of null to ~

javiereguiluz:
| Q             | A
| ------------- | ---
| License       | MIT

I can't believe that #143 was merged. To me it's **a big step backwards** in the right direction Symfony 4 was heading.

Why? Because in Symfony 4 we're replacing a lot of things that need an explanation for **things that are self-explained**.

Examples?

* "etc/" is a great name for config ... but "config/" is better and needs no explanation. 
* "web/" is a great name for front controllers ... but "public/" is better because it's the standard in the computing industry and it fits multiple use cases, not only web apps.
* "app.php" and "app_dev.php" are great front controller names ... but "index.php" is better and requires no explanation.
* "app.yaml" and "container.yaml" are great names ... but a single "services.yaml" file is better and requires no explanation.
* etc.

So, why should we change the perfect `null` value which requires no explanation, by the strange `~` symbol that requires explaining and learning it.

*(You may tell me that every existing Symfony developer knows the meaning of `~`. That's true ... but we're not making Symfony 4 only for the existing Symfony developers. We want to onboard new developers).*

### DIFF
--- a/symfony/framework-bundle/3.3/config/packages/framework.yaml
+++ b/symfony/framework-bundle/3.3/config/packages/framework.yaml
@@ -1,14 +1,14 @@
 framework:
     secret: '%env(APP_SECRET)%'
     #default_locale: en
-    #csrf_protection: ~
+    #csrf_protection: null
     #http_method_override: true
-    #trusted_hosts: ~
+    #trusted_hosts: null
     # https://symfony.com/doc/current/reference/configuration/framework.html#handler-id
     #session:
     #    handler_id: session.handler.native_file
     #    save_path: '%kernel.project_dir%/var/sessions/%kernel.environment%'
-    #esi: ~
-    #fragments: ~
+    #esi: null
+    #fragments: null
     php_errors:
         log: true
--- a/symfony/framework-bundle/3.3/config/packages/test/framework.yaml
+++ b/symfony/framework-bundle/3.3/config/packages/test/framework.yaml
@@ -1,4 +1,4 @@
 framework:
-    test: ~
+    test: null
     session:
         storage_id: session.storage.mock_file
--- a/symfony/routing/3.3/config/packages/routing.yaml
+++ b/symfony/routing/3.3/config/packages/routing.yaml
@@ -1,3 +1,3 @@
 framework:
     router:
-        strict_requirements: ~
+        strict_requirements: null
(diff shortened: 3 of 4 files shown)


fabpot:
My reasoning is that `~` is used to also mean "use the default config" and `null` is confusing in that regards.

javiereguiluz:
That's a very good point! However, unless I'm wrong, Symfony never sees or deals with the `~` symbol. The Yaml component transforms `~` into `null` instantly when parsing the config file.

fabpot:
Of course, there is no difference at all from a code point of view. I was referring as DX here: when one wants to express: "give me the default behavior/config", I think `~` is better than `null`.

javiereguiluz:
Another solution could be to forget about `~` and `null` and use explicit config that requires no explanation:

### Before

```yaml
framework:
    secret: '%env(APP_SECRET)%'
    #default_locale: en
    #csrf_protection: ~
    #http_method_override: true
    #trusted_hosts: ~
    # https://symfony.com/doc/current/reference/configuration/framework.html#handler-id
    #session:
    #    handler_id: session.handler.native_file
    #    save_path: '%kernel.project_dir%/var/sessions/%kernel.environment%'
    #esi: ~
    #fragments: ~
    php_errors:
        log: true
```

### After

```yaml
framework:
    secret: '%env(APP_SECRET)%'
    #default_locale: en
    #csrf_protection: { enabled: false }
    #http_method_override: true
    #trusted_hosts: ['example.com']
    # https://symfony.com/doc/current/reference/configuration/framework.html#handler-id
    #session:
    #    handler_id: session.handler.native_file
    #    save_path: '%kernel.project_dir%/var/sessions/%kernel.environment%'
    #esi: { enabled: false }
    #fragments: { enabled: false }
    php_errors:
        log: true
```

Pierstoval:
Your last argument makes sense, indeed having `{ enabled: true }` is self-explanatory, but the issue is that it doesn't fit the "Gimme default conf" DX strategy, which is the goal of the `~` or `null` values.

I also agree with @fabpot about the symbol which has been used very widely for default config in many projects for years. That's even the case for some config params in Rails, for example.

chalasr:
To me `~` is just the preferred YAML syntax for getting a `null` once parsed. 
Writing default values instead of using `null` is indeed more explicit, perhaps better for our default configuration.

PR #145, opened 2017-08-03
Fallback to line formatter on missing cli dumper

greg0ire:
The ConsoleFormatter is not so great when it does not come configured
with a dumper. A better fix might be to provide a poor man's dumper when
the var dumper component is not installed.

| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/monolog-bundle/3.1/config/packages/dev/monolog.php
+++ b/symfony/monolog-bundle/3.1/config/packages/dev/monolog.php
@@ -2,6 +2,7 @@
 
 use Symfony\Component\Config\Resource\ClassExistenceResource;
 use Symfony\Component\Console\Application;
+use Symfony\Component\VarDumper\Dumper\CliDumper;
 
 $handlers = [
     'main' => [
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Some context: the original issue was that the application didn't have VarDumper installed and then the logs were poorly formatted: no context, no extra, etc.

greg0ire:
@javiereguiluz I added more context in the commit message.

sroze:
Shouldn't this be directly managed in MonologBundle?

greg0ire:
@sroze you mean like this?

> A better fix might be to provide a poor man's dumper when
the var dumper component is not installed.

fabpot:
I would indeed fix this issue upstream, instead of adding a workaround here.

nicolas-grekas:
ping @greg0ire

greg0ire:
@nicolas-grekas I'll try to have a look at that soonish

greg0ire:
@nicolas-grekas here you go: https://github.com/symfony/symfony/pull/24598

PR #146, opened 2017-08-03
Remove the Exception folder from autowiring configuration

iammichiel:
| Q             | A
| ------------- | ---
| License       | MIT

By default, the `Exception` folder is also registered for Autowiring. It should be not be the case. 
As the folder exists in many projects, it might be worth adding it to the exclusion filter.

### DIFF
--- a/symfony/framework-bundle/3.3/config/services.yaml
+++ b/symfony/framework-bundle/3.3/config/services.yaml
@@ -19,7 +19,7 @@ services:
         resource: '../src/*'
         # you can exclude directories or files
         # but if a service is unused, it's removed anyway
-        exclude: '../src/{Entity,Repository,Tests}'
+        exclude: '../src/{Entity,Repository,Exception,Tests}'
 
     # controllers are imported separately to make sure they're public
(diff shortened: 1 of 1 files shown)


Pierstoval:
I don't think that's a problem. As services are private by default, they'll be automatically removed if no service injects them, as well as other directories.

Only sensitive classes should be excluded, and I don't think `Exception` should be.

javiereguiluz:
I have no opinion about this ... but I just want to say that after migrating a complex app to full service autowiring, I didn't have any problem with the Exception/ folder (I had some problems with other folders and I reported some issues in the main Symfony repo). But let's wait and read more opinions about this. Thanks!

greg0ire:
Did you experience any performance slowdown or other issue because of this? Or is it just because you feel it would make sense? I'm afraid this could lead to unforeseen consequences if, for some reason, some people have services in that namespace.

fabpot:
Closing as it does not have any drawbacks to keep it as is. We try to limit the number of exceptions.

PR #148, opened 2017-08-04
Use escaping rather than quoting

greg0ire:
This should make the recipe compatible with docker-compose's .env files,
which will keep quotes in the variable:

> There is no special handling of quotation marks (i.e. they will be
part of the VAL, you have been warned ;) ).

Refs #74

| Q             | A
| ------------- | ---
| License       | MIT


greg0ire:
I will be able to test this Monday, not sure if the environment variable that this produces inside the container is correct.

PR #156, opened 2017-08-22
Dont use getenv(), use $_SERVER instead

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

As described in https://github.com/laravel/framework/issues/7354 and linked issues, getenv() is not thread safe - this can lead to issue with concurrent requests, on Windows especially.

### DIFF
--- a/symfony/framework-bundle/3.3/public/index.php
+++ b/symfony/framework-bundle/3.3/public/index.php
@@ -8,11 +8,11 @@ use Symfony\Component\HttpFoundation\Request;
 require __DIR__.'/../vendor/autoload.php';
 
 // The check is to ensure we don't use .env in production
-if (!getenv('APP_ENV')) {
+if (!isset($_SERVER['APP_ENV'])) {
     (new Dotenv())->load(__DIR__.'/../.env');
 }
 
-if (getenv('APP_DEBUG')) {
+if ($_SERVER['APP_DEBUG'] ?? true) {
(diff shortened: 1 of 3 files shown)


symfony/framework-bundle/3.3/public/index.php
```diff
@@ -23,7 +23,7 @@
 
 // Request::setTrustedProxies(['0.0.0.0/0'], Request::HEADER_FORWARDED);
 
-$kernel = new Kernel(getenv('APP_ENV'), getenv('APP_DEBUG'));
+$kernel = new Kernel($_SERVER['APP_ENV'] ?? 'dev', $_SERVER['APP_DEBUG'] ?? true);
```
stof: the behavior has changed. Not setting the env variables was in no-debug mode before. Now, the only way to set it in no-debug mode is to set it to the empty string (which won't work for the console, which expects `'0'`)

yceruto: and `(bool) '0'` is `false`: https://github.com/symfony/symfony/blob/edb8c33f35b981b72adbe23798753703edfe3819/src/Symfony/Component/HttpKernel/Kernel.php#L83

stof: my bad, I thought that the only falsy string was the empty one.

At least, this is consistent with `empty('0')`, which also considers it as empty...

PR #157, opened 2017-08-23
Wrap CONSOLE value with quote so it can work on a path with spaces

jamalsa:
| Q             | A
| ------------- | ---
| License       | MIT

Currently if we run make command in spaced path it will return error: Command not found. This PR fix that by wrapping CONSOLE value with quote.


jamalsa:
It seems Symfony doesn't allowed command to be wrapped in Makefile task. I will close this PR and open another PR soon.

PR #159, opened 2017-08-23
Escape space in CONSOLE path

jamalsa:
| Q             | A
| ------------- | ---
| License       | MIT

(Recreate PR #158 with fixed code)
Currently if we run make command in spaced path it will return error: Command not found. This PR fix that by escaping spaces in path using sed command.

### DIFF
--- a/symfony/framework-bundle/3.3/Makefile
+++ b/symfony/framework-bundle/3.3/Makefile
@@ -1,4 +1,4 @@
-CONSOLE := $(shell which bin/console)
+CONSOLE := $(shell which bin/console | sed 's/ /\\ /g')
 sf_console:
 ifndef CONSOLE
 	@printf "Run \033[32mcomposer require cli\033[39m to install the Symfony console.\n"


jamalsa:
Not sure why CI failed this commit because I am not wrapping command but escaping it. Is this because Makefile contains quote?

fabpot:
Don't worry about the CI, that's a false positive.

javiereguiluz:
I'm not an expert in Bash, but looking for ways to escape white spaces, I found [this alternative solution](https://stackoverflow.com/a/12811033/2804294):

```
If you are using bash, you can use its builtin printf's %q formatter
(type help printf in bash):

FILENAME=$(printf %q "$FILENAME")

This will not only quote space, but also all special characters for shell.
```

I don't know if it's better or worse

fabpot:
@javiereguiluz's solution looks more elegant... but I'm no bash expert either. /cc @tucksaun?

tucksaun:
@javiereguiluz's solution is indeed more elegant.
But it's weird escaping is necessary in the first place: in this case, `which` should not look into the $PATH and should only return `bin/console` and not an absolute path.

fabpot:
@tucksaun Indeed.

@jamalsa Can you tell us under which circumstances you get an issue with the current code? Are you using Windows?

aik099:
Just a guess, but maybe path to website on developer's machine has space in it, e.g. `C:\Users\username\My Documents\Projects\symfony-projecy\bin\console`.

fabpot:
@aik099 I understand that, but `which bin/console` returns `bin/console`, not an absolute path. At least, on MacOS.

aik099:
On Linux however (tried on Slackware) it returns absolute path.

damienflament:
@aik099 I'm on ArchLinux, and the `which` command gives a relative path.

It depends on the _type_ of `which` you invoke. There are you're shell builtin and the binary. Both seems to have a different behavior.

Find them (using the Zsh `where` builtin):

```console
> where which
which: shell built-in command
/usr/bin/which
```

Try with both of them:
```console
> which bin/console
bin/console

> /usr/bin/which bin/console
/path/to/your/project/bin/console
```

Let's see the binary help message:

```console
> /usr/bin/which --help
Usage: /usr/bin/which [options] [--] COMMAND [...]
Write the full path of COMMAND(s) to standard output.

  --version, -[vV] Print version and exit successfully.
  --help,          Print this help and exit successfully.
  --skip-dot       Skip directories in PATH that start with a dot.
  --skip-tilde     Skip directories in PATH that start with a tilde.
  --show-dot       Don't expand a dot to current directory in output.
  --show-tilde     Output a tilde for HOME directory for non-root.
  --tty-only       Stop processing options on the right if not on tty.
  --all, -a        Print all matches in PATH, not just the first
  --read-alias, -i Read list of aliases from stdin.
  --skip-alias     Ignore option --read-alias; don't read stdin.
  --read-functions Read shell functions from stdin.
  --skip-functions Ignore option --read-functions; don't read stdin.

Recommended use is to write the output of (alias; declare -f) to standard
input, so that which can show aliases and shell functions. See which(1) for
examples.

If the options --read-alias and/or --read-functions are specified then the
output can be a full alias or function definition, optionally followed by
the full path of each command used inside of those.

Report bugs to <which-bugs@gnu.org>.
```

So, the `which` command can give you an absolute or relative path depending on your OS but also on the availability of a shell builtin command.

It seems taking care of spaces within paths is necessary !

fabpot:
see #159

damienflament:
@fabpot Hmm... You linked to this same PR...

PR #160, opened 2017-08-24
Change the default mailer configuration

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

The current configuration for the mailer is done in 2 different places:

* `.env` configures the mailer to use localhost as a transport;
* but the `dev` environment disables delivery.

That's very confusing. As `.env` is for dev, I propose to use a `null` transport by default (which disables delivery) and remove `disable_delivery` in the `dev` configuration. That makes the configuration centralized in one file instead of two (and yes, I had a WTF moment with this :)).

I'm not sure about the test environment as we don't have a good story about test env + env vars for now, so I haven't changed it for now.


pyrech:
:+1: This caused me a WTF few days ago when I deleted the dev config file but Flex recreated it on another install

PR #162, opened 2017-08-28
Make it possible create environment-specific services and parameters.

simensen:
| Q             | A
| ------------- | ---
| License       | MIT

I ran into issues trying to create parameters based on `%env()%` during compile-time. This made sense and was expected once I thought about it more.

In discussion with @nicolas-grekas, the solution I was looking for was being able to set parameters directly based on environment. However, `parameters.ini` is no longer a thing going forward.

Since there was no easy or obvious way to do this, we discussed adding `/conf/services_{envs}.yaml` as optional files that can be read to configure the container's services and parameters based on the appropriate environment.

This should allow me to configure `/conf/services_dev.yaml` to be specific to my local develompent environment without having any impact on anyone else running my application.

### DIFF
--- a/symfony/framework-bundle/3.3/src/Kernel.php
+++ b/symfony/framework-bundle/3.3/src/Kernel.php
@@ -42,6 +42,7 @@ class Kernel extends BaseKernel
             $loader->load($confDir.'/packages/'.$this->environment.'/**/*'.self::CONFIG_EXTS, 'glob');
         }
         $loader->load($confDir.'/services'.self::CONFIG_EXTS, 'glob');
+        $loader->load($confDir.'/services_'.$this->environment.self::CONFIG_EXTS, 'glob');
     }
 
     protected function configureRoutes(RouteCollectionBuilder $routes): void


javiereguiluz:
I'm :-1: about this. The removal of the prefixed config files is one of the most essential features of Flex. Fabien worked hard designing a new config system which departs from the old Symfony conventions to simplify everything ... and you now are asking to complicate things again mixing old and new Symfony components.

javiereguiluz:
I'm :-1: about this. The removal of the prefixed config files is one of the most essential features of Flex. Fabien worked hard designing a new config system which departs from the old Symfony conventions to simplify everything ... and you now are asking to complicate things again mixing old and new Symfony concepts.

fabpot:
That's already possible. Just put your file under the `dev/` sub-directory.

simensen:
@fabpot When discussing with @nicolas-grekas, I mentioned that putting stuff in `packages/ENV_NAME/` felt wrong. I felt like the `packages/` directory was reserved for packages.

It might be nice to make it clear that the `packages/ENV_NAME/` directories are fair game for anyone for any reason. I don't think I ever picked up on that intention. It was not at all intuitive to me that I should create new files in `packages/dev/services.yaml`.

I'm happy to close this if `packages/ENV_NAME/` is intended and expected.

Pierstoval:
Nothing prevents you from creating a `packages/dev/app.yaml`, your application is the "main" package after all, and it wouldn't conflict with anything.

nicolas-grekas:
I still agree with @simensen, having this file would really help - vs a nested one in `packages/%env%/services.yaml`

nesting means more difficult to grab, discover and communicate. Yet, the need is just very common.

javiereguiluz:
@nicolas-grekas splitting the config in small files per package + environment was an essential feature of flex and Symfony 4 since day one. We have tens and tens of recipes that rely on that behavior. We can't change that a few weeks before Flex/Symfony 4 are released.

As you know, Config was the Top 4 problem for Symfony developers according to our last community survey. Most developers are having problems with the existing config system. That's why Fabien proposed to change it entirely. As you can read on Symfony Slack channels, most people agree with this proposal.

Mixing the new config system with the old suffixed files would be a huge huge huge huge mistake in my opinion.

nicolas-grekas:
I'm not sure to get what you mean. Nobody suggested to change what's already been built and is really fine.

Let me ask you a question: where do you put test(or dev)-specific service configs for your app with Flex?
This is a very common need. Yet, right now, you have to put some file in `packages/%env%/services.yaml`. More importantly, you then don't have control over config ordering so that you cannot be sure you can override a specific service for eg testing.

I don't see where this proposal raises red flags. This is just about adding new files that are loaded last and have simple paths to define per-env specific config.

This need is not handled yet - because of the no-control-over-ordering issue.
Note also that this is *not* the same as what we do right now in SF-std: there would be *no* "import" at the beginning of these files. There would be just appended last to the list of loaded configs.

javiereguiluz:
I've chatted with Nicolas about this. We still don't agree, but I want to mention another potential problem that this proposal would introduce. If I'm right, this would break the best feature of Flex: automation.

If we allow people to put config in `services_dev.yaml`, `config_dev.yaml`, `doctrine_dev.yaml`, etc. files, then Flex won't be able to remove that config when you uninstall some package. So you must remove that config manually or the app will break.

nicolas-grekas:
The chat with @javiereguiluz was productive :)
So, just to be clear, this issue is *not* about have `doctrine_dev.yaml` style files at all.

But we raised this issue:
app do need config that relates to the services in `src/` - let's call that app-only config
that's params and services that relate only to what's inside `src/`
here, it's raw config
(no semantic one, we don't do bundles anymore for apps)
so, `services.yaml` is fine
then there is this need of per-env app-only config
OK?
having `services.yaml` at root of `config/`
and per-env config inside `packages/` is just not consistent
either `services.yaml` is moved inside `packages/`
or per-env app-only config can exist inside `config/`
right now, this is mixed, thus unintuitive as hell
and just not consistent

Pierstoval:
I was also thinking about proposing the removal of `packages/`, prefix third-party configuration and have a structure like this:

```
./
├── config
│   ├── dev/
│   │   ├── _framework.yaml
│   │   └── services.yaml
│   ├── prod/
│   ├── test/
│   │   └── services.yaml
│   ├── _framework.yaml
│   ├── _doctrine.yaml
│   ├── ...
│   └── services.yaml
├── src/
└── ...
```

This would guarantee that custom non-prefixed yaml files can still be decoupled in multiple files (sometimes, a single `services.yaml` is just not enough for bigger apps...) and loaded at the end of the loading process.

Another advantage is the removal of a directory level, which seems nice to me.

But not sure it's relevant nor useful for a DX perspective, and it would need a big change in current recipes (even if it's a simple renaming of all currently versioned yaml files), and maybe another explanation about third-party config vs app config...

WDYT?

nicolas-grekas:
@Pierstoval see https://github.com/symfony/flex/issues/141

symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -42,6 +42,7 @@ protected function configureContainer(ContainerBuilder $container, LoaderInterfa
             $loader->load($confDir.'/packages/'.$this->environment.'/**/*'.self::CONFIG_EXTS, 'glob');
         }
         $loader->load($confDir.'/services'.self::CONFIG_EXTS, 'glob');
+        $loader->load($confDir.'/services_'.$this->environment.self::CONFIG_EXTS, 'glob');
```
yceruto: What if puts the environment directories at root instead? it'll keep the same convention just to app-conf.
```
├── config
│   ├── test/
│   │   └── services.yaml
│   ├── packages/
│   │   └── dev/
│   └── services.yaml
```
```php
        if (is_dir($confDir.'/'.$this->environment)) {
            $loader->load($confDir.'/'.$this->environment.'/**/*'.self::CONFIG_EXTS, 'glob');
        }
```

symfony/framework-bundle/3.3/manifest.json
```diff
@@ -18,6 +18,7 @@
     },
     "gitignore": [
         ".env",
+        "/config/services_*.yaml",
```
yceruto: No sure why these file should be ignored :/ they are part of the repository, no?

symfony/framework-bundle/3.3/src/Kernel.php
Pierstoval: See symfony/flex#141 too, I think 😜

fabpot:
I'm in favor of this change as I don't see any other "simple" way to fix this issue. I'm not too concerned with this addition as it should be rare enough, and nobody forces you to use it. You can still so an explicit import if you like to.

PR #163, opened 2017-08-30
Default to non-debug in index.php

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

ping @stof

### DIFF
--- a/symfony/framework-bundle/3.3/public/index.php
+++ b/symfony/framework-bundle/3.3/public/index.php
@@ -12,7 +12,7 @@ if (!isset($_SERVER['APP_ENV'])) {
     (new Dotenv())->load(__DIR__.'/../.env');
 }
 
-if ($_SERVER['APP_DEBUG'] ?? true) {
+if ($_SERVER['APP_DEBUG'] ?? false) {
     // WARNING: You should setup permissions the proper way!
     // REMOVE the following PHP line and read
     // https://symfony.com/doc/current/book/installation.html#checking-symfony-application-configuration-and-setup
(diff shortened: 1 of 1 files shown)


symfony/framework-bundle/3.3/public/index.php
```diff
@@ -23,7 +23,7 @@
 
 // Request::setTrustedProxies(['0.0.0.0/0'], Request::HEADER_FORWARDED);
 
-$kernel = new Kernel($_SERVER['APP_ENV'] ?? 'dev', $_SERVER['APP_DEBUG'] ?? true);
+$kernel = new Kernel($_SERVER['APP_ENV'] ?? 'dev', $_SERVER['APP_DEBUG'] ?? false);
```
stof: Setting `APP_DEBUG` to `'0'` explicitly (which triggers non-debug mode in the console) will pass  a non-empty string here, and so be truthy.

And you also need to update line 15

nicolas-grekas: L15 updated

yceruto:
Should be updated `bin/console` too.

nicolas-grekas:
@yceruto nope, nothing to change in console

yceruto:
Yes, I mean, even though `true !== '0'` is `false` looks weird default `true` for non-debug, no?

symfony/framework-bundle/3.3/public/index.php
yceruto: Anyway `(bool) '0'` should be `false` in https://github.com/symfony/symfony/blob/edb8c33f35b981b72adbe23798753703edfe3819/src/Symfony/Component/HttpKernel/Kernel.php#L83

nicolas-grekas:
@yceruto the console has debug *on* by default - "true" hints this to me.

yceruto:
Oh I see, the previous behavior didn't change for `bin/console`, we're good then :)

symfony/framework-bundle/3.3/public/index.php
Tobion: The bool cast should be added here. Otherwise it's not passing the right type, esp. if we add type hints for sf 4.

nicolas-grekas: @Tobion we don't have strict type enabled so this is not required: php will do the cast for us based on the type hint.

PR #167, opened 2017-08-31
Uppercase the first letter in the printing message

bocharsky-bw:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/framework-bundle/3.3/Makefile
+++ b/symfony/framework-bundle/3.3/Makefile
@@ -16,7 +16,7 @@ cache-warmup: cache-clear
 ifdef CONSOLE
 	@$(CONSOLE) cache:warmup
 else
-	@printf "cannot warmup the cache (needs symfony/console)\n"
+	@printf "Cannot warm up the cache (needs symfony/console)\n"
 endif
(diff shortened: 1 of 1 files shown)


symfony/framework-bundle/3.3/Makefile
```diff
@@ -16,7 +16,7 @@ cache-warmup: cache-clear
 ifdef CONSOLE
 	@$(CONSOLE) cache:warmup
 else
-	@printf "cannot warmup the cache (needs symfony/console)\n"
+	@printf "Cannot warmup the cache (needs symfony/console)\n"
```
javiereguiluz: If I'm right, "warmup" word doesn't exist. It's `warm-up` as a noun and `warm up` as a verb, so in this case it should be "Cannot warm up the cache ..."

bocharsky-bw: 👍  for "warm up"

fabpot: @bocharsky-bw Can you make the change?

bocharsky-bw: Done

PR #168, opened 2017-09-01
Makefile: remove trailing dot on server:stop command

andrerom:
| Q             | A
| ------------- | ---
| License       | MIT


javiereguiluz:
There seems to be an issue with the recipe validator. @fabpot maybe we can whitelist this recipe or disable this validator temporarily? Thanks!

error-recipe

Issue #147, opened 2017-08-04
when the recipe api will be available in a stable version?


mappedinn:
Hi,

when the recipe api will be available in a stable version?

Actually, it can be installed only if `"minimum-stability"="dev"` as it can be seen in [the issue 140](https://github.com/symfony/recipes/issues/140). And this is causing me problem when traying to install the recipe `logger` and `cli`.

Thanks

fabpot:
Not sure what "recipe api" means in this context. Anyway, I'm going to answer the other issue as it's not related to flex itself.

stof:
The `api` alias references the api-platform project. And Flex support in api-platform is only available for 2.1+, which is not yet stable.

But the question about when api-platform 2.1 will be released should be asked to the api-platform maintainers. The recipes repo has no take on this date.

PR #170, opened 2017-09-06
Add default env vars for PHPUnit

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/phpunit/phpunit/4.7/phpunit.xml.dist
+++ b/phpunit/phpunit/4.7/phpunit.xml.dist
@@ -9,7 +9,11 @@
 >
     <php>
         <ini name="error_reporting" value="-1" />
-        <server name="KERNEL_CLASS" value="App\Kernel" />
+        <env name="KERNEL_CLASS" value="App\Kernel" />
+        <env name="APP_ENV" value="test" />
+        <env name="APP_DEBUG" value="1" />
+        <env name="APP_SECRET" value="s$cretf0rt3st" />
+        <!-- define your env variables for the test env here -->
     </php>
 
     <testsuites>
--- a/symfony/phpunit-bridge/3.3/phpunit.xml.dist
+++ b/symfony/phpunit-bridge/3.3/phpunit.xml.dist
@@ -9,7 +9,11 @@
 >
     <php>
         <ini name="error_reporting" value="-1" />
-        <server name="KERNEL_CLASS" value="App\Kernel" />
+        <env name="KERNEL_CLASS" value="App\Kernel" />
+        <env name="APP_ENV" value="test" />
+        <env name="APP_DEBUG" value="1" />
+        <env name="APP_SECRET" value="s$cretf0rt3st" />
+        <!-- define your env variables for the test env here -->
     </php>
 
     <testsuites>


fabpot:
Maybe these could be managed by Flex like done for the `.env.dist` file. Let me see that before merging.

phpunit/phpunit/4.7/phpunit.xml.dist
```diff
@@ -9,7 +9,11 @@
 >
     <php>
         <ini name="error_reporting" value="-1" />
-        <server name="KERNEL_CLASS" value="App\Kernel" />
+        <env name="KERNEL_CLASS" value="App\Kernel" />
```
xabbuh: Why using `env` here when we just started checking  `$_SERVER` (see e.g. #156)?

nicolas-grekas: because it's the correct concept, and maps to the .env file
_SERVER mixes several sources, so it's confusing

xabbuh: But the difference if I don't miss anything is that PHPUnit unline the Dotenv component does not populate the `$_SERVER` array with this config.

For example, take this config:

```xml
    <php>
        <server name="foo" value="bar" />
        <env name="foobar" value="baz" />
    </php>
```

It ends up with this output:

```
(
    [DOCUMENT_ROOT] => 
    [PATH_TRANSLATED] => ./phpunit
    [PHP_SELF] => ./phpunit
    [PWD] => /Users/christian/symfony
    [REQUEST_TIME] => 1504717866
    [REQUEST_TIME_FLOAT] => 1504717866.6028
    [SCRIPT_FILENAME] => ./phpunit
    [SCRIPT_NAME] => ./phpunit
    [TMPDIR] => /var/folders/9_/wkjkz_ms2hn7rkr4vm4b68t00000gn/T/
    [USER] => christian
    [_] => ./phpunit
    [argc] => 3
    [argv] => Array
        (
            [0] => ./phpunit
            [1] => --colors=always
            [2] => src/Symfony/Component/Filesystem/
        )

    [foo] => bar
)
Array
(
    [foobar] => baz
)
```

Produced by:

```php
ksort($_SERVER);
print_r($_SERVER);
ksort($_ENV);
print_r($_ENV);
```

yceruto:
Will be great that Flex comes into play here to update the `phpunit.xml.dist` file. Even, could we configure the default env vars in recipe for `phpunit.xml.dist`? It would allow to `doctrine-bundle` manifest to do something like this:
```json
{
    ...
    "env": {
        "DATABASE_URL": "\"mysql://root@127.0.0.1:3306/symfony?charset=utf8mb4&serverVersion=5.7\""
    },
    "env_phpunit": {
        "DATABASE_URL": "\"sqlite:///%kernel.project_dir%/var/test.db\""
    }
}
```
Then, the current changes could be moved to `framework-bundle` manifest.

phpunit/phpunit/4.7/phpunit.xml.dist
nicolas-grekas: That's stupid :)
Anyway, since `getenv('foo')` returns only from `<env>`, I think we have no choice but use it.

yceruto: Also this line would not meet its goal because: https://github.com/symfony/symfony/blob/d102fc08e48451aee932a8138d9cfaa726cee288/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L117.

nicolas-grekas: @yceruto good catch, can you send a PR to fix this? it should work!

yceruto: Not sure what I should fix here? you mean check both vars `$_SERVER` and `$_ENV` in `KernelTestCase`?

nicolas-grekas: yes, that's what I mean

fabpot:
Closing in favor of #147

fabpot:
Managing env vars in Flex is great, but as PHPUnit is **always** installed after the main packages, these env vars will never be there, so this PR is relevant after all.

fabpot:
RFR

yceruto:
> but as PHPUnit is always installed after the main packages, these env vars will never be there...

The same would happen if `phpunit` is installed after any package with env vars, right? So for the sake of not losing the previous env vars, could Flex review the current recipes installed and reconfigure this part? If possible, the order in which it was installed wouldn't matter.

phpunit/phpunit/4.7/phpunit.xml.dist
```diff
@@ -10,6 +10,10 @@
     <php>
         <ini name="error_reporting" value="-1" />
         <server name="KERNEL_CLASS" value="App\Kernel" />
+        <server name="APP_ENV" value="test" />
+        <server name="APP_DEBUG" value="0" />
```
yceruto: Both [`APP_ENV` and `APP_DEBUG` vars have no effect][1] on kernel initialization, so new feature or bugfix? ... or simply we can remove them? 

Also note that the current default value for debug mode is `true` in test env, here we are changing this to `false`.

  [1]: https://github.com/symfony/symfony/blob/7ab8efb74ecd422ed627753971637bd7d8dc92ca/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L192-L193

fabpot: Changed false to true for `APP_DEBUG`

fabpot: @yceruto I think we need to add support for these env vars. Can you submit a PR for Symfony 3.4 to use env vars when options are not defined explicitly, and keep the fallback as is if the env vars are not define?

yceruto: Sure, I'm on it.

fabpot: Great! Thank you

phpunit/phpunit/4.7/phpunit.xml.dist
```diff
@@ -9,7 +9,11 @@
 >
     <php>
         <ini name="error_reporting" value="-1" />
-        <server name="KERNEL_CLASS" value="App\Kernel" />
+        <env name="KERNEL_CLASS" value="App\Kernel" />
+        <env name="APP_ENV" value="test" />
+        <env name="APP_DEBUG" value="0" />
```
yceruto: Same here `<env name="APP_DEBUG" value="1" />`

fabpot: good catch

fabpot:
force

phpunit/phpunit/4.7/phpunit.xml.dist
yceruto: See https://github.com/symfony/symfony/pull/24177

PR #171, opened 2017-09-07
Public controllers

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Requires https://github.com/symfony/symfony/pull/24126

### DIFF
--- a/symfony/framework-bundle/3.3/config/services.yaml
+++ b/symfony/framework-bundle/3.3/config/services.yaml
@@ -21,9 +21,8 @@ services:
         # but if a service is unused, it's removed anyway
         exclude: '../src/{Entity,Repository,Tests}'
 
-    # controllers are imported separately to make sure they're public
-    # and have a tag that allows actions to type-hint services
+    # controllers are imported separately to make sure they
+    # have the tag that allows actions to type-hint services
     App\Controller\:
         resource: '../src/Controller'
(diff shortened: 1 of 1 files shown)


chalasr:
Merging now would make any controller private, right? Should this wait for the next 3.3 release of HttpKernel?

javiereguiluz:
Can someone please explain this change to me? I can't understand why we remove `public: true` here.

xabbuh:
With the changes done in symfony/symfony#24126, services tagged with `controller.service_arguments` will become public implicitly.

javiereguiluz:
@xabbuh thanks! I totally missed that change!

symfony/framework-bundle/3.3/config/services.yaml
```diff
@@ -21,9 +21,8 @@ services:
         # but if a service is unused, it's removed anyway
         exclude: '../src/{Entity,Repository,Tests}'
 
-    # controllers are imported separately to make sure they're public
-    # and have a tag that allows actions to type-hint services
+    # controllers are imported separately to make sure they have the tag
```
javiereguiluz: I'm not sure about the need to mention here that this tag internally makes these services public. Would a comment like this be better?

```
# controllers are imported separately to apply the tag that allows actions to type-hint services
```

PR #173, opened 2017-09-10
Add twig-extensions extensions as services

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

see symfony/symfony#22849

### DIFF
--- a/twig/extensions/1.0/config/packages/twig_extensions.yaml
+++ b/twig/extensions/1.0/config/packages/twig_extensions.yaml
@@ -1,0 +1,10 @@
+services:
+    _defaults:
+        public: false
+        autowire: true
+        autoconfigure: true
+
+    #Twig\Extensions\ArrayExtension: ~
+    #Twig\Extensions\DateExtension: ~
+    #Twig\Extensions\IntlExtension: ~
+    #Twig\Extensions\TextExtension: ~
--- a/twig/extensions/1.0/manifest.json
+++ b/twig/extensions/1.0/manifest.json
@@ -1,0 +1,6 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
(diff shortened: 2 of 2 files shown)


javiereguiluz:
This looks a bit confusing to me. Twig extensions are a different project/repo from Twig, so if you uncomment this, you'll see an error (unless you install extensions). And having this config if you don't use extensions, looks strange.

fabpot:
@javiereguiluz You're right. Perhaps we need to create a `twig_extensions.yml` file that is installed only when `twig/extensions` is installed. WDYT?

fabpot:
@javiereguiluz Does it look better?

twig/extensions/1.0/config/packages/twig_extensions.yaml
```diff
@@ -0,0 +1,10 @@
+services:
+    _defaults:
+        autowire: true
+        autoconfigure: true
+
+    #Twig\Extensions\ArrayExtension: ~
+    #Twig\Extensions\DateExtension: ~
+    #Twig\Extensions\I18nExtension: ~
```
stof: this one should not be included IMO, as it conflicts with the Symfony TranslatorExtension

twig/extensions/1.0/config/packages/twig_extensions.yaml
```diff
@@ -0,0 +1,10 @@
+services:
+    _defaults:
```
stof: should also have `public: false`

fabpot: fixed

twig/extensions/1.0/config/packages/twig_extensions.yaml
fabpot: removed

PR #176, opened 2017-09-12
checking for .env file exists instead of using the env variable

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

Hi guys!

This comes from a workshop I gave last week. One user - from using Laravel - already had an `APP_ENV` environment variable defined in their VM. Because of this, their `.env` file was being completely ignored, which was difficult to track down.

I think it's much simpler if we require the `.env` file simply if it exists. If you agree, there's one other change to make (https://github.com/symfony/flex/blob/d68d28624e3ec972d6887cca06f7df5e07ba03a9/src/Configurator/MakefileConfigurator.php#L34), which I can do.

Cheers!

### DIFF
--- a/behat/symfony2-extension/2.1/features/bootstrap/bootstrap.php
+++ b/behat/symfony2-extension/2.1/features/bootstrap/bootstrap.php
@@ -3,6 +3,6 @@
 use Symfony\Component\Dotenv\Dotenv;
 
 // The check is to ensure we don't use .env in production
-if (!isset($_SERVER['APP_ENV'])) {
+if (file_exists(__DIR__.'/../../.env')) {
     (new Dotenv())->load(__DIR__.'/../../.env');
 }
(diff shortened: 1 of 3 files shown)


stof:
We also need to make a change in Flex, as it would copy the `.env.dist` to `.env` on `composer install`, which would make the .env file to be loaded in production (which we don't want)

fabpot:
I'm very much against this change. It's a question of precedence and I don't want to change it. It's consistent with the behavior of other frameworks and we/I take advantage of this property in many projects. What we could do instead is having a CLI tool that list defined env vars  and tell the dev where they are coming from (env var, .env, $_SERVER, ...).

fabpot:
see symfony/symfony#24144

weaverryan:
Hmm. Ok. Then we'll solve it in other ways - i.e. finding ways to be a bit more verbose about where environment vars are coming from, or if the `.env` file is loaded or not. We need to try to make it as obvious as possible when you have a `.env` file, but it is not being used (especially if you're in the debug mode where we can be more verbose with things).

PR #179, opened 2017-09-15
Change the way we register annotations

### LINKED ISSUE #118: [RFC] Improve the way annotations are enabled
In a Symfony Flex app, I expected to enable annotations by running this:

```
$ composer require annotations
```

However, that installs the `symfony/annotation-pack`, which only installs the low-level libraries needed for annotations (doctrine/cache and doctrine/annotations) but it doesn't really enable annotations in the application.

---

First, this was a problem because it crashed my browser, [as reported here](https://github.com/symfony/symfony/issues/23566).

Second, even if this comment from `config/routing.yaml` is perfectly clear:

```
# Depends on sensio/framework-extra-bundle, doctrine/annotations, and doctrine/cache
#   install with composer req sensio/framework-extra-bundle annot
```

It's a bit disappointing that `composer require annotations` doesn't enable annotations and you must also execute `composer require sensio/framework-extra-bundle`

-----

Solutions:

* Maybe we should create a better recipe for annotations?
* Maybe we should add annotations support in frameworkbundle and forget about the framework extra bundle?

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

Several changes here:

 * We bump the minimal version of framework extra bundle to 4.0, which is the only one to not depend on the `templating` component.

 * The `annotations` alias (and its derivative are now on framework extra bundle directly) as aliases are for Symfony integration. When requiring annotations, we want to use them in the framework, and for controllers (fixes #118).

 * Finally, the `annotations-pack` is removed as framework extra bundle requires `doctrine/common`, which in turn requires `doctrine/annotations` and `doctrine/cache`. This last piece looks suspicious as we clearly documented in `routes.yaml` to require both annotations and the bundle, but I don't understand why this is needed. At least, I've just tried and there is no need to do that.

To sum up, enabling annotations in controllers is now just about doing `composer req annot`.


fabpot:
https://github.com/symfony/annotations-pack should be removed after merging this PR.

javiereguiluz:
I tested this as follows:

```
$ composer create symfony/skeleton my_project
```

Create the controller with annotations:

```yaml
# config/routes.yaml
controllers:
    resource: ../src/Controller/
    type: annotation
```

```php
// src/Controller/DefaultController.php
namespace App\Controller;

use Symfony\Component\HttpFoundation\Response;
use Symfony\Component\Routing\Annotation\Route;

class DefaultController
{
    /** @Route("/") */
    public function index()
    {
        return new Response('Hello World!');
    }
}
```

And nothing works. Then, install `annot` package:

```
$ export SYMFONY_ENDPOINT=https://symfony.sh/r/github.com/symfony/recipes/179

$ composer req annot

  Package operations: 6 installs, 0 updates, 0 removals
    - Installing doctrine/lexer (v1.0.1): Loading from cache
    - Installing doctrine/inflector (v1.2.0): Loading from cache
    - Installing doctrine/collections (v1.5.0): Loading from cache
    - Installing doctrine/annotations (v1.5.0): Loading from cache
    - Installing doctrine/common (v2.8.1): Loading from cache
    - Installing sensio/framework-extra-bundle (v3.0.27): Loading from cache
```

Reload the page and everything works!

---

About the suspicious comment in `symfony/routing/3.3/config/routes.yaml`, I can't find the reason why we added it ... but it can be safely removed.

adamelso:
There's a problem we encountered with the recipe providing the `resource: ../src/Controller/` config for annotations - it assumes the user's project has kept the `App` namespace to autoload from `src/`.

When we used the Doctrine ORM recipe for installation, the hook that clears the cache after running `composer require orm` failed spectacularly with a misleading error message - indicating our current config was incorrect rather than the one provided by the recipe.

Should the config be provided commented-out? Or are we recommended not the use the recipe in this case? Also I'm not sure which project to report this problem.

weaverryan:
Hey there!

Could you create an issue (on this repo) with these details? It seems a bit weird to me - I wouldn’t have thought autoloading would have caused an issue. Include the error message you got :). 

Cheers!

Tobion:
I think we should reconsider making FrameworkExtraBundle an alias for annotations with https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/562

webdevilopers:
I can confirm the issue mentioned by @adamelso . We moved our `App` namespace to `src/App`.
It looks like the `annotations.yaml` was added after we required `orm-pack`.

PR #180, opened 2017-09-15
Remove unneeded configuration

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

As of 3.3, Symfony automatically registers the annotations when installed.


fabpot:
@javiereguiluz You've tested already in the other PR, where the recipe was not installed as you installed framework extra version 3 and the recipe is for version 4 :)

PR #182, opened 2017-09-20
[FrameworkBundle] Add a env vars to configure trusted proxies and hosts

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

Add two convenient environment variables to set trusted proxies and hosts: `APP_TRUSTED_PROXIES` and `APP_TRUSTED_HOSTS`.

This very convenient when deploying a Symfony app to a cloud platform using load balancers (exemple: Google Container Engine) and/or when using Docker for local development (ex: HTTP/2 reverse proxy, Varnish...). For an example of real world usage see api-platform/api-platform#422.

Backported from api-platform/api-platform#422

### DIFF
--- a/symfony/framework-bundle/3.3/manifest.json
+++ b/symfony/framework-bundle/3.3/manifest.json
@@ -14,8 +14,8 @@
     "env": {
         "APP_ENV": "dev",
         "APP_SECRET": "%generate(secret)%",
-        "#APP_TRUSTED_PROXIES": "127.0.0.1,127.0.0.2",
-        "#APP_TRUSTED_HOSTS": "localhost,example.com"
+        "#TRUSTED_PROXIES": "127.0.0.1,127.0.0.2",
+        "#TRUSTED_HOSTS": "localhost,example.com"
     },
     "gitignore": [
         ".env",
--- a/symfony/framework-bundle/3.3/manifest.json
+++ b/symfony/framework-bundle/3.3/manifest.json
@@ -13,7 +13,9 @@
     },
     "env": {
         "APP_ENV": "dev",
-        "APP_SECRET": "%generate(secret)%"
+        "APP_SECRET": "%generate(secret)%",
+        "#APP_TRUSTED_PROXIES": "127.0.0.1,127.0.0.2",
+        "#APP_TRUSTED_HOSTS": "localhost,example.com"
     },
     "gitignore": [
         ".env",
--- a/symfony/framework-bundle/3.3/public/index.php
+++ b/symfony/framework-bundle/3.3/public/index.php
@@ -21,11 +21,11 @@ if ($_SERVER['APP_DEBUG'] ?? ('prod' !== ($_SERVER['APP_ENV'] ?? 'dev'))) {
     Debug::enable();
 }
 
-if ($trustedProxies = $_SERVER['APP_TRUSTED_PROXIES'] ?? false) {
+if ($trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? false) {
     Request::setTrustedProxies(explode(',', $trustedProxies), Request::HEADER_X_FORWARDED_ALL ^ Request::HEADER_X_FORWARDED_HOST);
 }
 
-if ($trustedHosts = $_SERVER['APP_TRUSTED_HOSTS'] ?? false) {
+if ($trustedHosts = $_SERVER['TRUSTED_HOSTS'] ?? false) {
     Request::setTrustedHosts(explode(',', $trustedHosts));
 }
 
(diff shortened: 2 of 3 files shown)


fabpot:
duplicate of #111, not a big fan of the `json_decode` part

dunglas:
`json_decode` is in sync with `symfony/symfony#23901` capabilities. It will be weird to have some arrays encoded as a CSV-like structure and some others as JSON IMO.

vonalbert:
I'm wondering if filter_input() could be used instead of accessing to the superglobal $_SERVER variable

Pierstoval:
Why not use `getenv()` instead?

dunglas:
For thread safety.

stof:
@vonalbert accessing the superglobal is one of the way to read environment variables in PHP.

`filter_input` is not necessary here (and it also accesses the superglobal content, but hides it). Thus, I saw comments on php.net saying that `filter_input(INPUT_SERVER, ...)` used to cause some issues with FastCGI (maybe the same thread safety issues that `getenv`...)

vonalbert:
@stof thanks, wasn't aware of those issues, good to know. For what I read, `filter_input()` is problematic only if called with `INPUT_SERVER` or `INPUT_ENV` at the first arg..

Anyway, the sole reason for which I proposed using that function is because `filter_input()` would remove the need to check the existance of the keys performed with the `??` operator and, maybe, improve readability, just that :)

dunglas:
Checking the existence of the key is also faster than executing a function. And I like the ?? style but it’s just a matter of taste 😊

symfony/framework-bundle/3.3/public/index.php
```diff
@@ -21,7 +21,13 @@
     Debug::enable();
 }
 
-// Request::setTrustedProxies(['0.0.0.0/0'], Request::HEADER_FORWARDED);
+if ($trustedProxies = $_SERVER['APP_TRUSTED_PROXIES'] ?? false) {
+    Request::setTrustedProxies(json_decode($trustedProxies, true), Request::HEADER_X_FORWARDED_ALL);
```
nicolas-grekas: The `Request::HEADER_X_FORWARDED_ALL` would also need env-based configuration.
Can we autodetect AWS btw?

dunglas: Should we allow to set the list of headers to trust in a JSON or just let the user set the constant value directly?
How can we detect AWS securely (without letting an attacker usurp its identity in case the app run elsewhere)?

nicolas-grekas: let's use HEADER_X_FORWARDED_AWS_ELB by default! it's generally enough, and the most portable

nicolas-grekas: of be more explicit:
Request::HEADER_X_FORWARDED_ALL & ^Request::HEADER_X_FORWARDED_HOST

symfony/framework-bundle/3.3/public/index.php
```diff
@@ -21,7 +21,13 @@
...
+}
+
+if ($trustedHosts = $_SERVER['APP_TRUSTED_HOSTS'] ?? false) {
+    Request::setTrustedHosts(json_decode($trustedHosts, true));
```
nicolas-grekas: I would suggest to move this in 
 symfony/framework-bundle/3.3/config/packages/framework.yaml now that we can json-decode env vars

symfony/framework-bundle/3.3/public/index.php
```diff
@@ -21,7 +21,13 @@
     Debug::enable();
 }
 
-// Request::setTrustedProxies(['0.0.0.0/0'], Request::HEADER_FORWARDED);
+if ($trustedProxies = $_SERVER['APP_TRUSTED_PROXIES'] ?? false) {
```
nicolas-grekas: `$_ENV['TRUSTED_PROXIES'] ?? $_SERVER['TRUSTED_PROXIES'] ?? false`

symfony/framework-bundle/3.3/manifest.json
```diff
@@ -13,7 +13,9 @@
     },
     "env": {
         "APP_ENV": "dev",
-        "APP_SECRET": "%generate(secret)%"
+        "APP_SECRET": "%generate(secret)%",
+        "#APP_TRUSTED_PROXIES": "127.0.0.1,127.0.0.2",
+        "#APP_TRUSTED_HOSTS": "localhost,example.com"
```
fabpot: I would remove the `APP_` prefix

dunglas: Ok, but other vars below are prefixed by `APP` (and I think it's a common practice to prefix env vars: https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-custom-environment-variables)

fabpot: I think it makes sense for "core" env var names as they are quite generic (DEBUG, ENV, ... would be too generic), but for these ones, I don't think we should fear a conflict.

PR #183, opened 2017-09-20
WIP: OverblogGraphQLBundle recipe

renatomefi:
| Q             | A
| ------------- | ---
| License       | MIT

### disclaimer
I'm not maintainer of this package

### info
I consider this bundle is worth to be in the main recipes, if you disagree I'll gladly open it in the recipes-contrib

### todo
- [x] Wait on https://github.com/overblog/GraphQLBundle/pull/191 for smoother installation

### DIFF
--- a/overblog/graphql-bundle/0.9/config/routes/dev/graphql_graphiql.yaml
+++ b/overblog/graphql-bundle/0.9/config/routes/dev/graphql_graphiql.yaml
@@ -1,0 +1,2 @@
+overblog_graphql_graphiql:
+    resource: "@OverblogGraphQLBundle/Resources/config/routing/graphiql.yml"
--- a/overblog/graphql-bundle/0.9/config/routes/graphql.yaml
+++ b/overblog/graphql-bundle/0.9/config/routes/graphql.yaml
@@ -1,0 +1,2 @@
+overblog_graphql_endpoint:
+    resource: "@OverblogGraphQLBundle/Resources/config/routing/graphql.yml"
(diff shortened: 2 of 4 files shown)


stof:
I suggest adding it to recipes-contrib first. I'm not sure we want to make this an official recommendation yet (this is the first time I heard about this package, and I don't know what its quality is)

javiereguiluz:
@renatomefi thanks for contributing this recipe! Everything is technically correct, but as @stof said, we first add recipes to the https://github.com/symfony/recipes-contrib repository instead of this one. In the future, if the recipe gets popular, we promote it to this repository. Thanks!

renatomefi:
@stof @javiereguiluz sure, thank you for the feedback and I hope the bundle makes it to the main recipes soon!

Follow up here: https://github.com/symfony/recipes-contrib/pull/85

omarsoufiane:
@stof @javiereguiluz 
this bundle is definitely one of my favorites in symfony, 
symfony is all about building maintainable apps, and graphql's approach compared to rest is much more maintainable;
the bundle uses many symfony's great features like the PropertyAccess component to automatically resolve the graphql tree from entities/document objects ;
it uses the expression language on specific fields of the graphql shema which is something that doesn't exist on any other graphql librairy in other languages,
it supports .yml files for graphql types (with expression language in yml )
and again this is something that would work fine only in a compiler di container,
performance wise it can be paired to a dataloader bundle wich can be a replacement for varnish since graphal request are POST only, 
so yes this bundle rocks and i believe it will gain more popularity than api_platform in the  future.

PR #186, opened 2017-09-22
Fix kernel name

ro0NL:
| Q             | A
| ------------- | ---
| License       | MIT

`bin/console about`

Before: `Kernel name: src`
After `Kernel name: app`

### DIFF
--- a/symfony/framework-bundle/3.3/src/Kernel.php
+++ b/symfony/framework-bundle/3.3/src/Kernel.php
@@ -14,6 +14,11 @@ class Kernel extends BaseKernel
 
     private const CONFIG_EXTS = '.{php,xml,yaml,yml}';
 
+    public function getName(): string
+    {
+        return 'app';
+    }
+
     public function getCacheDir(): string
     {
         return dirname(__DIR__).'/var/cache/'.$this->environment;


fabpot:
I would instead remove the Kernel name of the command.

ro0NL:
That does not fix the bug :) (agree though; this is not really useful info). Added it for consistency with web profiler, but the value doesnt change; which was the purpose of the command.

Ill look into that. For this PR; isnt it straightforward like this? Do we really care about auto-discovery here, today?

fabpot:
There is no bug. I would really prefer to remove the information, and perhaps remove this `getName()` method which is not really useful anyway. Perhaps deprecate it in 3.4 and remove it in 4.0... which would mean no patch here :)

ro0NL:
Removed kernel type/name in https://github.com/symfony/symfony/pull/24144 (3.4)

> perhaps remove this getName() method which is not really useful anyway.

:+1: 

Also related, KernelInterface defines getLogDir, getCacheDir and such. However `getParameter('kernel.cache_dir')` is kinda source of truth. Should we get rid of this getters in the interface? So it's consistent with not defining getProjectDir as well :)

PR #187, opened 2017-09-22
Add env var to specify the dotenv file to use

jmwri:
| Q | A |
|-- | -- |
|License | MIT |

I have a use case where I will test different environments locally, e.g. dev, uat, test etc. It would be helpful to be able to store the env vars in different files rather having to constantly update one file.

### DIFF
--- a/symfony/framework-bundle/3.3/public/index.php
+++ b/symfony/framework-bundle/3.3/public/index.php
@@ -9,7 +9,8 @@ require __DIR__.'/../vendor/autoload.php';
 
 // The check is to ensure we don't use .env in production
 if (!isset($_SERVER['APP_ENV'])) {
-    (new Dotenv())->load(__DIR__.'/../.env');
+    $envFile = $_SERVER['ENV_FILE'] ?? '.env';
+    (new Dotenv())->load(__DIR__.'/../' . $envFile);
 }
 
 if ($_SERVER['APP_DEBUG'] ?? false) {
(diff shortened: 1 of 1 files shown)


symfony/framework-bundle/3.3/public/index.php
```diff
@@ -9,7 +9,8 @@
 
 // The check is to ensure we don't use .env in production
 if (!isset($_SERVER['APP_ENV'])) {
-    (new Dotenv())->load(__DIR__.'/../.env');
+    $envFile = $_SERVER['ENV_FILE'] ?? '.env';
+    (new Dotenv())->load(__DIR__.'/../' . $envFile);
```
fabpot: We don't add spaces around the dot.

fabpot:
I don't like this change. For tests, this should be configured via phpunit.xml.dist if PHPUnit is used. For CI, generating the right `.env` file sounds easier. So, I'm not sure there are many uses cases for such a change.

jmwri:
I was unaware that it was possible to set env vars in phpunit.xml. Closing the PR.

Pierstoval:
What about using phpunit with behat both locally and in CI, as well as Insight (which has support for `.env`, but still, these files have to be generated)? Multiple env files at the root directory might be really helpful when dealing with different kind of environments, even for testing

fabpot:
As you say, those files are anyway generated by "something". I don't think we need to support more than one file at a time.

Pierstoval:
It's true, these files are generated "by something" but allowing different env files would avoid having to generate dirty hacks to inject env vars to other envs, and having explicit environment-specific configuration in one file.

fabpot:
Not sure I understand, but feel free to change your local version if you have very specific needs. But as I see it right now, there is no need to make such a change in the generic recipe.

PR #188, opened 2017-09-22
Disable debug by default in CLI

### LINKED ISSUE #184: Different default debug value between CLI/web
https://github.com/symfony/recipes/blob/master/symfony/console/3.3/bin/console#L25 (`$_SERVER['APP_DEBUG'] ?? true`)
vs. 
https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/public/index.php#L15 (`$_SERVER['APP_DEBUG'] ?? false`)

Is that intended?

Also id suggest a default value of "1" or "0". This mixes string/boolean type.

ro0NL:
| Q             | A
| ------------- | ---
| License       | MIT

Fixes #184 

cc @fabpot

### DIFF
--- a/symfony/console/3.3/bin/console
+++ b/symfony/console/3.3/bin/console
@@ -22,7 +22,7 @@ if (!isset($_SERVER['APP_ENV'])) {
 
 $input = new ArgvInput();
 $env = $input->getParameterOption(['--env', '-e'], $_SERVER['APP_ENV'] ?? 'dev');
-$debug = ($_SERVER['APP_DEBUG'] ?? true) !== '0' && !$input->hasParameterOption(['--no-debug', '']);
+$debug = ($_SERVER['APP_DEBUG'] ?? false) && !$input->hasParameterOption(['--no-debug', '']);
 
 if ($debug && class_exists(Debug::class)) {
     Debug::enable();


stof:
As we have a `--no-debug` option but no `--debug` option, defaulting to debug mode makes sense.

fabpot:
which makes me think that we should probably deprecate this `--no-debug` option and the `--env` option as well. People can just use env vars: `ENV_DEBUG=1 ./bin/console ...` (does that work on Windows?).

stof:
Well, on Windows, I know you can set it before. I'm not sure Windows has a feature to set an env variable only for the duration of a command

Pierstoval:
Using the `ENV_DEBUG=1 php bin/console` in Windows is possible with a certain syntax, but it's so ugly no one would probably ever use it (check [this SO thread if you like](https://superuser.com/questions/223104/setting-and-using-variable-within-same-command-line-in-windows-cmd-exe)).

Instead of `--no-debug` I think it's better to use `--debug[=0|1]` instead, so we can have something more readable and "CLI customizable" like `php bin/console ... --env=prod --debug=0`. And maybe using `--debug` would be similar to `--debug=1` so we won't have to write it unless we want to deactivate it via `--debug=0`.

WDYT?

nicolas-grekas:
I think the default should be changed in index.php:
if there is no `.env` and no real env vars, the config should either be prod+no-debug or dev+debug,
but no dev+no-debug as index.php is now.

The default in this situation for both index.php+console should be prod+no-debug IMHO.

dunglas:
> The default in this situation for both index.php+console should be prod+no-debug IMHO.

👍

nicolas-grekas:
see https://github.com/symfony/recipes/pull/239

PR #192, opened 2017-09-23
Allow PHP 7.0

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

See reasoning in https://github.com/symfony/flex/pull/162.

Does the recipes CI check that PHP files are syntactically valid btw?

### DIFF
--- a/symfony/framework-bundle/3.3/src/Kernel.php
+++ b/symfony/framework-bundle/3.3/src/Kernel.php
@@ -12,7 +12,7 @@ class Kernel extends BaseKernel
 {
     use MicroKernelTrait;
 
-    private const CONFIG_EXTS = '.{php,xml,yaml,yml}';
+    const CONFIG_EXTS = '.{php,xml,yaml,yml}';
 
     public function getCacheDir(): string
     {


stof:
this Pr is totally incomplete. The `void` and `iterable` typehints (used in the same file) are also breaking compat with PHP 7.0

fabpot:
Well, that's the issue with auto-merged PRs I suppose :) Anyway, we need someone who can really move this story forward. Do we really want to support 7.0? If yes, it looks like the work is far from being done.

aik099:
How it's even possible to support PHP 7.0.x when Symfony 4.x will require PHP 7.1.3+ and Symfony Flex is expected to be used mostly with Symfony 4.x?

stof:
@aik099 Flex is supported by Symfony 3.4 too (even though Symfony 3.4 still support the standard-edition)

aik099:
I see. Then we really need to run `php -l` as part of the build to catch all such PHP syntax errors early.

stof:
But we need to run them based on the min supported version.

@fabpot is it something that can be done by the CI ? I know you are getting the Packagist metadata, so you may be able to find the lowest PHP version supported by a recipe (I would only care about finding the minor PHP release, I'm fine using latest PHP 7.0 for the linting, as PHP patch releases should not change syntax).

@aik099 a linting would not have catched `iterable` and `void`. The syntax was valid in 7.0 though, but referred to class names (even though these names were meant to be considered reserved by users).

fabpot:
@stof linting is probably possible, but I would need to install a bunch of PHP versions, something I don't have right now. Not sure if this is worth it though as PHP files in recipes is pretty rare.

Pierstoval:
Linting is important only for php 7.0 compatible recipes, you could even do this with a simple docker image used as command line instruction

xabbuh:
@Pierstoval A recipe could be for a package that requires PHP 7.1 or higher. In these recipes, it's totally fine to use newer PHP features for which a PHP 7.0 linter would fail.

PR #195, opened 2017-09-25
Fix PHP 7.0

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

There is no other 7.1 specific code in this repo, and there is none at all in recipes-contrib.


stof:
Do we want to create a new recipe for FrameworkBundle 4.0, which would leverage PHP 7.1 ?

PR #199, opened 2017-10-01
Updated the admin recipe for the new version of the bundle

javiereguiluz:
| Q             | A
| ------------- | ---
| License       | MIT

From 1.16 to 1.17 we updated the namespace from `JavierEguiluz\...` to `EasyCorp\...`. In the coming days/weeks we'll also move the repository vendor from `javiereguiluz/...` to `easycorp/...` so I'll need to create a new recipe under a different vendor.

### DIFF
--- a/javiereguiluz/easyadmin-bundle/1.16/manifest.json
+++ b/javiereguiluz/easyadmin-bundle/1.16/manifest.json
@@ -4,6 +4,5 @@
     },
     "copy-from-recipe": {
         "config/": "%CONFIG_DIR%/"
-    },
-    "aliases": ["admin-gen", "admin-generator", "admin"]
+    }
 }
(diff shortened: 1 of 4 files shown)


xabbuh:
@fabpot Is it expected that only one version of the package can define the aliases?

fabpot:
@xabbuh aliases are independent of the version. We should probably move them outside of the manifest.

fabpot:
For the record, this is how it works right now: aliases are defined in the main manifest file for a recipe (`symfony/console/1.0/manifest.json`, which means that it is part of one version but really work for all versions -- also, aliases must only be defined in only one version to make things more discoverable).

We could change it to either:

 * Move aliases definition under the main package directory by introducing a manifest that defines configuration for a package independently of its version (`symfony/console/manifest.json`; it would only support `aliases` for now);

 * Move aliases definition under a main `aliases.json` file at the root of the repository.

But I think the current way is "best". Having to looks for a global file to get aliases could be nice, but I like to keep all configuration for a package at the same place, and having a non-versioned manifest sounds like too much complexity added for just managing aliases (can be revisited if we add something that needs to be configured independently of the version).

PR #202, opened 2017-10-03
fix CONSOLE path setting

ElectricMaxxx:
| Q             | A
| ------------- | ---
| License       | MIT

This fix would fix the issue described in https://github.com/symfony/skeleton/issues/23 on my (ubuntu 16.04) mashine. No clue if there is a more elegant way to do it, but this one works. The former line gave me something equal to an empty string, but it was no empty string. Strange.


nicolas-grekas:
(but doesn't pass the CI)

nicolas-grekas:
For reference, the issue is that `sh` is used, and `%q` exists only on bash.

nicolas-grekas:
See #203 as alternative

ElectricMaxxx:
What is behind that CI test?

fabpot:
closing in favor of #203

PR #200, opened 2017-10-02
Fix space issue in bin/console

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

PR #203, opened 2017-10-03
Fix Makefile

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Alternative to #202, reverts #200 and #201.

### DIFF
--- a/symfony/framework-bundle/3.3/Makefile
+++ b/symfony/framework-bundle/3.3/Makefile
@@ -1,31 +1,31 @@
-CONSOLE := $(shell printf %q "`which bin/console`")
+CONSOLE := $(shell which bin/console)
 sf_console:
-ifeq ($(CONSOLE), '')
+ifndef CONSOLE
 	@printf "Run \033[32mcomposer require cli\033[39m to install the Symfony console.\n"
 endif
 
 cache-clear:
-ifneq ($(CONSOLE), '')
-	@$(CONSOLE) cache:clear --no-warmup
+ifdef CONSOLE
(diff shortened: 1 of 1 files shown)


symfony/framework-bundle/3.3/Makefile
```diff
@@ -1,31 +1,31 @@
...
 .PHONY: cache-warmup
 
 serve_as_sf: sf_console
-ifeq ($(CONSOLE), '')
+ifndef CONSOLE
 	@${MAKE} serve_as_php
 endif
-	@$(CONSOLE) | grep server:start > /dev/null || ${MAKE} serve_as_php
-	@$(CONSOLE) server:start
+	@bin/console list | grep server:start > /dev/null || ${MAKE} serve_as_php
```
nicolas-grekas: let's not assume `list` is the default command

symfony/framework-bundle/3.3/Makefile
```diff
@@ -1,31 +1,31 @@
-CONSOLE := $(shell printf %q "`which bin/console`")
+CONSOLE := $(shell which bin/console)
```
fabpot: That's what we had before, but that does not work well if there is a space in your directory names.

fabpot: ah, sorry, I see that you've changed the code using $CONSOLE as well. nvm

ElectricMaxxx: Why not `test -e` when using it bool'sch only?

PR #204, opened 2017-10-04
[DoctrineBundle] resolve parameters inside DATABASE_URL

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Not ready to merge yet, because this relies on DI >=3.4.
If we want to allow this with 3.3, we might patch DoctrineBundle.
But before doing so, what about just moving to require >=3.4 for all Symfony components?
Any other idea?

### DIFF
--- a/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
@@ -1,6 +1,7 @@
 doctrine:
     dbal:
-        url: '%env(DATABASE_URL)%'
+        # With Symfony 3.3, remove the `resolve:` prefix
+        url: '%env(resolve:DATABASE_URL)%'
     orm:
         auto_generate_proxy_classes: '%kernel.debug%'
         naming_strategy: doctrine.orm.naming_strategy.underscore


aik099:
Actually I don't like `DATABASE_URL` as single DSN for database, because it only works with Doctrine and doesn't work with PdoSessionHandler, that expects DSN in different format and username/password as separate variables.

I wish we can revert back to use 4 env variables as in Symfony 3.3 (that's what I did for now in my project to avoid specifying DB credentials in 2 different formats).

javiereguiluz:
Could we fix PdoSessionHandler to support this new DSN format instead?

aik099:
The DSN format isn't new, but isn't standard PDO DSN either. Not sure how to do this without breaking BC, but I see 2 options:

1. either duplicate Doctrine DSN parser in `PdoSessionHandler`

pros: PdoSessionHandler works with Doctrine bundle not installed
cons: code duplication

2. or make `PdoSessionHandler` literally share connection that Doctrine creates

pros: you will see DB queries made by PdoSessionHandler in Web Profiler
cons: PdoSessionHandler would depend on DoctrineBundle

nicolas-grekas:
@aik099 please open an issue if you want to discuss thus, that's another topic.

aik099:
Created #206.

aik099:
Will the `resolve:` prefix work for Symfony 3.3 as well?

fabpot:
We need to decide what to do here. I think we should probably just merge as is and let 3.3 users behind.

weaverryan:
I don't see any other option. We could temporarily (until 3.4 is released) add a comment above this line:

> For Symfony 3.3, remove the "resolve:" part: parameters cannot live inside environment vars

fabpot:
Problem is that cache:clear will fail immediately as the syntax is not valid in 3.3.

fabpot:
So?

doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
```diff
@@ -1,6 +1,6 @@
 doctrine:
     dbal:
-        url: '%env(DATABASE_URL)%'
+        url: '%env(resolve:DATABASE_URL)%'
```
fabpot: Can you add a comment for people using 3.3? Something like `# If you are using Symfony 3.3, remove the resolve: prefix`

PR #209, opened 2017-10-09
Use the native PHP session handler by default

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

In 2015, we changed from the native PHP session handler to the file based one, which stores in `var/session`. Here is the history: https://github.com/symfony/symfony-standard/pull/904

In that PR, we (imo) narrowly decided to do that (versus keeping with the native session handler), and one big reason was because the SE had a `var/session` dir (the argument being... if we have that dir, shouldn't we use it?). @xabbuh also had one argument about conflicting sessions with multiple PHP apps on one machine (https://github.com/symfony/symfony-standard/pull/904#issuecomment-163192748). Is that actually a real issue?

There are 2 big advantages to using the native PHP session handler:
* This is one less directory that you need to worry about making writeable, making deploying and local development less error-prone
* To make it writeable, in practice, people will commonly deploy and make `var/sessions` 777. This is not ideal security. PHP's native file session handler automatically handles this in a more secure way.

Cheers!

### DIFF
--- a/symfony/framework-bundle/3.3/config/packages/framework.yaml
+++ b/symfony/framework-bundle/3.3/config/packages/framework.yaml
@@ -6,7 +6,7 @@ framework:
     #trusted_hosts: ~
     # https://symfony.com/doc/current/reference/configuration/framework.html#handler-id
     #session:
-    #    The native PHP session handler will be used
+    #    # The native PHP session handler will be used
     #    handler_id: ~
     #esi: ~
     #fragments: ~
--- a/symfony/framework-bundle/3.3/config/packages/framework.yaml
+++ b/symfony/framework-bundle/3.3/config/packages/framework.yaml
@@ -6,8 +6,8 @@ framework:
     #trusted_hosts: ~
     # https://symfony.com/doc/current/reference/configuration/framework.html#handler-id
     #session:
-    #    handler_id: session.handler.native_file
-    #    save_path: '%kernel.project_dir%/var/sessions/%kernel.environment%'
+    #    The native PHP session handler will be used
+    #    handler_id: ~
     #esi: ~
     #fragments: ~
     php_errors:


xabbuh:
> @xabbuh also had one argument about conflicting sessions with multiple PHP apps on one machine (symfony/symfony-standard#904 (comment)). Is that actually a real issue?

Well, it means that the garbage collector may destroy sessions of your application just because you have different expiration settings in another PHP application running on the same server.

fabpot:
@xabbuh I agree but that's not Symfony specific. It would happen with any other PHP application. I think going standard whenever we can is a good idea.

aik099:
But the session configuration settings in `yml` file, that override `php.ini` defined ones would still work, right? E.g. `cookie_lifetime`, `gc_maxlifetime`?

weaverryan:
I believe so, yes: https://github.com/symfony/symfony/blob/473911b091cadbaa1566a31d90f508479093f1f1/src/Symfony/Component/HttpFoundation/Session/Storage/NativeSessionStorage.php#L326

xabbuh:
fair enough, but let's clarify the documentation then to avoid confusion: symfony/symfony-docs#8498

weaverryan:
Thanks @xabbuh :)

Tobion:
Why are we not using sys tmp dir for all caches then? Either we say things are configured per app or not. Now it's inconsistent again.

Tobion:
It's fine eithe way to me.

weaverryan:
So more accurately, the user is responsible to secure their php installation just like with any php app :)

teohhanhui:
Actually, if you look at the code there's no difference with using var/sessions. Not using var/sessions means that by default /tmp is used. All else being equal, it means we've just introduced the possibility of PHP session collision.

PR #210, opened 2017-10-09
Always using umask(0000) in debug mode

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

In debug mode, cache files need to be writeable by everyone for a good dev experience. In `prod`, they do not (and the way the cache system is built these days makes this very easy - yay!).

This change is consistent with how `index.php` already exists: enable `umask` for debug, but not otherwise. This gives great DX for dev, and forces users to setup permissions properly for prod.

I've also proposed removing the warning about permissions... basically because if this is for debug mode only, is there really an issue with `umask(0000)`? There are many things in debug mode you would not want to do/expose in prod mode (e.g. web debug toolbar). This is just another one of those.

P.S. I wasn't sure why there is a `class_exists(Debug::class)` in `console`, but not in `index.php`, but I left that alone.

### DIFF
--- a/symfony/console/3.3/bin/console
+++ b/symfony/console/3.3/bin/console
@@ -7,7 +7,6 @@ use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Debug\Debug;
 use Symfony\Component\Dotenv\Dotenv;
 
-umask(0000);
 set_time_limit(0);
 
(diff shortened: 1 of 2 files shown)


stof:
@weaverryan the reason there is no class_exists in index.php is because HttpKernel has a requirement on `symfony/debug`, and so the class will always exist when using FrameworkBundle.

PR #214, opened 2017-10-15
[DoctrineBundle] Enable caches by default on prod

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

by leveraging Symfony's cache pools, we can provide a default cache configuration for Doctrine.
Using "cache.system" for metadata and query caches will make them cleareable with a simple "cache:clear", even when APCu is used as a backend - and will auto-select the best available backend, usually APCu, or OPcache.
The result cache is not a system cache, so it should not be cleared on cache:clear, and should be shared when eg a Redis cache is configured for cache.app.

ping @weaverryan as you were interested into this.

### DIFF
--- a/doctrine/doctrine-bundle/1.6/config/packages/prod/doctrine.yaml
+++ b/doctrine/doctrine-bundle/1.6/config/packages/prod/doctrine.yaml
@@ -1,5 +1,31 @@
 doctrine:
     orm:
-        #metadata_cache_driver: apcu
-        #result_cache_driver: apcu
-        #query_cache_driver: apcu
+        metadata_cache_driver:
+            type: service
+            id: doctrine.system_cache_provider
+        query_cache_driver:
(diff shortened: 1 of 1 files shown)


Koc:
@nicolas-grekas is it possible set this config values by default inside `Configuration` class?

weaverryan:
Sort of yes. We discussed this - ideally, someone would submit this change to DoctrineBundle. Then the recipe for the next version of the bundle could be way smaller.

aik099:
If apc/apcu caches aren't shared between web/cli then how it's possible to reset/use web cache from cli, e.g. during deploy?

nicolas-grekas:
By changing the prefix of the keys.

aik099:
But isn't key prefix stored in same cache (but in different key), which different for web and cli?

PR #215, opened 2017-10-16
Remove Makefile support

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT


README.rst
```diff
@@ -77,7 +77,7 @@ unconfiguring the dependencies.
 
 There are eight types of tasks, which are called **configurators**:
```
OndraM: Well now there are only seven of them :-).

aik099:
So the `make cache-warmup` (as on https://medium.com/@fabpot/symfony-4-automate-your-workflow-fbbf609b5a1d) no longer possible? What is the alternative or details why `make` support was removed?

stof:
@aik099 `bin/console cache:warmup`. The make command was a simple proxy in this case anyway (and the validator was forbidding contrib recipes to do the same, to avoid bloating the makefile with proxies not giving access to command options).

The removal of Makefile is due to all the cross-platform issues we have with it (see closed issues in the flex repo). The windows port of Make does not actually support using the same commands (forcing people to go through WSL on Windows 10, and to find other alternatives for older versions). And the macOS version of `make` is also not compatible (forcing people to use `gmake` instead of `make`).
Decision was taken to simply add a requirement on `symfony/console` in the symfony skeleton and keeping `bin/console cache:warmup`, `bin/console cache:clear` and `bin/console server:run`.

aik099:
> Decision was taken to simply add a requirement on symfony/console in the symfony skeleton.

Good news for me. On my first project with Symfony Flex I was also a bit confused how it's possible to have a project without `symfony/console` installed with all polyfills in the `Makefile`.

Thanks for explaining.

Adapik:
Command make cache-warmup is still persent in manifest.json and leads to an error
make: *** No rule to make target 'cache-warmup'. Stop.

Pierstoval:
@Adapik Wait until recipes are updated :wink:

fabpot:
fixed now

PR #216, opened 2017-10-17
when make is gone, composer should use symfony console command

ElectricMaxxx:
| Q             | A
| ------------- | ---
| License       | MIT


symfony/framework-bundle/3.3/manifest.json
```diff
@@ -8,7 +8,7 @@
         "src/": "%SRC_DIR%/"
     },
     "composer-scripts": {
-        "make cache-warmup": "script",
```
fabpot: should be `cache:clear` instead.

PR #217, opened 2017-10-19
Adding php-cache's adapter-bundle and cache-bundle

aequasi:
| Q             | A
| ------------- | ---
| License       | MIT

@nyholm


aequasi:
I've updated the types on https://packagist.org/packages/cache/adapter-bundle and https://packagist.org/packages/cache/cache-bundle

fabpot:
That one should probably be moved to `symfony/recipes-contrib`.

aequasi:
👍 Will move

PR #221, opened 2017-10-23
Enable container.autowiring.strict_mode

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

To preserve BC, autowiring still wires things in hybrid 2.8/3.3 modes.
But 2.8 mode is really a foot gun.
I propose to add a new parameter in SF3.4, to opt-out of this 2.8 mode, and enable this strict mode for all new projects.
WDYT?

### DIFF
--- a/symfony/framework-bundle/3.3/src/Kernel.php
+++ b/symfony/framework-bundle/3.3/src/Kernel.php
@@ -36,6 +36,7 @@ class Kernel extends BaseKernel
 
     protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader)
     {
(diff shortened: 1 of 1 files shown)


stof:
Why does such discussion happens on the recipes repository rather than on Symfony (where the feature needs to be developed) ? I think much less people are following discussions there.

nicolas-grekas:
PR ready for votes, feature https://github.com/symfony/symfony/pull/24671 merged

Tobion:
This should be enabled in symfony-standard 3.4 as well.

PR #222, opened 2017-10-25
Add SensioGeneratorBundle recipe

maidmaid:
| Q             | A
| ------------- | ---
| License       | MIT

cf https://symfony.com/doc/master/bundles/SensioGeneratorBundle/index.html

### DIFF
--- a/sensio/generator-bundle/3.0/manifest.json
+++ b/sensio/generator-bundle/3.0/manifest.json
@@ -1,0 +1,6 @@
+{
+    "bundles": {
+        "Sensio\\Bundle\\GeneratorBundle\\SensioGeneratorBundle": ["dev", "test"]
+    },
+    "aliases": ["generator"]
+}


sstok:
I think dev would be enough, but I don't use this bundle anymore :)

weaverryan:
There will be updates made / a different generator in time for 4.0. When we have that, we can open a recipe, because I agree with @maidmaid that having an alias is a great idea.

So -1 for me, but only because we will do it soon, but not for this bundle.

maidmaid:
@sstok I agree with you, but `test` env accords to the current documentation.

@weaverryan Yes, an alias is useful. Moreover, this recipe prevents to enable this bundle with `all` envs, but only for `dev` (and `test`) env.

fabpot:
Closing as we will have another solution for Symfony 4/Flex.

aik099:
@fabpot , what solution, any blog post/documentation page on that?

PR #223, opened 2017-10-25
Do not enable session in test env by default

Tobion:
Session system is not enabled by default in prod & dev. see https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/config/packages/framework.yaml#L8
So it should also not be enabled by default in test env.

| Q             | A
| ------------- | ---
| License       | MIT


weaverryan:
This makes sense. So, if a user *does* want sessions, will they need to enable in both files?

Tobion:
If they have functional tests using session, then usually yes.

PR #224, opened 2017-10-27
Configure directories in Kernel only at one place

kolibri:
| Q             | A
| ------------- | ---
| License       | MIT

This simplifies to change the location of the `App\Kernel` class (e.g. if you decide to separate Business Logic from Framework integration). Yeah, and removes double logic of getting `config` and `var` directory.

### DIFF
--- a/symfony/framework-bundle/3.3/src/Kernel.php
+++ b/symfony/framework-bundle/3.3/src/Kernel.php
@@ -16,12 +16,12 @@ class Kernel extends BaseKernel
 
     public function getCacheDir(): string
     {
-        return dirname(__DIR__).'/var/cache/'.$this->environment;
+        return $this->getVarDir().'/cache/'.$this->environment;
     }
 
     public function getLogDir(): string
     {
-        return dirname(__DIR__).'/var/log';
+        return $this->getVarDir().'/log';
     }
 
     public function registerBundles()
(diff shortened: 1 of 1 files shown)


aik099:
If the #225 is merged, then in here you might need to remove return type hints and other PHP 7 stuff.

symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -34,27 +34,35 @@ public function registerBundles()
...
+
+    protected function getVarDir(): string
+    {
+        return dirname(__DIR__).'/var';
+    }
+
     protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader)
     {
         $container->setParameter('container.autowiring.strict_mode', true);
-        $confDir = dirname(__DIR__).'/config';
```
nicolas-grekas: I suggest to keep the $confDir variable (same in configureRoutes)

rosier:
The kernel also provides the method getProjectDir(). Maybe it's better to use that instead of this change.

All `dirname(__DIR__)` strings could be replaced by `$this->getProjectDir()`.

symfony/framework-bundle/3.3/src/Kernel.php
kolibri: done

rosier:
I was thinking more along the lines of this https://github.com/symfony/recipes/pull/244

fabpot:
Closing in favor of #244

PR #225, opened 2017-10-27
FrameworkBundle kernel should be compatible

blazarecki:
| Q             | A
| ------------- | ---
| License       | MIT

Hello,

This PR fix an issue when creating a new `symfony/skeleton` project.

`Fatal error: Declaration of App\Kerne_::getCacheDir() must be compatible with App\Kernel::getCacheDir(): string`

The exception appear when configuring `symfony/framework-bundle`

Hope it's helpful.

### DIFF
--- a/symfony/framework-bundle/3.3/src/Kernel.php
+++ b/symfony/framework-bundle/3.3/src/Kernel.php
@@ -14,12 +14,12 @@ class Kernel extends BaseKernel
 
     const CONFIG_EXTS = '.{php,xml,yaml,yml}';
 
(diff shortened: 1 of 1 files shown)


weaverryan:
I don't understand what causes this? What causes the `App\Kerne_` part. Is that during the `cache:clear` from the FWBundle recipe?

stof:
people are still free to add the return type in their own class on PHP 7.2+, and then removing it again would also cause issue. We should detect whether the return type is there in the user-defined class

PR #230, opened 2017-10-31
[FrameworkBundle] Add env vars for router.request_context.*

blazarecki:
| Q             | A
| ------------- | ---
| License       | MIT

Hi,

This PR add env var for `router.request_context.*` parameters as proposed here https://github.com/symfony/symfony/pull/24771#issuecomment-340777957

### DIFF
--- a/symfony/framework-bundle/3.3/config/packages/framework.yaml
+++ b/symfony/framework-bundle/3.3/config/packages/framework.yaml
@@ -1,3 +1,11 @@
+parameters:
+    env(ROUTER_REQUEST_CONTEXT_HOST): null
+    router.request_context.host: '%env(ROUTER_REQUEST_CONTEXT_HOST)%'
+    env(ROUTER_REQUEST_CONTEXT_SCHEME): null
+    router.request_context.scheme: '%env(ROUTER_REQUEST_CONTEXT_SCHEME)%'
+    env(ROUTER_REQUEST_CONTEXT_BASE_URL): null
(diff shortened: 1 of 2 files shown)


symfony/framework-bundle/3.3/config/services.yaml
```diff
@@ -1,6 +1,9 @@
 # Put parameters here that don't need to change on each machine where the app is deployed
 # https://symfony.com/doc/current/best_practices/configuration.html#application-related-configuration
 parameters:
+    router.request_context.host: '%env(ROUTER_REQUEST_CONTEXT_HOST)%'
+    router.request_context.scheme: '%env(ROUTER_REQUEST_CONTEXT_SCHEME)%'
+    router.request_context.base_url: '%env(ROUTER_REQUEST_CONTEXT_BASE_URL)%'
```
nicolas-grekas: if we want these, shouldn't we move them in `config/packages/framework.yaml`?
I'm not sure the env var actually need to be defined, what about setting their default value to `null` also?

javiereguiluz: I agree. Reducing the number of env vars is a must (we're receiving more and more messages from people annoyed by them because maybe we unintentionally overused them).

Nyholm:
@blazarecki Could you update this PR according to the feedback?

blazarecki:
I'll update this PR according to the feedback today.

Tobion:
👎 To me this should not be added to the default recipes as it gives the impression that you need to configure these env vars for the routing component to work. But this is actually only relevant for routing in the CLI context, which is an edge case.

PR #232, opened 2017-11-02
Remove twig extensions config

covex-nn:
| Q             | A
| ------------- | ---
| License       | MIT

Services defaults section should not be inside package configuration. In addition, `twig/extensions` has no configuration at all.


stof:
-1 for this PR

xabbuh:
I am closing here for the reasons given above.

@covex-nn Thank you anyway for your contribution and please leave a comment if you do not agree and maybe try to explain why think we need to make changes here.

covex-nn:
@xabbuh the goal was to remove `_defaults` block from `services`. I think, that default settings for all services should be only in `config/services.yaml`. Without this block a config file is empty, so - i removed a file.

xabbuh:
`_defaults` are only applied to the services that are defined in the same file. Thus, having the block here does not affect your other config files.

xabbuh:
And we need the defaults here as otherwise the services defined below would not be autoconfigured when uncommented by the user.

covex-nn:
I did not know that. Thank you for explanation. There is no sence in my changes.

PR #236, opened 2017-11-03
Use getenv() instead of $_SERVER

gietos:
| Q             | A
| ------------- | ---
| License       | MIT

From documentation:
> There is no guarantee that every web server will provide any of these; servers may omit some, or provide others not listed here.

For example, PHP built-in web server doesn't copy environment variables into `$_SERVER` array.
More straight approach is to use `getenv` function as it's purpose is clear.


stof:
`getenv` was used at first, and it caused lots of issues for Apache users (mod_php for Apache relies on thread-safe builds of PHP), because `getenv` is not thread safe (changing the environment in one thread affects all of them through `getenv`)

PR #238, opened 2017-11-06
Remove "Repository" directory from being ignored in service auto-registration

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

See https://github.com/doctrine/DoctrineBundle/pull/727

But also, this directory was always safe to auto-register. If they cannot be autowired, unless you try to *use* them as services, they will be discarded without showing the autowiring failure.


Pierstoval:
This needs DoctrineBundle 1.8, for people not having a compatible project, this will still lead to repositories as service that'll be deleted in the container

PR #240, opened 2017-11-07
Remove defaults for APP_DEBUG env vars

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

After #239, we don't need APP_DEBUG anymore, and it can be confusing in fact (when mixing a `.env` setting with `-e prod`)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -186,8 +186,7 @@ files stored in the root of the Symfony project:
 
     {
         "env": {
-            "APP_ENV": "dev",
-            "APP_DEBUG": "1"
+            "APP_ENV": "dev"
         }
     }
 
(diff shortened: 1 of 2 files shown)


fabpot:
We still need it for "other" envs, like test. 👎

nicolas-grekas:
Where is it needed? It's never read AFAIK.

nicolas-grekas:
OK, in KernelTestCase, what about keeping only in phpunit.xml.dist?

nicolas-grekas:
updated

nicolas-grekas:
votes pending (tests already have it, see phpunit.xml.dist)

PR #241, opened 2017-11-08
Enable container.dumper.inline_class_loader param by default

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Sibling to https://github.com/symfony/symfony/pull/24872

### DIFF
--- a/symfony/framework-bundle/3.3/src/Kernel.php
+++ b/symfony/framework-bundle/3.3/src/Kernel.php
@@ -37,6 +37,7 @@ class Kernel extends BaseKernel
     protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader)
     {
         $container->setParameter('container.autowiring.strict_mode', true);
+        $container->setParameter('container.dumper.inline_class_loader', true);
         $confDir = $this->getProjectDir().'/config';
         $loader->load($confDir.'/packages/*'.self::CONFIG_EXTS, 'glob');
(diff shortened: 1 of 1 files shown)


symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -37,6 +37,7 @@ public function registerBundles()
     protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader)
     {
         $container->setParameter('container.autowiring.strict_mode', true);
+        $container->setParameter('container.dumper.inline_class_loader', true);
```
dunglas: Can't it be set to `on` by default in 4.0?

nicolas-grekas: it is in the flex recipe, would be a hard BC break otherwise

nicolas-grekas:
rebased

symfony/framework-bundle/3.3/src/Kernel.php
stof: We should still work on making this on by default in the future (we could trigger a deprecation notice in 3.4 if the parameter is not set explicitly to either `true` or `false`, saying that the default behavior in 4.0 would be to enable this).

`container.autowiring.strict_mode` does not suffer from this, because it is just about disabling a BC layer (so it is gone entirely in 4.0)

PR #242, opened 2017-11-08
Remove debug for default PHPUnit config

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/phpunit/phpunit/4.7/phpunit.xml.dist
+++ b/phpunit/phpunit/4.7/phpunit.xml.dist
@@ -11,7 +11,7 @@
         <ini name="error_reporting" value="-1" />
         <env name="KERNEL_CLASS" value="App\Kernel" />
         <env name="APP_ENV" value="test" />
-        <env name="APP_DEBUG" value="1" />
+        <env name="APP_DEBUG" value="0" />
         <env name="APP_SECRET" value="s$cretf0rt3st" />
         <!-- define your env variables for the test env here -->
     </php>
--- a/symfony/phpunit-bridge/3.3/phpunit.xml.dist
+++ b/symfony/phpunit-bridge/3.3/phpunit.xml.dist
@@ -11,7 +11,7 @@
         <ini name="error_reporting" value="-1" />
         <env name="KERNEL_CLASS" value="App\Kernel" />
         <env name="APP_ENV" value="test" />
-        <env name="APP_DEBUG" value="1" />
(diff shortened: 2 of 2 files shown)


stof:
Running tests in no-debug mode has the benefit of running tests much faster (as you don't need to check the cache freshness in each test, while the code would not have changed).
But this has one drawback: it requires you to have either a fresh cache for the test env or no cache. If you have an stale cache, tests will run using it (and will likely fail).

Here is how I solved it at Incenteev:

```php
namespace Incenteev\Tests\Util;

use Symfony\Component\Filesystem\Filesystem;

class CacheClearer
{
    public function clearCache()
    {
        if (getenv('INCENTEEV_KEEP_CACHE')) {
            return;
        }

        $filesystem = new Filesystem();

        $filesystem->remove($this->getCacheFolders());
    }

    private function getCacheFolders()
    {
        return array(
            __DIR__.'/../../../var/api-cache/test',
            __DIR__.'/../../../var/api-cache/behat',
            __DIR__.'/../../../var/cache/test',
            __DIR__.'/../../../var/cache/behat',
        );
    }
}
```

And then, my PHPUnit bootstrap script is this:

```php
require __DIR__.'/../app/autoload.php';

(new \Incenteev\Tests\Util\CacheClearer())->clearCache();
```

This ensures that a developer running tests locally does not have a stale cache (the env variable `INCENTEEV_KEEP_CACHE` allows to avoid clearing the cache on CI, where I know that the cache created by the previous CI command is not stale).

fabpot:
There is another issue with enabling debug, and that's a bunch of logs on the console. Maybe that should be fixed as well. /cc @dunglas

dunglas:
@fabpot maybe can we just remove the `logger` service in test mode?

fabpot:
@dunglas Makes sense. How is it configured for the console (do we have the same issue)?

Pierstoval:
Just by using this and enabling debug mode manually in certain tests (that need profiler or that test debug-related services) I made my tests run 3 times faster 👍

dunglas:
@fabpot see #248.

For the console, it's ok to write logs on stderr isn't it?

fabpot:
@dunglas Having logs entangled with some "real" output always looks like a bug to me.

nicolas-grekas:
What about setting the SHELL_VERBOSITY env var instead?

fabpot:
closing in favor of #281

PR #243, opened 2017-11-10
[MonologBundle] Using yaml configuration

renatomefi:
| Q             | A
| ------------- | ---
| License       | MIT

Since Symfony/Console is now default on Flex I think the special monolog configuration isn't necessary anymore.
What do you think?

Yaml looks more standard with the remaining recipes.

To test it I've dumped the effective configuration `bin/console debug:config MonologBundle` and compared with/without yaml and dev/prod environments.

The comment is kept in a way when you uncomment using your IDE the indentation is valid.


symfony/monolog-bundle/3.1/config/packages/dev/monolog.yaml
```diff
@@ -0,0 +1,19 @@
+monolog:
+    handlers:
+        main:
+            type: stream
+            path: "%kernel.logs_dir%/%kernel.environment%.log"
+            level: debug
+            channels: ["!event"]
+        ## uncomment to get logging in your browser
+        ## you may have to allow bigger header sizes in your Web server configuration
```
Pierstoval: No need for double `#` here

renatomefi: @Pierstoval keep or not the space as well?

Pierstoval: Simple: someone that uncomments this code may also remove it, so make it like it's a normal comment

PR #245, opened 2017-11-10
[WIP]: Use path variable inside config file

kolibri:
This is a testing pull request, please do not close it immediately. I need the generated symfony.sh endpoint to this PR, to test a feature in symfony/flex, that enables the usage of directory variables inside of copied files.
I promise, to close this PR by myself until my tests are done (maybe, two weeks, until 24.Nov.2017?)

| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
@@ -9,6 +9,6 @@ doctrine:
             App:
                 is_bundle: false
                 type: annotation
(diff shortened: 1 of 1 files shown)


fabpot:
If you a question, I think it's better to ask it instead of hacking this repo.

kolibri:
Yeah, I asked @xabbuh about my problem to test with modified recipes, and he showed me this way of doing this. But you are right, there should be a better way to do that, so I asked about it in #246
Pardon for hacking.

PR #248, opened 2017-11-10
[framework-bundle] Disable the logger in test env

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/framework-bundle/3.3/config/packages/test/framework.yaml
+++ b/symfony/framework-bundle/3.3/config/packages/test/framework.yaml
@@ -2,3 +2,7 @@ framework:
     test: ~
     #session:
     #    storage_id: session.storage.mock_file
+
+services:
+    # Disable logs
+    logger: Psr\Log\NullLogger


symfony/framework-bundle/3.3/config/packages/test/framework.yaml
```diff
@@ -2,3 +2,7 @@ framework:
     test: ~
     #session:
     #    storage_id: session.storage.mock_file
+
+services:
+    # Disable logs
```
ogizanagi: As soon as Monolog is used with a proper stream handler for instance, registering this null logger probably isn't a good idea.

dunglas: Why? It will speed up the builds, even with Monolog.

ogizanagi: Logs can be useful to understand why a test fail. I'm not sure about this as is in the recipe.

dunglas: Anyway monolog has no specific config for tests right now.
It's just a recipe, for specific needs, debugging etc, the user can (should?) modify this file to fit his needs.

stof: the question then is whether this logger is winning over monolog or no. This should be checked
The goal is to have it winning over the default logger added in FrameworkBundle when no logger is defined

fabpot:
closing in favor of #281

symfony/framework-bundle/3.3/config/packages/test/framework.yaml
lsmith77: or maybe we should make it possible to disable the default logger via config

PR #255, opened 2017-11-13
Change the default route to API platform to /api

fabpot:
If not, it overrides the / namespace, which is not so nice.

| Q             | A
| ------------- | ---
| License       | MIT


sroze:
That's less user friendly 🤔

fabpot:
Yes and no :) Without this change, there was a conflict between `/` (API platform) and `/`, the homepage. So, having a clear distinction is a must have. Up to the dev of course to change this default.

PR #258, opened 2017-11-16
Always set Doctrine server version & platform

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

In order to run `cache:warmup`, you need to set the server platform and server version. I believe this can be accomplished by using some fake string like `DATABASE_URL: 'mysql://?server_version=5.7`

In this PR, I attempt to make the `DATABASE_URL` *truly* optional during `cache:warmup`. I do this by committing the server (e.g. mysql) and version (e.g. 5.7) into a config file. This allows us to create a fallback when needed.

On the downside, `doctrine.yaml` looks a little strange. Also, I suppose it might be less obvious if you forget to specify `DATABASE_URL` on production (because your connection will fail vs a `Missing DATABASE_URL environment variable` error).

On the upside, you can once again run `cache:warmup` with no environment variables at all (very nice for PAAS).

### DIFF
--- a/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
@@ -1,13 +1,9 @@
 parameters:
-    # update these for your app
-    doctrine.database_server: mysql
-    doctrine.server_version: 5.7
-
     # adds a fallback URL if the env var is not set
     # this is useful to allow you to run cache:warmup
     # even when your environment variables don't yet exist
     # you should not need to change this value
-    DATABASE_URL: '%doctrine.database_server%://'
+    env(DATABASE_URL): '%doctrine.database_server%://'
 
 doctrine:
     dbal:
--- a/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
@@ -10,6 +10,7 @@ doctrine:
         # configure these for your database server
         driver: 'pdo_mysql'
         server_version: '5.7'
+        charset: utf8mb4
 
         # With Symfony 3.3, remove the `resolve:` prefix
         url: '%env(resolve:DATABASE_URL)%'
--- a/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
@@ -1,8 +1,19 @@
+parameters:
+    # update these for your app
+    doctrine.database_server: mysql
+    doctrine.server_version: 5.7
+
+    # adds a fallback URL if the env var is not set
+    # this is useful to allow you to run cache:warmup
+    # even when your environment variables don't yet exist
+    # you should not need to change this value
+    DATABASE_URL: '%doctrine.database_server%://'
+
 doctrine:
     dbal:
         # With Symfony 3.3, remove the `resolve:` prefix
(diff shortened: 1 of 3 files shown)


doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
```diff
@@ -1,7 +1,15 @@
+parameters:
+    # adds a fallback URL if the env var is not set
+    # this is useful to allow you to run cache:warmup
+    # even when your environment variables don't yet exist
+    # you should not need to change this value
+    env(DATABASE_URL): '%doctrine.database_server%://'
```
weaverryan: Is it possible to specify a default value for the `DATABASE_URL` environment value in some other way (i.e. not below `parameters`)? I *could* move this to `services.yaml` as @symfony-flex-server suggests, but it's really better here, as the user should not really configure this.

weaverryan:
I've just simplified this a *bunch*:

A) Things like `driver` and `server_version` would now be committed in `doctrine.yaml` instead of embedded in the env var. I think this makes sense: you really don't want to vary your db driver or version from server to server (but you still could be adding the options to `DATABASE_URL`).

B) Thanks to this, `DATABASE_URL` no longer needs to be set for cache warmup. Actually, it *does* need to be set, but we can safely default it to an empty string.

doctrine/doctrine-bundle/1.6/manifest.json
```diff
@@ -10,6 +10,6 @@
         "#1": "Format described at http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url",
         "#2": "For an SQLite database, use: \"sqlite:///%kernel.project_dir%/var/data.db\"",
         "#3": "Set \"serverVersion\" to your server version to avoid edge-case exceptions and extra database calls",
-        "DATABASE_URL": "mysql://db_user:db_password@127.0.0.1:3306/db_name?charset=utf8mb4&serverVersion=5.7"
+        "DATABASE_URL": "mysql://db_user:db_password@127.0.0.1:3306/db_name"
```
fabpot: Would it make sense to mention that one should also not forget to change the `doctrine.yaml` file? And vice-versa?

weaverryan: Good idea. I just changed the `#3` comment line to mention `doctrine.yaml`. But I didn't add a note in `doctrine.yaml` about `.env`: it seemed weird, because `DATABASE_URL` is an env variable, so saying "edit it in `.env`" is only true in dev mode.... I wasn't convinced if it should be added there or not.

weaverryan:
Note to self: if this is merged, https://github.com/symfony/symfony-docs/pull/8705/files#diff-d796577eeafc4ece119cf5df3492e5dbR124 can be simplified slightly.

weaverryan:
The use-case *is* a little bit specific, but I think it'll be more and more common (it's standard practice on a PAAS like platform.sh or Heroku to warm up the cache during a build stage where no env parameters are added). This would make it work for everyone automatically instead of PAAS users needing to do extra.

I don't think there's much downside:

* For `driver`, if you don't realize the config is present and choose a conflicting `DATABASE_URL`, (e.g. one that uses pgsql) the `DATABASE_URL` wins

* For `server_version`, if you don't realize this is here, you could possibly get some weird behavior if you connect to a DB with a different version. That is also true of the current recipe, though it would be a bit easier to miss with this change.

* Putting the `charset` in config is a homerun I think :)

PR #259, opened 2017-11-17
Add maker-bundle recipe

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/maker-bundle/1.0/manifest.json
+++ b/symfony/maker-bundle/1.0/manifest.json
@@ -1,0 +1,6 @@
+{
+    "bundles": {
+        "Symfony\\Bundle\\MakerBundle\\MakerBundle": ["dev"]
+    },
+    "aliases": ["maker", "generator"]
+}


fabpot:

fabpot:
@symfony-flex-server review one more time please

fabpot:
@symfony-flex-server review as I'm still sleeping a bit, messed up last time, sorry.

symfony/maker-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,6 @@
+{
+    "bundles": {
+        "Symfony\\Bundle\\MakerBundle\\MakerBundle": ["dev"]
+    },
+    "aliases": ["maker", "generator"]
```
xabbuh: What about also adding `code-generator` (as used in the keywords of the bundle) and maybe also `generator-bundle`?

fabpot: Aliases should be short and sweet.

fabpot: generator-bundle would be confusing as the current one generates CRUD, not this one. So, make-bundle is still not a replacement for SensioGeneratorBundle (unfortunately).

PR #260, opened 2017-11-17
Move routes config for annotations

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

Let's make routes configuration when using annotations automatic.

### DIFF
--- a/doctrine/annotations/1.0/config/routes/annotations.yaml
+++ b/doctrine/annotations/1.0/config/routes/annotations.yaml
@@ -1,0 +1,3 @@
+controllers:
+    resource: ../src/Controller/
+    type: annotation
--- a/doctrine/annotations/1.0/manifest.json
+++ b/doctrine/annotations/1.0/manifest.json
@@ -1,0 +1,5 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    }
+}
--- a/symfony/routing/3.3/config/routes.yaml
+++ b/symfony/routing/3.3/config/routes.yaml
@@ -1,8 +1,3 @@
 #index:
 #    path: /
 #    defaults: { _controller: 'App\Controller\DefaultController::index' }
-
-# first, run composer req annotations
-#controllers:
-#    resource: ../src/Controller/
-#    type: annotation


fabpot:
We cannot merge it now as it fails when the `src/Controller` is empty. But it shouldn't. That's a bug that should be fixed in Symfony first.

xabbuh:
@fabpot Are you sure? This works for me even when the directory is empty. It only fails when the directory does not exist.

doctrine/annotations/1.0/config/routes/annotations.yaml
```diff
@@ -0,0 +1,3 @@
+controllers:
+    resource: ../../src/Controller/
```
xabbuh: `resource: ../src/Controller/`

xabbuh: relative paths are interpreted relative to the location of kernel (probably related to symfony/symfony#24982)

weaverryan:
Yes, I believe it only errors out when the directory does NOT exist. But, it should always exist in a new Flex app, thanks to the framework-bundle recipe, correct? I mean, it would be better if it did *not* error if the dir didn't exist, but I think maybe it's not a problem?

doctrine/annotations/1.0/config/routes/annotations.yaml
backbone87: But then `resource: Controller/` should be enough

Pierstoval:
Shouldn't this have been set for `SensioFrameworkBundle` instead of `doctrine/annotations`?

weaverryan:
@Pierstoval I was just thinking the same thing. I'm not sure if it was done on purpose or not. Could you open a new PR to suggest adding it to SensioFrameworkExtraBundle instead?

nicolas-grekas:
The Route annotation doesn't need to be from the bundle: 3.4 has a fine working annotation class
That's why the config is there.
Did you spot any actual issue?

backbone87:
But then this should be in the framework recipe, because If i use doctrine annotations without the framework bundle, this config will cause errors, no?

nicolas-grekas:
framework is a mandatory requirement of flex apps (and it doesn't require annotation, so double "no" :) )

fabpot:
To sum up: it was done on purpose :) Nothing to change here.

doctrine/annotations/1.0/config/routes/annotations.yaml
```diff
@@ -0,0 +1,3 @@
+controllers:
+    resource: ../src/Controller/
```
backbone87: since this is relative to `%kernel.root_dir%` it should be just `Controller/`

yceruto: I think it's relative to the entry point (`public/index.php` or `bin/console`) where the kernel is instantiate.

PR #272, opened 2017-11-22
Use config-relative path for routing annotation resources

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Requires https://github.com/symfony/symfony/pull/25113

### DIFF
--- a/doctrine/annotations/1.0/config/routes/annotations.yaml
+++ b/doctrine/annotations/1.0/config/routes/annotations.yaml
@@ -1,3 +1,3 @@
 controllers:
-    resource: ../src/Controller/
+    resource: ../../src/Controller/
     type: annotation


doctrine/annotations/1.0/config/routes/annotations.yaml
```diff
@@ -1,3 +1,3 @@
 controllers:
-    resource: ../src/Controller/
+    resource: ../../src/Controller/
```
Nyholm: Couldn't we be using the following instead: 

```
controllers:
    resource: "%kernel.project_dir%/src/Controller"
```
???

nicolas-grekas: was my first idea, but parameters are not parsed in the config area, would be a whole new feature

Nyholm:
When testing this change I get an error. 

Using `../src/Controller` works just fine. 

screen shot 2017-11-23 at 12 52 19

doctrine/annotations/1.0/config/routes/annotations.yaml
Nyholm: Ah, okey. Thanks

nicolas-grekas:
@Nyholm yep it works, but it's WTF (why a single "..", doesn't make any sense unless you know some internal details).
Fixed by symfony/symfony#25113 (still pending.)

doctrine/annotations/1.0/config/routes/annotations.yaml
fabpot: As the related change is done in 3.4, it means that this config will break on project using 3.3, right?

nicolas-grekas: right, so I'll rebase against 3.3 as bug fix!

nicolas-grekas: they won't: the fallback mechanism will look on the previous path and work as before

fabpot: ok, great then.

fabpot: IIRC, this was what we had at first and it didn't work.

nicolas-grekas:
votes pending

PR #273, opened 2017-11-22
[DoctrineBundle] Proposing utf table options

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

In #258, I propose putting the database `charset` into `doctrine.yaml` instead of having it in the `DATABASE_URL`. While updating the Doctrine docs, I noticed we also recommend setting the `default_table_options`.

Question: If `charset` is set to `utf8mb4` under `doctrine.dbal`, does the user also need to set these table options, or is it redundant? If it is not redundant, then I'd like to propose this change and properly set people up with UTF8 out-of-the-box.

### DIFF
--- a/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
@@ -2,6 +2,10 @@ doctrine:
     dbal:
         # With Symfony 3.3, remove the `resolve:` prefix
         url: '%env(resolve:DATABASE_URL)%'
+
+        default_table_options:
+            charset: utf8
+            collate: utf8_unicode_ci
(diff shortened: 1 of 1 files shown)


Nyholm:
I do not think they are redundant. DoctrineBundle does not specify any default values for these parameters.

fabpot:
2 questions:

* First is for my personal curiosity: are these values read when creating the database?
* Can we use the same values for all supported DBs (are they standard or abstracted in Doctrine DBAL)?

weaverryan:
Actually, I don't think we should have this. First I believe these *are* engine-specific. And also, after some digging, I see that, for MySQL, these default to `charset: utf8` and `collate: utf8mb4_general_ci`, which I think are at least sensible defaults.

PR #275, opened 2017-11-23
Removing \ which does not work for XML

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

With annotations, the code seems to work with or without the trailing `\`. But if you change this to `type: xml`, then you *must* remove the trailing `\`.

So, let's remove it - it makes changing to XML easier... if you're into that sorta thing.

Thanks!

### DIFF
--- a/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
@@ -23,5 +23,5 @@ doctrine:
                 is_bundle: false
                 type: annotation
                 dir: '%kernel.project_dir%/src/Entity'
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Could this be considered a Doctrine bug? Removing the trailing `\` when using XML would be such a trivial fix ... but I don't know if it's possible.

weaverryan:
I really don’t know, but it doesn’t sound like a bug (since annotations works, but not xml with the slash). But without the slash works everywhere, so I think it’s a better default for us.

If someone wants to try it, the bug I believe would be in the SymfonyFileLocator - there is a line that appends a \ before doing a strpos.

PR #277, opened 2017-11-24
Prefer simple-phpunit over direct phpunit deps

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Several changes to have a better phpunit integration:

- ~~move `phpunit.xml.dist` to framework bundle so that a global phpunit works, and env vars are correctly added to it from the beginning~~
- add `SymfonyTestsListener` to `phpunit.xml.dist` so that using phpunit directly still enables the bridge (if the bridge is not installed, phpunit just ignores the listener)
- recommend installing `simple-phpunit` when `phpunit/phpunit` is installed
- add `bin/phpunit` helper script to run phpunit when simple-phpunit is installed (borrowed from [our own in symfony/symfony](https://github.com/symfony/symfony/blob/master/phpunit))

### DIFF
--- a/phpunit/phpunit/4.7/post-install.txt
+++ b/phpunit/phpunit/4.7/post-install.txt
@@ -1,0 +1,2 @@
+Adding <comment>phpunit/phpunit</> as a dependency is <comment>discouraged</>.
+You should require <fg=green>simple-phpunit</> instead.
--- a/symfony/phpunit-bridge/3.3/bin/phpunit
+++ b/symfony/phpunit-bridge/3.3/bin/phpunit
@@ -1,0 +1,18 @@
+#!/usr/bin/env php
+<?php
+
+if (!file_exists(dirname(__DIR__).'/vendor/symfony/phpunit-bridge/bin/simple-phpunit')) {
+    echo "Unable to find the `simple-phpunit` script in `vendor/symfony/phpunit-bridge/bin/`.\n";
+    exit(1);
+}
+if (false === getenv('SYMFONY_PHPUNIT_REMOVE')) {
+    putenv('SYMFONY_PHPUNIT_REMOVE=symfony/yaml');
+}
+if (false === getenv('SYMFONY_PHPUNIT_VERSION')) {
+    putenv('SYMFONY_PHPUNIT_VERSION=6.4');
+}
+if (false === getenv('SYMFONY_PHPUNIT_DIR')) {
+    putenv('SYMFONY_PHPUNIT_DIR='.__DIR__.'/.phpunit');
+}
+
+require dirname(__DIR__).'/vendor/symfony/phpunit-bridge/bin/simple-phpunit';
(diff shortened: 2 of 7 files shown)


phpunit/phpunit/4.7/post-install.txt
```diff
@@ -0,0 +1,4 @@
+
+<comment>    Adding </>phpunit/phpunit<comment> as a dependency is discouraged. </>
```
fabpot: whitespace at the end should be removed

xabbuh:
> if the bridge is not installed, phpunit just ignores the listener

Is that true for all PHPUnit versions?

nicolas-grekas:
@xabbuh AFAIK yes

> ~~move `phpunit.xml.dist` to framework bundle so that a global phpunit works, and env vars are correctly added to it from the beginning~~

reverted this one because then "APP_DEV=dev" is then set in the file.

fabpot:
@symfony-flex-server review

symfony/phpunit-bridge/3.3/bin/phpunit
```diff
@@ -0,0 +1,18 @@
...
+
+if (!file_exists(dirname(__DIR__).'/vendor/symfony/phpunit-bridge/bin/simple-phpunit')) {
+    echo "Unable to find the `simple-phpunit` script in `vendor/symfony/phpunit-bridge/bin/`.\n";
+    exit(1);
+}
+if (false === getenv('SYMFONY_PHPUNIT_REMOVE')) {
+    putenv('SYMFONY_PHPUNIT_REMOVE=symfony/yaml');
+}
+if (false === getenv('SYMFONY_PHPUNIT_VERSION')) {
+    putenv('SYMFONY_PHPUNIT_VERSION=6.4');
```
stof: -1 for that. We don't want people to have to update the minor version manually all the time. If the PHPUnit version is not set externally, we should keep the default guessed in the bridge IMO (which should be upgraded to use PHPUnit 6 on all supported PHP versions in SF4)

symfony/phpunit-bridge/3.3/bin/phpunit
```diff
@@ -0,0 +1,18 @@
...
+}
+if (false === getenv('SYMFONY_PHPUNIT_DIR')) {
+    putenv('SYMFONY_PHPUNIT_DIR='.__DIR__.'/.phpunit');
```
stof: this folder must be added in `.gitignore`

nicolas-grekas: see #282

symfony/phpunit-bridge/3.3/bin/phpunit
nicolas-grekas: Not sure I agree: Flex requires PHP 7.0, so that any 6.* version of phpunit is fine.
The bridge currently set "6.0" for PHP 7.0/7.1 or "6.3" otherwise (>=7.2).
The bridge cannot be updated easily for BC reasons.
The recipe *can* be as often as we want.

PR #279, opened 2017-11-25
[FrameworkBundle][Routing] Add a welcome controller

abellion:
| Q             | A
| ------------- | ---
| License       | MIT

It would be easier for newcomers to start with a predefined controller and a working route.
Also, I think we could get rid of this code : https://github.com/symfony/symfony/pull/24403

I hardcoded the template path in the `manifest.json` because Flex doesn't have it in the options : https://github.com/symfony/flex/blob/master/src/Flex.php#L284

image

### DIFF
--- a/symfony/framework-bundle/3.3/manifest.json
+++ b/symfony/framework-bundle/3.3/manifest.json
@@ -5,7 +5,8 @@
     "copy-from-recipe": {
         "config/": "%CONFIG_DIR%/",
         "public/": "%PUBLIC_DIR%/",
-        "src/": "%SRC_DIR%/"
+        "src/": "%SRC_DIR%/",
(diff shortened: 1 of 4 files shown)


Pierstoval:
Starting from Symfony 3.4, this behavior is hardcoded in debug mode when there is no route:

https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/HttpKernel/EventListener/RouterListener.php#L127-L131

fabpot:
As you noticed, we decided to go another route with the PR you referenced. We don't want to generate code that is not useful (when developing an API, you don't need a template directory, we also never recommend using PHP for templates, ...). Moreover, maker-bundle now allows to easily generate controllers and templates, so having a default controller is not really needed anymore.

PR #282, opened 2017-11-27
gitignore /bin/.phpunit

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/phpunit-bridge/3.3/manifest.json
+++ b/symfony/phpunit-bridge/3.3/manifest.json
@@ -5,6 +5,7 @@
         "tests/": "tests/"
     },
     "gitignore": [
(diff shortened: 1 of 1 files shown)


symfony/phpunit-bridge/3.3/manifest.json
```diff
@@ -5,6 +5,7 @@
         "tests/": "tests/"
     },
     "gitignore": [
+        ".phpunit",
```
stof: should be `/bin/.phpunit` instead

nicolas-grekas: wouldn't account for `%BIN_DIR%`, while this does

PR #283, opened 2017-11-28
Clean symfony from PHPUnit library

yoannrenard:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/phpunit/phpunit/4.7/phpunit.xml.dist
+++ b/phpunit/phpunit/4.7/phpunit.xml.dist
@@ -22,8 +22,4 @@
             <directory>tests/</directory>
         </testsuite>
     </testsuites>
-
-    <listeners>
-        <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener" />
-    </listeners>
 </phpunit>


fabpot:
Those lines are here for a reason. Why would you want to remove them?

yoannrenard:
@fabpot IMO, those lines shouldn't be there as, if I understood it well, this recipe is aim to install `phpunit/phpunit` not `symfony/phpunit-bridge` which already has its own recipe.

xabbuh:
@yoannrenard But we still want the listener to be used if the PhpUnitBridge is installed (see symfony/recipes#277).

Nyholm:
Thank you @yoannrenard for this PR. 
I could agree that it is weird having those lines there since they refer to a class that belongs to a different library. But as @xabbuh says, they do make sense if you have the PHPUnitBridge installed which you will have in 90% of the cases. 

PS. They do not harm if PHPUnitBridge is not installed.

yoannrenard:
Yes it is kinda weird but actually understandable @Nyholm . 
The thing is that the class `TestRunner` from `phpunit` throws me an exception _Class "Symfony\Bridge\PhpUnit\SymfonyTestsListener" does not exist_ when I use `phpunit` without `phpunit-bridge`

I give you, underneath, an extract of my composer.json file for more context :  
```json
{
    "require": {
        "php": "^7.1.3",
        "symfony/flex": "^1.0"
    },
    "require-dev": {
        "phpunit/phpunit": "^6.4"
    }
}
```

xabbuh:
It looks like the handling of non-existent listeners changed between PHPUnit 4 and 5.

Nyholm:
Oh, I guess you are correct. Thank you. 

```cli
$ composer create-project "symfony/skeleton:^4.0@rc" demo
$ cd demo
$ composer require phpunit/phpunit
$ ./vendor/bin/phpunit
Class "Symfony\Bridge\PhpUnit\SymfonyTestsListener" does not exist
```

Christian showed me that PHPUnit has changed this behaviour in between [version 4](https://github.com/sebastianbergmann/phpunit/blob/4.8/src/TextUI/TestRunner.php#L808-L823) and [version 5](https://github.com/sebastianbergmann/phpunit/blob/5.6/src/TextUI/TestRunner.php#L885-L892).

xabbuh:
ping @nicolas-grekas

nicolas-grekas:
Is there a way to ban phpunit? The bridge should always be used really.

yoannrenard:
If the point is to ban phpunit from Flex, I guess we just need to get rid of this phpunit [recipe](https://github.com/symfony/recipes/tree/master/phpunit/phpunit/4.7) and add an alias "phpunit/phpunit" into the phpunit-bridge [recipe](https://github.com/symfony/recipes/blob/master/symfony/phpunit-bridge/3.3/manifest.json#L10) one.

As long as we still can install `phpspec/prophecy` upon it it is fine to me.

nicolas-grekas:
> Class "Symfony\Bridge\PhpUnit\SymfonyTestsListener" does not exist

:(

> As long as we still can install phpspec/prophecy upon it it is fine to me.

prophecy is bundles into phpunit, isn't it?
and phpspec can just be installed as a dev dep.
can you try installing phpunit via the bridge and report back please?

yoannrenard:
Yes @nicolas-grekas the phpunit/phpunit lib embeds phpspec/prophecy.

Ok so, I tried and installed : (_using last versions_)  

| Package | Using Flex ? | Contains Prophecy? | Comment |
| ----------- | --------------- | -------------------- | ------------ |
| phpunit/phpunit | No | Yes |  |
| phpunit/phpunit | Yes | Yes | Doesn't work though because of the listener that doesn't exists |
| symfony/phpunit-bridge | No | No | As expected |
| symfony/phpunit-bridge | Yes | Yes | It does contain Prophecy (Disturbing isn't it?) |

Hoping that I've been clear enough

nicolas-grekas:
OK, so since we're talking about a recipe, we always have Flex.
Yes, the default behavior of the bridge is to disable prophecy.
And the bridge's recipe re-enables it.

What about phpspec? Can you use it when only the bridge is used?

yoannrenard:
Yes it does work using the bridge

PR #285, opened 2017-11-29
Hopefully improve comments in services.yaml

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT


symfony/framework-bundle/3.3/config/services.yaml
```diff
@@ -5,24 +5,23 @@ parameters:
 services:
     # default configuration for services in *this* file
     _defaults:
-        # automatically injects dependencies in your services
```
nicolas-grekas: The current "before line the line" comments make the section scary to me
I propose to put these comments on the end of each lines - looks better to the eye IMHO

symfony/framework-bundle/3.3/config/services.yaml
```diff
@@ -5,24 +5,23 @@ parameters:
...
+        public: false       # Allows optimizing the container by removing unused services; this also means
+                            # fetching services directly from the container via $container->get() won't work.
+                            # The best practice is to be explicit about your dependencies anyway.
 
     # makes classes in src/ available to be used as services
     # this creates a service per class whose id is the fully-qualified class name
     App\:
         resource: '../src/*'
-        # you can exclude directories or files
-        # but if a service is unused, it's removed anyway
```
nicolas-grekas: not needed anymore, esp since service removal is mentioned above now

symfony/framework-bundle/3.3/config/services.yaml
```diff
@@ -5,24 +5,23 @@ parameters:
...
         exclude: '../src/{DataFixtures,Entity,Migrations,Tests}'
 
-    # controllers are imported separately to make sure they
-    # have the tag that allows actions to type-hint services
+    # controllers are imported separately to make sure services can be injected
+    # as action arguments even if you don't extend any base controller class
```
nicolas-grekas: let's explain *why* this is needed in the first place (ie: only if you do not extend a base controller)

symfony/framework-bundle/3.3/config/services.yaml
```diff
@@ -5,24 +5,23 @@ parameters:
...
     App\Controller\:
         resource: '../src/Controller'
         tags: ['controller.service_arguments']
+
+    # add more service definitions when explicit configuration is needed
+    # please note that last definitions always *replace* previous ones
```
nicolas-grekas: I think this last note is the most important one, as we've already got several reports from ppl who've got that wrong

symfony/framework-bundle/3.3/config/services.yaml
```diff
@@ -5,24 +5,23 @@ parameters:
...
+                            # fetching services directly from the container via $container->get() won't work.
+                            # The best practice is to be explicit about your dependencies anyway.
 
     # makes classes in src/ available to be used as services
     # this creates a service per class whose id is the fully-qualified class name
     App\:
         resource: '../src/*'
-        # you can exclude directories or files
-        # but if a service is unused, it's removed anyway
         exclude: '../src/{DataFixtures,Entity,Migrations,Tests}'
```
weaverryan: Remove `DataFixtures`, we just fixed this with DoctrineFixturesBundle 3.0. Woo!

symfony/framework-bundle/3.3/config/services.yaml
javiereguiluz: Sadly I don't think this --> *"please note that last definitions always *replace* previous ones"* will be enough to understand the problem. The worst thing is that is a tricky problem and not easy to explain: if you define a controller services bellow the `App\Controller\` config, it's natural to think that your service will inherit everything from above.

nicolas-grekas: that's why we absolutely need this line: to hint the user when something doesn't work as one would maybe expected

weaverryan: Should we include a commented-out example config below?

```yml
    # Example: override a service so you can explicitly wire a `$isDebug` constructor arg
    # App\Service\ExampleService
    #     arguments:
    #         $isDebug: '%kernel.debug%'
```

nicolas-grekas: not sure: we did not add examples for parameters, and i think it was a conscious

PR #286, opened 2017-11-29
adding DoctrineFixturesBundle 3.0 recipe

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

We released DoctrineFixturesBundle 3.0 yesterday. In it, we actually recommend putting fixtures classes in the more "shallow" `DataFixtures` directory. The `ORM` is not needed anymore, because fixtures are tagged services - they can really live anywhere.

There is a 2.4 recipe in `recipes-contrib`. But I thought 3.0 could be official.

There are no documentation changes needed. We don't doc usage of this bundle on `symfony-docs` currently, and the docs in the bundle are already updated.

### DIFF
--- a/doctrine/doctrine-fixtures-bundle/3.0/manifest.json
+++ b/doctrine/doctrine-fixtures-bundle/3.0/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "Doctrine\\Bundle\\FixturesBundle\\DoctrineFixturesBundle": ["dev", "test"]
+    },
+    "copy-from-recipe": {
+        "src/": "%SRC_DIR%/"
+    }
+}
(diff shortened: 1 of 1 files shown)


weaverryan:
Should we create an alias `fixtures`, `orm-fixtures` or `doctrine-fixtures`?

Pierstoval:
Is there any fixture tool/bundle for the ODM? Maybe `orm-fixtures` can be a good start, and if there's nothing, or if it's not popular, simply `fixtures` could be nice too

alcaeus:
Not a fan of `doctrine-fixtures`, `orm-fixtures` is a good one. Not sure about `fixtures` - I'll let someone else decide that one.

weaverryan:
Added `orm-fixtures` alias - that's a pretty safe alias to use.

This is ready.

fabpot:
@weaverryan Can you submit a PR on contrib to remove the recipe there?

weaverryan:
symfony/recipes-contrib#166

PR #288, opened 2017-11-30
Move routes config

michalpipa:
| Q             | A
| ------------- | ---
| License       | MIT

Config for routes annotations should be defined in _sensio/framework-extra-bundle_ instead of _doctrine/annotations_. Now _doctrine/annotations_ installation fails (without _sensio/framework-extra-bundle_ installed), because Route annotation is not defined.


weaverryan:
Hmm, this was done on purpose actually. Can you give more details about the failure? Do you have an @Route annotation in a controller?

michalpipa:
It is very easy to reproduce:
```
composer create-project symfony/skeleton test/
cd test/
composer req doctrine/annotations
```
_doctrine/annotations_ installation fails with error message: "Cannot load resource "../src/Controller/". Make sure annotations are enabled". It happens because I don't have _sensio/framework-extra-bundle_. Same thing happens if you require _jms/serializer_, which depends on _doctrine/annotations_.

I don't have any controller, It happens on a clean project.

xabbuh:
This is indeed an issue when using Symfony 3.3 packages (which is done by default until tonight).

If the project were created with 3.4, this would not be an issue.

michalpipa:
For me it doesn't make sense to configure Route annotation while installing _doctrine/annotations_. This annotation is defined by _sensio/framework-extra-bundle_, not _doctrine/annotations_.

What if I want to use annotations, but I don't want to use Route annotation? What if I want to use _jms/serializer_, but I don't want to use _sensio/framework-extra-bundle_?

And it still will be a problem for 3.3 projects.

stof:
@michalpipa as of Symfony 3.4, FrameworkBundle itself supports annotation-based routing, as soon as `doctrine/annotations` is installed.

michalpipa:
Ok, I have found this comment: https://github.com/symfony/recipes/pull/260#issuecomment-345768144, so this only problem is when you use this with Symfony 3.3.

stof:
if you can always *remove* the config for the annotation routing if you don't want it. Flex puts an initial config in your project, but after that, you can do **whatever you want** with it after that (changing the config, removing it, etc...) and Flex will not mess with it (unless you remove the package name from the symfony.lock file, as Flex would then forget that it already handled it).

The only gotcha is that Flex will remove the corresponding file when the package gets removed later, so adding totally unrelated config might not be a good idea.

stof:
Using Flex with Symfony 3.3 is considered experimental. To be in a safe land when using Flex, it has to be done with Symfony 3.4+

fabpot:
Closing as Symfony 3.3 has some issues like this one, which we cannot really fix. Flex should really be used on 3.4+.

PR #289, opened 2017-11-30
Create recipe for the Apache pack

### LINKED ISSUE #190: Recipe for "public/.htaccess" file
I can't find a recipe that would install default ".htaccess" file (see https://github.com/symfony/symfony-standard/blob/3.3/web/.htaccess) into "public" folder so that all mod-rewrite urls would still work.

Without it after fresh install I'm getting 404 on any route access (including routes generated by Web Profiler).

Pierstoval:
| Q             | A
| ------------- | ---
| License       | MIT

Fixes #190

### DIFF
--- a/symfony/apache-pack/1.0/manifest.json
+++ b/symfony/apache-pack/1.0/manifest.json
@@ -1,0 +1,6 @@
+{
+    "copy-from-recipe": {
+        "public/.htaccess": "%PUBLIC_DIR%/.htaccess"
+    },
+    "aliases": ["htaccess", "apache", "apache-pack"]
+}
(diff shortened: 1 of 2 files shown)


fabpot:
I think it belongs to `recipes-contrib`. I would certainly not recommend using Apache.

Pierstoval:
Users will not be able to use the alias then, and even though the Symfony team doesn't recommend apache, maybe providing an "official" apache config could be a better option to say _"We don't recommend it, but here's the way to do it if you really want it"_.

WDYT?

xabbuh:
IMO even when using Apache, using a `.htaccess` file is not the recommend approach, but the configuration should be done on the vhost level instead.

weaverryan:
The alias is not a huge deal... it *is* kinda nice, but people will need to read the docs anyways to find this. Let's put it on contrib for now.

There is already an identical PR waiting in contrib - symfony/recipes-contrib#164

stof:
Closing this, as @fabpot explicitly said in the issue that he wants this recipe to live in contrib.

PR #292, opened 2017-12-01
Adding services_test.yaml when you install phpunit

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

Soon, we will document that you should create public aliases if you need to access a private service from a test. What do you guys think about giving them this configuration file? Creating it in the FWBundle recipe seemed overkill. But if you install phpunit, it seems much more likely that it will be needed.

### DIFF
--- a/symfony/phpunit-bridge/3.3/config/services_test.yaml
+++ b/symfony/phpunit-bridge/3.3/config/services_test.yaml
@@ -1,0 +1,9 @@
+services:
+    _defaults:
+        public: true
+
+    # If you need to access services in a test, create an alias
+    # and then fetch that alias from the container. As a convention,
+    # aliases are prefixed with test. For example:
+    #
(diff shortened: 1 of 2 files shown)


fabpot:
Should probably be part of the `symfony/phpunit-bridge` recipe instead as we discourage people from installing `phpunit/phpunit` with Composer.

weaverryan:
... which would also mean that we should start recommending that users install `symfony/phpunit-bridge` into their project instead of stock phpunit?

fabpot:
That's already the case, you get a warning when requiring phpunit/phpunit in a Flex project.

weaverryan:
Mind blown :)

I've moved the recipe to the bridge and added an issue on the docs: we need to recommend the bridge there instead of phpunit.

weaverryan:
Bump :). Any voters?

PR #293, opened 2017-12-03
Bump phpunit to 6.5, discourage phpunit/phpunit more

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

phpunit 6.5 is the last to support 7.0, so this is the last bump.

when installing phpunit/phpunit:
image

when installing the bridge:
image

### DIFF
--- a/phpunit/phpunit/4.7/post-install.txt
+++ b/phpunit/phpunit/4.7/post-install.txt
@@ -1,2 +1,7 @@
-Adding <comment>phpunit/phpunit</> as a dependency is <comment>discouraged</>.
-You should require <fg=green>simple-phpunit</> instead.
+<bg=yellow;fg=black>                                                        </>
+<bg=yellow;fg=black> Adding phpunit/phpunit as a dependency is discouraged. </>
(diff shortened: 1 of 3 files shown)


votemike:
Why is having phpunit/phpunit as a dependency discouraged? And what benefit does using the bridge have? Would it be worth explaining these questions in a comment when the message you added is displayed?

OndraM:
@nicolas-grekas Hi, may I kindly have the same question as @votemike? Flex discourages it, but why? The simple-phpunit wrapper complicates things (IDE integration, CI build for static analysis are missing PHPUnit classes etc.), but in fact I don't know why it should be used instead of phpunit directy?

Thanks (and thanks for you great work!)

nicolas-grekas:
Here is what the bridge provides:
https://symfony.com/doc/current/components/phpunit_bridge.html
About IDE integration and static analysis, the bridge still downloads phpunit, so the classes are here. Run `./bin/phpunit install` if you want to be sure this is done.
The bridge also prevents phpunit from messing up with your dependencies, by adding its own constraints to yours.

PR #296, opened 2017-12-06
Fixed default phpunit.xml.dist: add filter/whitelist

lyrixx:
| Q             | A
| ------------- | ---
| License       | MIT

If we don't do that, phpunit is not able to compute the code coverage.
We get the following error without it:

```
Error:         Incorrect whitelist config, no code coverage will be generated.
```


fabpot:
The same change should be applied to the file under `phpunit/phpunit`

lyrixx:
Good catch. It's now fixed.

PR #306, opened 2017-12-11
[WIP][TwigBundle] Improve base template structure

covex-nn:
| Q             | A
| ------------- | ---
| License       | MIT

This PR improves base template to make it more re-usable, all changes are compatible with old base template.

Each `html`, `head`, `stylesheets`, `body` and `javascripts` area consists of three nested levels: `wrapper`, `core` and area itself.

Example:

```twig
{% block body_wrapper %}<body>{% block body_core %}
    {% block body %}{% endblock %}

    {% block javascripts_wrapper %}
        {% block javascripts_core %}
            {% block javascripts %}{% endblock %}
        {% endblock %}
    {% endblock %}
{% endblock %}</body>{% endblock %}
```

With these levels it is easy to add css-class to `<body>` or, for example, to include additional external js/css file between general js/css framework and project's code.

```twig
{% extends 'base.html.twig' %}

{% block body_wrapper %}<body class="example">{{ block('body_core') }}</body>{% endblock %}

{% block javascripts_wrapper %}
    <script type="text/javascript" src="//cdn.js-framework1.com/file1.js"></script>
    {{ parent() }}
{% endblock %}

{% block javascripts_core %}
    <script type="text/javascript" src="//cdn.js-framework2.com/file2.js"></script>
    {{ parent() }}
    <script type="text/javascript">
        jFramework2.alert(1);
    </script>
{% endblock %}

{% block javascripts %}
    {{ parent() }}
    <script type="text/javascript">
        jFramework1.alert(1);
    </script>
{% endblock
```

If my html is simple, i use `stylesheets`, `body` and `javascripts`, otherwise if html is more complex - i can use `*_wrapper` and `*_core` blocks without rewriting base template.

### DIFF
--- a/symfony/twig-bundle/3.3/templates/base.html.twig
+++ b/symfony/twig-bundle/3.3/templates/base.html.twig
@@ -1,12 +1,25 @@
 <!DOCTYPE html>
-<html>
-    <head>
-        <meta charset="UTF-8">
-        <title>{% block title %}Welcome!{% endblock %}</title>
-        {% block stylesheets %}{% endblock %}
-    </head>
-    <body>
+{% block html_wrapper %}<html>{% block html_core %}
+    {% block head_wrapper %}<head>{% block head_core %}
+        {% block head %}
+            <meta charset="UTF-8">
+            <title>{% block title %}Welcome!{% endblock %}</title>
+        {% endblock %}
+
+        {% block stylesheets_wrapper %}
+            {% block stylesheets_core %}
+                {% block stylesheets %}{% endblock %}
+            {% endblock %}
+        {% endblock %}
+    {% endblock %}</head>{% endblock %}
+
+    {% block body_wrapper %}<body>{% block body_core %}
         {% block body %}{% endblock %}
-        {% block javascripts %}{% endblock %}
-    </body>
-</html>
+
+        {% block javascripts_wrapper %}
+            {% block javascripts_core %}
+                {% block javascripts %}{% endblock %}
+            {% endblock %}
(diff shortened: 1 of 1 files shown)


Pierstoval:
Even though I update the template myself on any app, I don't think it's useful to propose such template structure.

Are you making a mobile-first app? Full mobile-app? Desktop-only app? Responsive app? Will you make a blog with one sidebar at the right? Or at the left? Will you make a one-page application? Will you use Angular to manage the frontend? Or React? Or Vue? Will you have a totally dynamic frontend configurable in a backend?

Every app is different. The "javascripts" and "stylesheets" and "body" parts are fairly enough to tell newcomers that they can decouple assets definition from the body in their templates, the rest is just glue for their frontend needs.

I'm 👎 on this

javiereguiluz:
@covex-nn thanks for helping us improve the Symfony recipes and congrats on your first contribution!

Your proposal is technically correct, and it's true that real pass need to define more Twig blocks than the ones provided by the default template ... but your proposal looks a bit opinionated. It'd be great for complex and traditional web apps, but it would require lots of changes for other kinds of applications.

Of course the same is going to happen with any template we create ... that's why we added the minimum amount of contents in the default template so people can easily customize it to their own needs.

fabpot:
I agree with the previous comment. The base template we provide is really just a simple base template to get you started. It's not meant to be universal. So, I'm closing this PR as I prefer to keep it as simple as possible. Thanks.

covex-nn:
@javiereguiluz @fabpot yes, sure. I understood

PR #308, opened 2017-12-11
Add a recipe for NelmioCorsBundle

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

Adds a recipe as well as an official alias for [NelmioCorsBundle](https://github.com/nelmio/NelmioCorsBundle).

### DIFF
--- a/nelmio/cors-bundle/1.5/config/packages/nelmio_cors.yaml
+++ b/nelmio/cors-bundle/1.5/config/packages/nelmio_cors.yaml
@@ -1,0 +1,9 @@
+nelmio_cors:
+    defaults:
+        origin_regex: true
+        allow_origin: ['%env(CORS_ALLOW_ORIGIN)%']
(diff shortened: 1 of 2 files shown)


Pierstoval:
As CORS is for multi-domains, I'm not sure having one single `CORS_ALLOW_ORIGIN` host is enough.

Why not an array resolved with the 3.4 env resolver? I know the bundle is even compatible with Symfony 2.7+, but it would probably encourage people migrating to FWB 3.4 too?

dunglas:
@Pierstoval thanks to `origin_regex: true`, you can configure several hosts easily (trough a regex).
The env resolver doesn't work because it's not supported yet in the semantic config validation process (there is some ongoing work, but I suggest to go this the regex for now).

Pierstoval:
Ok, didn't know about the resolver not being compatible with this, so yes in this case, a regex is enough 👍

PR #312, opened 2017-12-13
Add local empty git repo by default

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/flex/1.0/git/config
+++ b/symfony/flex/1.0/git/config
@@ -1,0 +1,5 @@
+[core]
+	repositoryformatversion = 0
+	filemode = true
+	bare = false
+	logallrefupdates = true
--- a/symfony/flex/1.0/git/HEAD
+++ b/symfony/flex/1.0/git/HEAD
@@ -1,0 +1,1 @@
+ref: refs/heads/master
--- a/symfony/flex/1.0/manifest.json
+++ b/symfony/flex/1.0/manifest.json
@@ -1,5 +1,6 @@
 {
     "copy-from-recipe": {
-        ".env.dist": ".env.dist"
+        ".env.dist": ".env.dist",
+        "git/": ".git"
     }
 }


javiereguiluz:
Just for reference: Ruby On Rails did this very same change early this year. See https://github.com/rails/rails/pull/27632  But they run `git init` command instead of committing the empty Git files.

nicolas-grekas:
Thanks for the link.
Just tested, works as expected :)

fabpot:
I would move the changes to the symfony/flex recipes instead.

nicolas-grekas:
The flex recipe is for existing projects, whereas the FWB one is mostly for new ones. Are you sure this should be in the recipe?

nicolas-grekas:
@fabpot I added an empty root commit to the repos, to make rebases possible even for the very first commit. This involves adding a binary file in the recipe. It shouldn't be an issue IMHO, but it doesn't work in the end: when I use the recipe, the binary file is corrupted - it's different than the one committed here.
Could that be a bug in Flex, where binary files are not correctly copied for now?

fabpot:
I would remove the root commit. That's too opinionated, and I don't see why we should try to "solve" this issue, which is not related to our mission :)

symfony/flex/1.0/git/config
```diff
@@ -0,0 +1,5 @@
+[core]
+	repositoryformatversion = 0
+	filemode = true
```
stof: this has a drawback: on windows, the default here should be `false` (otherwise, git would change files to executable when committing on windows due to Windows not supporting Unix filemode)

nicolas-grekas: should we just remove the line?

weaverryan:
Docs issue created: https://github.com/symfony/symfony-docs/issues/8891

I'm not even sure if we will need any docs changes for this - but just be careful: when we change a recipe, we change the user's experience using Symfony. And that often means that the experience that the docs are describing need to be updated.

PR #313, opened 2017-12-14
"There are eight types of tasks" is correct?

chihiro-adachi:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -76,7 +76,7 @@ only contain the tasks needed to install and configure the dependency because
 Symfony is smart enough to reverse those tasks when uninstalling and
 unconfiguring the dependencies.
 
-There are eight types of tasks, which are called **configurators**:
+There are seven types of tasks, which are called **configurators**:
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -76,7 +76,7 @@ only contain the tasks needed to install and configure the dependency because
 Symfony is smart enough to reverse those tasks when uninstalling and
 unconfiguring the dependencies.
 
-There are eight types of tasks, which are called **configurators**:
+There are seven types of tasks, which are called **configurators**:
```
javiereguiluz: Technically speaking there are 8 configurators (see https://github.com/symfony/flex/tree/master/src/Configurator) but we don't talk about some of them (e.g. MakefileGenerator and ContainerGenerator). Maybe we can use just `several` instead of `eight` / `seven` and that will always be correct.

fabpot: Indeed, that's better. I've done the change here: a01ee58

chihiro-adachi: @javiereguiluz Thank you ! I understood very well.

PR #315, opened 2017-12-15
Use route.controller instead of route.defaults._controller

ostrolucky:
| Q             | A
| ------------- | ---
| License       | MIT

Recently, after 2+ years working with Symfony I've learned by accident that we don't have to specify controller in `defaults` section. Maybe we should promote it more.

### DIFF
--- a/symfony/routing/3.3/config/routes.yaml
+++ b/symfony/routing/3.3/config/routes.yaml
@@ -1,3 +1,3 @@
 #index:
 #    path: /
-#    defaults: { _controller: 'App\Controller\DefaultController::index' }
+#    controller: 'App\Controller\DefaultController::index'


Pierstoval:
It's only for 3.4+ so it needs a new recipe

ostrolucky:
I have tried it with 3.3 (`composer create-project --profile symfony/skeleton test 3.3` right)? Works fine.

Pierstoval:
That's strange, are you sure to have symfony/routing 3.3 and not 3.4 ?
Because original PR is related to 3.4 => https://github.com/symfony/symfony/pull/23227

ostrolucky:
Ahh ok so that command pulls >3.4 components. So I think you are right. Not worth to create new recipe just for this, gonna close until there is better reason to create it and then merge it there.

PR #316, opened 2017-12-15
[Routing] Use "controller" keyword in example config

mheki:
| Q             | A
| ------------- | ---
| License       | MIT

Since in Symfony 4  we encourage to use `controller` keyword in prefix, an example configuration should also use this keyword instead of `defaults`.

### DIFF
--- a/symfony/routing/3.3/config/routes.yaml
+++ b/symfony/routing/3.3/config/routes.yaml
@@ -1,3 +1,3 @@
 #index:
 #    path: /
-#    controller: App\Controller\DefaultController::index
+#    defaults: { _controller: 'App\Controller\DefaultController::index' }
--- a/symfony/routing/3.3/config/routes.yaml
+++ b/symfony/routing/3.3/config/routes.yaml
@@ -1,3 +1,3 @@
 #index:
 #    path: /
-#    defaults: { _controller: 'App\Controller\DefaultController::index' }
+#    controller: App\Controller\DefaultController::index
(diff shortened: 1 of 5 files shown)


weaverryan:
Definitely! Except that this is technically the 3.3 recipe... and that only works on 3.4. But, we have at least one place where we’ve made 3.4-specific changes. So +1 from me

Pierstoval:
I don't think it's ok to merge it. It's 3.4 only, so a recipe for 3.4 should be created.

mheki:
Yeah, that's what I thought. There was no separate branch for 3.4/4.0 recipes. Looks like we may need it... WDYT?

weaverryan:
Yea, let’s make a 3.4 directory. That will help force us to make a decision :)

Thanks!

Nyholm:
Looking at the docs: http://symfony.com/doc/3.4/routing.html

Do we really recommend using `controller:`  in 3.4? The docs is for 4.0 first mention `controller` instead of `defaults`. I know both will work Im just adding a comment for consistency.

mheki:
I agree @Nyholm . I wanted to make generated `routes.yaml` file consistent with the 4.0 docs. No need to change anything for 3.4, again, for consistency.

mheki:
I copied the content from `3.3` folder into `4.0` and now it fails on 
```
Alias "router" also defined for "symfony/routing"
```

Feels like there is an additional ground work needed to make different versions of the same recipe

fabpot:
Indeed, that's something I need to fix on my side.

fabpot:
@symfony-flex-server review please

PR #322, opened 2017-12-20
Updating notice wording

votemike:
| Q             | A
| ------------- | ---
| License       | MIT

I found the current wording confusing as to why phpunit/phpunit was being discouraged. Was there a bug in it? A security flaw? Is the recipe bad? After speaking with the community, it turns out it was merely because Symfony's bridge is more useful to the user than vanilla phpunit. I feel this new message is more useful to developers.

### DIFF
--- a/phpunit/phpunit/4.7/post-install.txt
+++ b/phpunit/phpunit/4.7/post-install.txt
@@ -1,5 +1,5 @@
 <bg=yellow;fg=black>                                                                                              </>
-<bg=yellow;fg=black> Adding phpunit/phpunit as a dependency is discouraged in favour of Symfony's PHPUnit Bridge. </>
+<bg=yellow;fg=black> Adding phpunit/phpunit as a dependency is discouraged in favor of Symfony's PHPUnit Bridge. </>
 <bg=yellow;fg=black>                                                                                              </>
 
   * <fg=blue>Instead</>:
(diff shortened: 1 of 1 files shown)


phpunit/phpunit/4.7/post-install.txt
```diff
@@ -1,7 +1,7 @@
-<bg=yellow;fg=black>                                                        </>
-<bg=yellow;fg=black> Adding phpunit/phpunit as a dependency is discouraged. </>
-<bg=yellow;fg=black>                                                        </>
+<bg=yellow;fg=black>                                                                                              </>
+<bg=yellow;fg=black> Adding phpunit/phpunit as a dependency is discouraged in favour of Symfony's PHPUnit Bridge. </>
```
xabbuh: [...] in favor of [...]

votemike: Does Symfony use UK or US spellings?

xabbuh: We always use American English.

phpunit/phpunit/4.7/post-install.txt
```diff
@@ -1,7 +1,7 @@
-<bg=yellow;fg=black>                                                        </>
-<bg=yellow;fg=black> Adding phpunit/phpunit as a dependency is discouraged. </>
-<bg=yellow;fg=black>                                                        </>
+<bg=yellow;fg=black>                                                                                              </>
```
xabbuh: looks like you now need to fix the alignment (i.e. removing one space here and two lines below)

PR #327, opened 2017-12-22
enable HTML5 e-mail validation

xabbuh:
| Q             | A
| ------------- | ---
| License       | MIT

https://github.com/symfony/symfony/pull/25478#issuecomment-351309384


xabbuh:
There's just one problem here: We not only need the Validator component 4.1, but the FrameworkBundle must be 4.1 or higher too. Should we solve this conflict by adding a PHP config instead why enables this based on the FrameworkBundle version?

Nyholm:
I think the more predictable way is to throw an exception if one uses this config with FrameworkBundle 4.0

PR #329, opened 2017-12-27
Added monolog in test env

lyrixx:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/monolog-bundle/3.1/config/packages/test/monolog.yaml
+++ b/symfony/monolog-bundle/3.1/config/packages/test/monolog.yaml
@@ -1,0 +1,7 @@
+monolog:
+    handlers:
+        main:
+            type: stream
+            path: "%kernel.logs_dir%/%kernel.environment%.log"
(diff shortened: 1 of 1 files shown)


symfony/monolog-bundle/3.1/config/packages/test/monolog.yaml
```diff
@@ -0,0 +1,7 @@
+monolog:
+    handlers:
+        main:
+            type: stream
+            path: "%kernel.logs_dir%/%kernel.environment%.log"
```
lyrixx: Question: Why don't we hardocde the "test.log" filename ? (same in other env)

Nyholm: I cannot come up with a good reason. The result would be the same...

lyrixx: Yes, indeed ;)

PR #330, opened 2017-12-31
Add a recipe for Thanks and display an incentive to use it when installing FrameworkBundle

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

Alternative to symfony/flex#256 and symfony/skeleton#49.

Benefits:

* All people installing Symfony 4 can't miss they **can** thank Symfony, PHP and the other packages they use.
* No extra dependencies installed by default

### DIFF
--- a/symfony/framework-bundle/3.3/post-install.txt
+++ b/symfony/framework-bundle/3.3/post-install.txt
@@ -11,3 +11,7 @@
        Run <comment>composer require server</> for a better web server.
 
   * <fg=blue>Read</> the documentation at <comment>https://symfony.com/doc</>
+
+  * <fg=blue>Thanks</> for using Symfony!
+    Spread some love by starring the GitHub repositories you use!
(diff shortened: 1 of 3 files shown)


symfony/framework-bundle/3.3/post-install.txt
```diff
@@ -11,3 +11,7 @@
        Run <comment>composer require server</> for a better web server.
 
   * <fg=blue>Read</> the documentation at <comment>https://symfony.com/doc</>
+
+  * <fg=blue>Thanks</> for using Symfony!
+    Spread some love by starring the GitHub repositories you use!
+    Run <comment>composer req --dev thanks && composer thanks</>
```
fabpot: Should be run once *all* deps are installed, not just after the initialization. Not sure if all this is really needed or even desirable.

nicolas-grekas:
I propose to replace this PR by https://github.com/symfony/thanks/pull/28

PR #333, opened 2018-01-06
[symfony/framework-bundle] Enable session support by default

### LINKED ISSUE #262: What to do if a bundle (recipe) needs session?
The defaults in framework uncomments the session and thus the session service is not being created. But what can one do if a bundle (recipe) needs the session to be available?

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Enabling session by default has only one practical downside: some classes/services are loaded but never used. Not a big deal.
STILL, this is something that is going to be fixed on Symfony by https://github.com/symfony/symfony/pull/25699, where the session is made extra-lazy.

Thus, this is my proposal to fix #262.

### DIFF
--- a/symfony/framework-bundle/3.3/config/packages/framework.yaml
+++ b/symfony/framework-bundle/3.3/config/packages/framework.yaml
@@ -4,10 +4,10 @@ framework:
     #csrf_protection: ~
     #http_method_override: true
 
-    # uncomment this entire section to enable sessions
-    #session:
-    #    # With this config, PHP's native session handling is used
-    #    handler_id: ~
+    # Enables session support. Note that the session will ONLY be started if you read or write from it.
+    # Remove or comment this section to explicitly disable session support.
+    session:
+        handler_id: ~
 
     #esi: ~
     #fragments: ~
--- a/symfony/framework-bundle/3.3/config/packages/test/framework.yaml
+++ b/symfony/framework-bundle/3.3/config/packages/test/framework.yaml
@@ -1,5 +1,4 @@
 framework:
     test: ~
-    # Uncomment this section if you're using sessions
-    #session:
-    #    storage_id: session.storage.mock_file
+    session:
+        storage_id: session.storage.mock_file


nicolas-grekas:
This can be merged without waiting for symfony/symfony#25699.

Nyholm:
Recipes should have default values that fit 80% of the use cases. 

Looking at [2017 survey](https://symfony.com/blog/symfony-community-survey-2017-results) Im not sure that is the case.

nicolas-grekas:
@Nyholm which part of the survey to you link to?
Anyway, this doesn't really matter: there is no downside to enable session - ie, is does nothing unless you actually use sessions (in which case it has to be enabled.)

symfony/framework-bundle/3.3/config/packages/framework.yaml
```diff
@@ -4,10 +4,9 @@ framework:
     #csrf_protection: ~
     #http_method_override: true
 
-    # uncomment this entire section to enable sessions
-    #session:
-    #    # With this config, PHP's native session handling is used
-    #    handler_id: ~
+    session:
```
sroze: I'd add a comment here to explain that it doesn't harm their application's performance.

Nyholm:
Page 3. 

Okey, fair.

symfony/framework-bundle/3.3/config/packages/framework.yaml
weaverryan: And also perhaps a note about how they could remove this to explicitly disable sessions?

weaverryan:
I opened a docs issue for this

symfony/framework-bundle/3.3/config/packages/framework.yaml
nicolas-grekas: Any suggestions?

sroze:
```yaml
# Enable session support. Note that the session will be started ONLY if you read or write from it.
# Remove or comment this section to disable the session support.
session:
```

weaverryan: 👍 I rudely made a few minor tweaks to your suggestion - hope you don't mind @sroze ;)

Toflar:
> This is awesome. This is currently the biggest DX issue, along with csrf_protection, which I believe we could also “fix” after this, since it’s complicated due to session maybe not being available.

@weaverryan We've switched from session based CSRF protection to the [Double Submit Cookie strategy](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#Double_Submit_Cookie) to avoid relying on the session just for CSRF protection. Just FYI, maybe that would be a good addition to SF in a future version 😄

symfony/framework-bundle/3.3/config/packages/framework.yaml
sroze: That’s perfect. I’ll survive it 😜

nicolas-grekas: updated, thanks

PR #334, opened 2018-01-06
Use .gitkeep instead of .gitignore

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT

Using `.gitkeep` is less confusing and you know that you safely can remove that file. Using an empty `.gitignore` make less sense to me.

From [stackoverflow](https://stackoverflow.com/questions/7229885/what-are-the-differences-between-gitignore-and-gitkeep): 

> .gitkeep isn’t documented, because it’s not a feature of Git.
>
> Git cannot add a completely empty directory. People who want to track empty directories in Git have created the convention of putting files called .gitkeep in these directories. The file could be called anything; Git assigns no special significance to this name.
>
> There is a competing convention of adding a .gitignore file to the empty directories to get them tracked, but some people see this as confusing since the goal is to keep the empty directories, not ignore them; .gitignore is also used to list files that should be ignored by Git when looking for untracked files.


fabpot:
We decided to remove usage of .gitkeep in favor of .gitignore a while ago.

julienfalque:
Out of curiosity, what was the reasoning behind that decision?

Pierstoval:
I think it was because `.gitkeep` is just a convention whereas `.gitignore` is a standard

Nyholm:
Thank you. I did not know that.

PR #335, opened 2018-01-07
suggesting other cache options

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

Hi guys!

I'm open to suggestions to make this better. I thought it would be useful to give people some great caching options just by showing them here. I think it's important because the keys you need to use (e.g. `cache.adapter.apcu`) are abstract. So let's make it as easy as possible.

### DIFF
--- a/symfony/framework-bundle/3.3/config/packages/framework.yaml
+++ b/symfony/framework-bundle/3.3/config/packages/framework.yaml
@@ -15,9 +15,11 @@ framework:
         log: true
 
     cache:
-        # app cache uses the filesystem by default. Other options include:
+        # The app cache caches to the filesystem by default. Other options include:
+
+        # apcu
         # app: cache.adapter.apcu
 
+        # redis
         # app: cache.adapter.redis
-        # configure your redis connnection via:
         # default_redis_provider: redis://localhost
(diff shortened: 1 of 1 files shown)


symfony/framework-bundle/3.3/config/packages/framework.yaml
```diff
@@ -13,3 +13,11 @@ framework:
     #fragments: ~
     php_errors:
         log: true
+
+    cache:
+        # app cache uses the filesystem by default. Other options include:
```
Pierstoval: I'd say `Defaul app cache (...)` instead

symfony/framework-bundle/3.3/config/packages/framework.yaml
```diff
@@ -13,3 +13,11 @@ framework:
...
+        # app: cache.adapter.apcu
+
+        # app: cache.adapter.redis
```
Pierstoval: I'd add a comment above this line saying `To use redis, use the eponym adapter and configure your redis connection:`

symfony/framework-bundle/3.3/config/packages/framework.yaml
```diff
@@ -13,3 +13,13 @@ framework:
     #fragments: ~
     php_errors:
         log: true
+
+    cache:
+        # The app cache caches to the filesystem by default. Other options include:
+
+        # apcu
+        # app: cache.adapter.apcu
```
nicolas-grekas: actually, apcu is not really suited to app caches, because its memory model works best with append only use cases (ie perfect fit for cache.system, already used by default)

PR #336, opened 2018-01-07
PHP CS Fixer: add fixer alias

keradus:
| Q             | A
| ------------- | ---
| License       | MIT


Pierstoval:
Not sure "fixer" is the right word. "cs-fixer" maybe. But not even sure we should have an alias for this package 🤔

friendsofphp/php-cs-fixer/2.2/manifest.json
```diff
@@ -1,4 +1,5 @@
 {
+    "aliases": ["cs-fixer", "fixer", "php-cs-fixer"],
```
fabpot: I would remove `fixer` which sounds too broad.

PR #338, opened 2018-01-10
Minor fixes in the comments of some recipes

javiereguiluz:
| Q             | A
| ------------- | ---
| License       | MIT


Nyholm:
Hm.. It seams like only @fabpot has the power to "merge now".

PR #339, opened 2018-01-10
Lazy load security:check command

Tobion:
| Q             | A
| ------------- | ---
| License       | MIT


xabbuh:
Do we actually need this now that we have sensiolabs/security-checker#102?

Tobion:
It's not required. But symfony framework also defines both ways.

PR #340, opened 2018-01-11
Don’t break when not using env vars

apfelbox:
| Q             | A
| ------------- | ---
| License       | MIT

This is an implementation of https://github.com/symfony/recipes/pull/247#issuecomment-356599423

### Background

In our case we are not using env vars, as they are too cumbersome for our hosting situation.
 With one exception: `APP_ENV`, as on the dev machines it can just default to `dev`, `APP_DEBUG` will nicely map to the correct value with the current default implementation and it is easy to pass the env in a `symfony/console` CLI call.

In this case the `console` and `framework-bundle` could be more lenient.

Current workflow:
1. `composer create-project symfony/skeleton .`
2. `composer remove --dev symfony/dotenv`
3. Exception 💥  (`APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.`)

I am sure this could be more comfortable, in a way that would just work for both ways.

### Implementation

Just use default values, if no explicit env vars are set. Overwrite the env vars with the ones from the `.env` file, if the package is loaded.

This should work for both ways.

* One of the issues sort-of reintroduced is that the nice error message introduce in #247 is gone. Maybe it would be possible to just check for errors during compilation and look for missing parameters. If they are `env(...)` and the `DotEnv` class doesn't exists, show the error message from the previous pr?
* Also I wasn't sure what to do about https://github.com/symfony/recipes/blob/master/behat/symfony2-extension/2.1/features/bootstrap/bootstrap.php#L6-L11

### DIFF
--- a/symfony/console/3.3/bin/console
+++ b/symfony/console/3.3/bin/console
@@ -15,16 +15,14 @@ if (!class_exists(Application::class)) {
     throw new \RuntimeException('You need to add "symfony/framework-bundle" as a Composer dependency.');
 }
 
-if (!isset($_SERVER['APP_ENV'])) {
-    if (!class_exists(Dotenv::class)) {
-        throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
-    }
-    (new Dotenv())->load(__DIR__.'/../.env');
-}
-
 $input = new ArgvInput();
 $env = $input->getParameterOption(['--env', '-e'], $_SERVER['APP_ENV'] ?? 'dev');
 $debug = ($_SERVER['APP_DEBUG'] ?? ('prod' !== $env)) && !$input->hasParameterOption(['--no-debug', '']);
+
+// The check is to ensure we don't use .env in production
+if ($debug && class_exists(Dotenv::class)) {
+    (new Dotenv())->load(__DIR__.'/../.env');
+}
 
 if ($debug) {
     umask(0000);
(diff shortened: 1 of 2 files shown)


symfony/console/3.3/bin/console
```diff
@@ -15,17 +15,15 @@ if (!class_exists(Application::class)) {
...
-    (new Dotenv())->load(__DIR__.'/../.env');
-}
-
 $input = new ArgvInput();
 $env = $input->getParameterOption(['--env', '-e'], $_SERVER['APP_ENV'] ?? 'dev');
 $debug = ($_SERVER['APP_DEBUG'] ?? ('prod' !== $env)) && !$input->hasParameterOption(['--no-debug', '']);
 
+// The check is to ensure we don't use .env in production
+if ($debug && class_exists(Dotenv::class)) {
+    (new Dotenv())->load(__DIR__.'/../.env');
```
stof: this is broken, as you cannot configure the environment or the debug mode in the .env file anymore.

stof:
This forbids using the prod environment locally without reconfiguring your shell to avoid the need to use the dotenv file.

And it also makes it much more likely to deploy the dev environment in your prod server by mistake if you forget to configure `APP_ENV`, which is a security issue (as this would expose the profiler in prod for instance).

apfelbox:
The whole concept (or at least Symfony's view of it) has recently changed, as now `.env` can be used in production.
The simplest solution would be to just merge both by default, so import the `.env` and set all env variables, that are not yet present.

But as the use cases for env variables differ widely between people hosting on one-project-per-host cloud machines (PaaS) and people doing mass-hosting with multiple projects per machine, I don't  see any way how this can be unified so that in the end everybody is happy.

Because some people don't need env vars at all (🙋) and some don't want anything else and especially not even the most minimal performance hit by just testing for the file existence... so I see no way to please anyone.

PR #341, opened 2018-01-12
Fix composer remove phpunit/phpunit command

thewilkybarkid:
| Q             | A
| ------------- | ---
| License       | MIT

Composer just gives a 'phpunit/phpunit could not be found in require but it is present in require-dev' notice without the `--dev` switch.

(Assumes PHPUnit isn't being used in production of course. 😉)

### DIFF
--- a/phpunit/phpunit/4.7/post-install.txt
+++ b/phpunit/phpunit/4.7/post-install.txt
@@ -3,5 +3,5 @@
 <bg=yellow;fg=black>                                                                                             </>
 
   * <fg=blue>Instead</>:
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I agree with this change, because most of those having `phpunit` as a requirement will have it in `require-dev`. Moreover, if `phpunit` is not in `require-dev`, Composer will display a message telling you about that and offering you to remove it from `require`, so the result will be the same.

Nyholm:
Oh, you beat me with a second.

PR #344, opened 2018-01-18
Clear cache and warm up properly in symfony/framework-bundle

cilefen:
| Q             | A
| ------------- | ---
| License       | MIT

`cache:clear --no-warmup` [seems not to have been removed](https://symfony.com/blog/new-in-symfony-3-3-deprecated-cache-clear-with-warmup) in SF4. Nevertheless, composer should clear then warm the cache properly as per the [deployment guide](https://symfony.com/doc/current/deployment.html).

### DIFF
--- a/symfony/framework-bundle/3.3/manifest.json
+++ b/symfony/framework-bundle/3.3/manifest.json
@@ -8,7 +8,8 @@
         "src/": "%SRC_DIR%/"
     },
     "composer-scripts": {
-        "cache:clear": "symfony-cmd",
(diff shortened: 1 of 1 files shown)


cilefen:
Perhaps this is unnecessary because I found [this update to cache:clear](https://github.com/symfony/symfony/commit/8b88d9fc364644402d1283462f17b7a2197fe518) in symfony/symfony commit log.

fabpot:
Indeed, we first deprecated the option and revert this change. So, the current code is better. Thanks.

PR #347, opened 2018-01-22
Correctly cast $_SERVER['APP_DEBUG'].

kevin-lot:
If .env defines APP_DEBUG parameter, APP_DEBUG is populated in string type in $_SERVER. String "false" will be casted in boolean true.
So, if you define APP_DEBUG=false in .env, your debug is enable.

| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/console/3.3/bin/console
+++ b/symfony/console/3.3/bin/console
@@ -24,7 +24,9 @@ if (!isset($_SERVER['APP_ENV'])) {
 
 $input = new ArgvInput();
 $env = $input->getParameterOption(['--env', '-e'], $_SERVER['APP_ENV'] ?? 'dev');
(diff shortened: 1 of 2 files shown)


stof:
the expected usage of this env variable is to put `1` or `0`. I think your change is breaking this.

curry684:
Given that this change does not fix any actual problems, and instead introduces a performance degradation on *every* page view, I'm 👎 

For the record: the valid values are documented at https://symfony.com/doc/current/configuration/environments.html.
> Important, but unrelated to the topic of environments is the second argument to the Kernel constructor. This specifies if the application should run in "debug mode". Regardless of the environment, a Symfony application can be run with debug mode set to true or false (respectively 1 or 0 for the APP_ENV variable defined in .env).

fabpot:
Closing as I agree with the other commenters.

PR #353, opened 2018-01-27
tweak internal web server arguments

xabbuh:
| Q             | A
| ------------- | ---
| License       | MIT

It looks like only passing the document root when starting PHP's internal web server behaves slightly different compared to the process that is bootstrapped when using the WebServerBundle (see symfony/symfony#25911). This changes the command suggestion to be (almost) the same of what is done when using the `server:start` command.

### DIFF
--- a/symfony/framework-bundle/3.3/post-install.txt
+++ b/symfony/framework-bundle/3.3/post-install.txt
@@ -4,7 +4,7 @@
 
   * <fg=blue>Run</> your application:
     1. Change to the project directory
-    2. Execute the <comment>php -S 127.0.0.1:8000 -t public</> command;
+    2. Execute the <comment>php -S 127.0.0.1:8000 public/index.php</> command;
     3. Browse to the <comment>http://localhost:8000/</> URL.
 
        Quit the server with CTRL-C.


Nyholm:
Looks good, Waiting for @mrljaime response on the original issue.

xabbuh:
@mrljaime confirmed in https://github.com/symfony/symfony/issues/25911#issuecomment-361370602

symfony/framework-bundle/3.3/post-install.txt
```diff
@@ -4,7 +4,7 @@
 
   * <fg=blue>Run</> your application:
     1. Change to the project directory
-    2. Execute the <comment>php -S 127.0.0.1:8000 -t public</> command;
+    2. Execute the <comment>php -S 127.0.0.1:8000 public/index.php</> command;
```
stof: I think we still need to set the document root to `public`, otherwise loading assets won't work.

The server:start command does not use the option, because the default behavior of PHP is to use the CWD as document root, and the command runs the process with the `public` folder as working directory.

weaverryan: Yea, I just had a user more or less hit this issue I think.

PR #354, opened 2018-01-29
[WIP][RFC] Stop using ~ in config files

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

This is just an RFC. One question I've gotten routinely over the years is "What is the `~`?" in config. I think it would be more clear if we just used `null` instead. Or, when the "presence" of a key is activating a component, use `true`, which reads a bit more clear.

Any major objections? I would propose making this change across *all* of the config files and making it a best-practice going forward.

### DIFF
--- a/symfony/framework-bundle/3.3/config/packages/framework.yaml
+++ b/symfony/framework-bundle/3.3/config/packages/framework.yaml
@@ -1,16 +1,16 @@
 framework:
     secret: '%env(APP_SECRET)%'
     #default_locale: en
-    #csrf_protection: ~
+    #csrf_protection: true
     #http_method_override: true
 
     # Enables session support. Note that the session will ONLY be started if you read or write from it.
     # Remove or comment this section to explicitly disable session support.
     session:
-        handler_id: ~
+        handler_id: null
 
-    #esi: ~
-    #fragments: ~
+    #esi: true
+    #fragments: true
     php_errors:
         log: true
 


Nyholm:
Hm, I've never really given this much thought. You are correct that it is tricky that `~` means both `null` and "Please enable this section". I think it is a good idea to be more explicit in our config.

javiereguiluz:
@weaverryan @Nyholm see #144 for a related discussion.

weaverryan:
Ah, fascinating! Here's a comment from fabpot:

> Of course, there is no difference at all from a code point of view. I was referring as DX here: when one wants to express: "give me the default behavior/config", I think ~ is better than null.

When we use it like `csrf_protection: ~`, I would argue that we are not saying "give me the default behavior/config", but rather: "please *activate* this functionality". For these cases, true looks better:

```yml
# activates CSRF protection
csrf_protection: ~

# does the same... but I think it's a bit more obvious :)
csrf_protection: true
```

@fabpot Can I bother you about this one? You and I apparently disagree :). You can help settle this with your opinion one way or another.

fabpot:
I still prefer `~` to `null`, but if `true` works as well, this is indeed even better.

weaverryan:
I'll finish this (finding any `~` that indicate "activation" and changing to true), and then we can get a vote.

weaverryan:
This is done! There are other `~`, but they usually mean null. I went for the easy wins: where `true` is really more descriptive. All of these situations DO work.

weaverryan:
Failure is false

symfony/framework-bundle/3.3/config/packages/framework.yaml
```diff
@@ -1,16 +1,16 @@
...
     #default_locale: en
-    #csrf_protection: ~
+    #csrf_protection: true
     #http_method_override: true
 
     # Enables session support. Note that the session will ONLY be started if you read or write from it.
     # Remove or comment this section to explicitly disable session support.
     session:
-        handler_id: ~
+        handler_id: null
```
gharlan: Is this intended? (It is the only change of `~` to `null`..)

weaverryan:
This is approved, but will need a force merge :)

fabpot:
merge now

PR #356, opened 2018-01-31
[framework-bundle] Let glob figure out existing dirs and fix resource tracking

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

By using `is_dir()` directly, we prevent correct resource tracking.
Let's delegate that to glob-tracking resources.
The brackets are required to tell our glob-resources where the static prefix stops (making the part after it optional - ie the folders/files don't have to exist.)

### DIFF
--- a/symfony/framework-bundle/3.3/src/Kernel.php
+++ b/symfony/framework-bundle/3.3/src/Kernel.php
@@ -39,23 +39,19 @@ class Kernel extends BaseKernel
         $container->setParameter('container.autowiring.strict_mode', true);
         $container->setParameter('container.dumper.inline_class_loader', true);
         $confDir = $this->getProjectDir().'/config';
-        $loader->load($confDir.'/packages/*'.self::CONFIG_EXTS, 'glob');
-        if (is_dir($confDir.'/packages/'.$this->environment)) {
-            $loader->load($confDir.'/packages/'.$this->environment.'/**/*'.self::CONFIG_EXTS, 'glob');
-        }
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
ping @symfony/deciders

symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -39,23 +39,19 @@ protected function configureContainer(ContainerBuilder $container, LoaderInterfa
...
-        $loader->load($confDir.'/packages/*'.self::CONFIG_EXTS, 'glob');
-        if (is_dir($confDir.'/packages/'.$this->environment)) {
-            $loader->load($confDir.'/packages/'.$this->environment.'/**/*'.self::CONFIG_EXTS, 'glob');
-        }
-        $loader->load($confDir.'/services'.self::CONFIG_EXTS, 'glob');
-        $loader->load($confDir.'/services_'.$this->environment.self::CONFIG_EXTS, 'glob');
+
+        $loader->load($confDir.'/{packages}/*'.self::CONFIG_EXTS, 'glob');
+        $loader->load($confDir.'/{packages}/'.$this->environment.'/**/*'.self::CONFIG_EXTS, 'glob');
+        $loader->load($confDir.'/{services}'.self::CONFIG_EXTS, 'glob');
```
stof: are you sure we need braces for this case ? It is not a directory name by a file name here

nicolas-grekas: I think you're right, it's not mandatory. To be verified if you want to remove them (but it doesn't hurt either :) )

PR #359, opened 2018-02-02
Adding access_control examples and small warning

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

Hi guys!

A friend of mine made a mistake in his `access_control` because he didn't realize that only the *first* access control that matches is used. I thought the best way to make this easier would be to add a few examples and some very small docs.

I've checked that leaving `access_control` empty like this is "ok" - the configuration system just treats it like an empty array.

### DIFF
--- a/symfony/security-bundle/3.3/config/packages/security.yaml
+++ b/symfony/security-bundle/3.3/config/packages/security.yaml
@@ -16,3 +16,9 @@ security:
 
             # form_login: ~
             # https://symfony.com/doc/current/security/form_login_setup.html
+
+    # easy way to control access for large sections of your site
+    # note: only the *first* access control that matches will be used
+    access_control:
+        # - { path: ^/admin, roles: ROLE_ADMIN }
+        # - { path: ^/profile, roles: ROLE_USER }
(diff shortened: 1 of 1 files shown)


weaverryan:
Failure is false I believe.

symfony/security-bundle/3.3/config/packages/security.yaml
```diff
@@ -16,3 +16,9 @@ security:
 
             # form_login: ~
             # https://symfony.com/doc/current/security/form_login_setup.html
+
+    # easy way to control access for large sections of your site
```
Nyholm: `s/easy/Easy/`

symfony/security-bundle/3.3/config/packages/security.yaml
```diff
@@ -16,3 +16,9 @@ security:
...
+    # note: only the *first* access control that matches will be used
```
Nyholm: `s/note/Note/`
`s/only/Only/`

weaverryan:
Good call - tweaks made!

weaverryan:
We’ll need to force merge this. Do you have that super power Tobias?

Nyholm:
I do not I’m afraid. 
I can click the merge but I’m not sure how that effects the CI/bot will handle that. :/

Only @fabpot can force the bot to merge now.

PR #361, opened 2018-02-05
[doctrine-bundle] add single quotes for DATABASE_URL

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

PR #363, opened 2018-02-08
Revert quotes: breaks database in phpunit.xml.dist

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

This reverts sha:ba7fcf10ccb0cc84f16e029dc0bf6a6509680f3b - #361.

That originally was meant to solve: https://github.com/symfony/symfony/issues/26056

The problem is that the change breaks the database in `phpunit.xml.dist`. It ends up like this:

```
<env name="DATABASE_URL" value="'mysql://root:@127.0.0.1:3306/test_maker'"/>
```

And so, because the string starts with a `'`, it seems to parse this as the database name - I get>

> Unknown database 'mysql://root:@127.0.0.1:3306/test_maker''

### DIFF
--- a/doctrine/doctrine-bundle/1.6/manifest.json
+++ b/doctrine/doctrine-bundle/1.6/manifest.json
@@ -10,6 +10,6 @@
         "#1": "Format described at http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url",
         "#2": "For an SQLite database, use: \"sqlite:///%kernel.project_dir%/var/data.db\"",
         "#3": "Configure your db driver and server_version in config/packages/doctrine.yaml",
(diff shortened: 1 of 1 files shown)


Nyholm:
Ping @nicolas-grekas

nicolas-grekas:
Or tweak flex to remove the quotes (double or single) when for the XML file?

Nyholm:
Yeah. That sounds better than merging this PR. 
Because I’m not sure how else to fix https://github.com/symfony/symfony/issues/26056 if we revert #361

sroze:
> Or tweak flex to remove the quotes (double or single) when for the XML file?

This feels more hacky to me... Maybe flex simply single escape all the values containing non `[a-z0-9]+` values when dumping it in the `.env` files.

fabpot:
@sroze This is exactly what Flex does: it escapes depending on the context (XML or .env). If not, then, there is a bug somewhere.

PR #364, opened 2018-02-09
Configure bundles with true/false values instead of NULL/anything else values

miguelsimoes:
| Q             | A
| ------------- | ---
| License       | MIT

Ensures that the Kernel only initializes bundles that have a true value associated with the the environment. At the moment if the value is true/false it will be enable and only "NULL" will not enable the bundle, however it is clearer/simpler to have a NULL/false not enabling the bundle and only true values to enable it.

### DIFF
--- a/symfony/framework-bundle/3.3/src/Kernel.php
+++ b/symfony/framework-bundle/3.3/src/Kernel.php
@@ -28,7 +28,7 @@ class Kernel extends BaseKernel
     {
         $contents = require $this->getProjectDir().'/config/bundles.php';
         foreach ($contents as $class => $envs) {
-            if (isset($envs['all']) || isset($envs[$this->environment])) {
+            if ((isset($envs['all']) && $envs['all']) || (isset($envs[$this->environment]) && $envs[$this->environment])) {
                 yield new $class();
             }
         }


maxhelias:
Same of #417

maxhelias:
Currently there is a check on the presence of the key.
I think, It should be checked also on the value of this key.

What do you think ?

stof:
@maxhelias but if you care about the values of the key, what do you do in the case I gave, where the `all` key enables it, but a more specific `prod` key disables it ? Your current PR would enable the bundle.

maxhelias:
I see what you want mean..
In your case the logic, would be to have the bundle on the environment of dev and test

stof:
> but still I think isset() has a better performance that in_array() function.

yeah, exactly

maxhelias:
We need something that has this behavior
`!empty($envs[$this->environment] || (!empty($envs['all']) && !empty($envs[$this->environment]))`

NAYZO:
what is made in this PR: https://github.com/symfony/recipes/pull/417 fix a real bug when loading bundles by environment

fabpot:
Values are not relevant, only the keys. This was done for performance reasons (to be able to use `isset`). Removing an env is just a matter of removing the key.

maxhelias:
@fabpot, Would not it be relevant in this case to write down this little tips of advice somewhere on the documentation?

If so, in which topic could we contextualize it?

PR #366, opened 2018-02-11
Adding test bootstrap.php file so .env vars are read

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

Fixes symfony/flex#251

The problem is that currently, environment variables must be completely duplicated between `.env` and `phpunit.xml.dist`. That's just unacceptable DX to put in the user... which includes me - I don't want to do that :).

This is an easy win:

* `.env` is read in the same way as `bin/console` and `public/index.php`
* Environment variables in `phpunit.xml.dist` WIN over `.env`. This is used to set the environment to `test`.

### DIFF
--- a/phpunit/phpunit/4.7/tests/bootstrap.php
+++ b/phpunit/phpunit/4.7/tests/bootstrap.php
@@ -1,0 +1,22 @@
+<?php
+
+require __DIR__.'/../vendor/autoload.php';
+
+use Symfony\Component\Dotenv\Dotenv;
+
+/*
+ * Environment variables can also be specified in phpunit.xml.dist.
+ * Those variables will override any defined in .env.
+ */
+if (!isset($_SERVER['APP_ENV'])) {
+    if (!class_exists(Dotenv::class)) {
+        throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
+    }
+    (new Dotenv())->load(__DIR__.'/../.env');
+}
+
+$debug = $_SERVER['APP_DEBUG'] ?? true;
+
+if ($debug) {
+    umask(0000);
+}
--- a/symfony/phpunit-bridge/3.3/tests/bootstrap.php
+++ b/symfony/phpunit-bridge/3.3/tests/bootstrap.php
@@ -1,0 +1,22 @@
+<?php
+
+require __DIR__.'/../vendor/autoload.php';
+
+use Symfony\Component\Dotenv\Dotenv;
+
+/*
+ * Environment variables can also be specified in phpunit.xml.dist.
+ * Those variables will override any defined in .env.
+ */
+if (!isset($_SERVER['APP_ENV'])) {
+    if (!class_exists(Dotenv::class)) {
+        throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
+    }
+    (new Dotenv())->load(__DIR__.'/../.env');
+}
+
+$debug = $_SERVER['APP_DEBUG'] ?? true;
+
+if ($debug) {
+    umask(0000);
+}
--- a/phpunit/phpunit/4.7/phpunit.xml.dist
+++ b/phpunit/phpunit/4.7/phpunit.xml.dist
@@ -5,16 +5,16 @@
          xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.1/phpunit.xsd"
          backupGlobals="false"
          colors="true"
-         bootstrap="vendor/autoload.php"
+         bootstrap="tests/bootstrap.php"
 >
     <php>
         <ini name="error_reporting" value="-1" />
         <env name="KERNEL_CLASS" value="App\Kernel" />
+        <env name="SHELL_VERBOSITY" value="-1" />
+
+        <!-- override or set env variables for the test env here -->
         <env name="APP_ENV" value="test" />
         <env name="APP_DEBUG" value="1" />
-        <env name="APP_SECRET" value="s$cretf0rt3st" />
-        <env name="SHELL_VERBOSITY" value="-1" />
-        <!-- define your env variables for the test env here -->
     </php>
 
     <testsuites>
--- a/symfony/phpunit-bridge/3.3/phpunit.xml.dist
+++ b/symfony/phpunit-bridge/3.3/phpunit.xml.dist
@@ -5,16 +5,16 @@
          xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.1/phpunit.xsd"
          backupGlobals="false"
          colors="true"
-         bootstrap="vendor/autoload.php"
+         bootstrap="tests/bootstrap.php"
 >
     <php>
         <ini name="error_reporting" value="-1" />
         <env name="KERNEL_CLASS" value="App\Kernel" />
+        <env name="SHELL_VERBOSITY" value="-1" />
+
+        <!-- override or set env variables for the test env here -->
         <env name="APP_ENV" value="test" />
         <env name="APP_DEBUG" value="1" />
-        <env name="APP_SECRET" value="s$cretf0rt3st" />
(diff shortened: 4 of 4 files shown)


phpunit/phpunit/4.7/tests/bootstrap.php
```diff
@@ -0,0 +1,22 @@
...
+
+require __DIR__.'/../vendor/autoload.php';
+
+use Symfony\Component\Dotenv\Dotenv;
+
+/*
+ * Environment variables can also be specified in phpunit.xml.dist.
+ * Those variables will override any defined in .env.
+ */
+if (!isset($_SERVER['APP_ENV'])) {
```
quentinus95: Just by curiosity, why do we use this condition, also in the console? This disallow from having a generic bash script with a conditional behavior depending of the APP_ENV, and then launch specific commands based on this. It does not make sense in my opinion that `APP_ENV=dev ./bin/console s:r` launches the server without reading the `.env` file while `./bin/console s:r` does. I will have the same issue here as I may set the `APP_ENV` var to `test` before launching the tests (I actually do) ; it won't load the `.env` file consequently...

I would expect that the `.env` file is always loaded (except if `APP_ENV` is defined and set to `prod` so we can prevent misuses of `.env` files), if it exists. The DotEnv component already takes care to avoid overriding already defined variables.

fabpot:
@weaverryan You should also create a companion PR on symfony/flex to remove the addition of env vars in `phpunit.xml.dist`.

ogizanagi:
@weaverryan: That's already a great improvement, but this doesn't solve another linked issue that should also be considered to me: running a command in another env, including `test` one, which is sometimes handy to debug your test suite, or even simply to init the database/other stuff (despite it can also be done from the test suite bootstrap file).

If it can be of any help, here is what I'm using currently on flex projects: A [DotEnvLoader](https://gist.github.com/ogizanagi/84f6e9c306e1dfac7c66ecffce06b858#file-dotenvloader-php) class allows to load an extra `.env.[env]` file (in addition to the `.env` one), where "[env]" is the symfony env. Other patches inside the gist shows how to use it.

weaverryan:
Hey guys!

I've just made this PR more complicated :). I've change to load the `.env` file *even* if the `APP_ENV`. The only time we *don't* load it is in the `prod` environment.

I think this is important: we continue to see users confused about why the `.env` variable suddenly isn't read if they set `APP_ENV`. The previous situation *also* made this PR strange: the `phpunit.xml.dist` env vars are loaded FIRST, and *then* `bootstrap.php` is executed. But, the `phpunit.xml.dist` env vars are ONLY added to `$_ENV`, not `$_SERVER`. So, `.env` was still loaded (which I liked). But if you suddenly started setting true env vars, you'd get a different behavior.

So the goal is simple: load the `.env` file in the *least* WTF kind of way. Because it doesn't override existing env vars, I think this makes sense.

Also, @ogizanagi brings up another point in his [comment](https://github.com/symfony/recipes/pull/366#issuecomment-364778024). There's not an easy way, for example, to set a different `DATABASE_URL` in test, and be able to use it in your tests and in `bin/console`. He suggests the possibility of having a `.env.[env]` file that would override the normal one, and I agree. It's also what Laravel does, which I think is important, because they've been using the `.env` idea longer so have probably learned a few things.

So.... I think we need to make several changes to our `.env` handling.... and so possibly some sort of `bootstrap.php` file that's used by all entry points - `index.php`, `console`, testing.

Cheers!

quentinus95:
I guess the possibility to use `.env.[env]` files is mitigated by the fact that we can now load the `.env` file even if some environment files are already loaded. One could easily create custom `.env.test` file (for instance) and load those using a source + allexport in a bash script and then run the tests normally.

ogizanagi:
What if `symfony/flex` provided something like the `DotEnvLoader` class mentioned earlier (or whatever name we use for it) and use it here ?

SergeC:
Maybe we should have a `.env.[env].dist` which will be updated by SF Flex the same way as `.env.dist`? The `.env.[env].dist` will be pushed to GIT while `.env.[env]` will be ignored. If env=prod the use `.env`.

bocharsky-bw:
Probably this file also should be updated: https://github.com/symfony/recipes/blob/master/behat/symfony2-extension/2.1/features/bootstrap/bootstrap.php#L6

phpunit/phpunit/4.7/tests/bootstrap.php
```diff
@@ -0,0 +1,23 @@
...
+require __DIR__.'/../vendor/autoload.php';
+
+use Symfony\Component\Dotenv\Dotenv;
+
+/*
+ * Environment variables can also be specified in phpunit.xml.dist.
+ * Those variables will override any defined in .env.
+ */
+
+if (!isset($_SERVER['APP_ENV']) || 'prod' !== $_SERVER['APP_ENV']) {
```
bocharsky-bw: Why don't we use `getenv('APP_ENV')` instead of `$_SERVER['APP_ENV']`. Are there any reasons behind? Symfony recommends to use `getenv()` in docs but has `$_SERVER[]` in code which is not consistent as for me.

ogizanagi: See https://github.com/symfony/recipes/pull/156 about that. It should probably be updated in the docs then.

symfony/phpunit-bridge/3.3/tests/bootstrap.php
```diff
@@ -0,0 +1,22 @@
...
+
+require __DIR__.'/../vendor/autoload.php';
+
+use Symfony\Component\Dotenv\Dotenv;
+
+/*
+ * Environment variables can also be specified in phpunit.xml.dist.
+ * Those variables will override any defined in .env.
+ */
+if (!isset($_SERVER['APP_ENV'])) {
```
Nyholm: Shouln't this be the same as for version 4.7?

`if (!isset($_SERVER['APP_ENV']) || 'prod' !== $_SERVER['APP_ENV']) {`

symfony/phpunit-bridge/3.3/tests/bootstrap.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+require __DIR__.'/../vendor/autoload.php';
+
+use Symfony\Component\Dotenv\Dotenv;
+
+/*
+ * Environment variables can also be specified in phpunit.xml.dist.
+ * Those variables will override any defined in .env.
```
Nyholm: Good!

symfony/framework-bundle/3.3/public/index.php
```diff
@@ -8,7 +8,7 @@
 require __DIR__.'/../vendor/autoload.php';
 
 // The check is to ensure we don't use .env in production
-if (!isset($_SERVER['APP_ENV'])) {
+if (!isset($_SERVER['APP_ENV']) || 'prod' !== $_SERVER['APP_ENV']) {
```
B-Galati: I just had an issue trying to warm-up my cache on the CI.
My `APP_ENV` was `test` and I was overriding the env with the `--env` option like so :

```
bin/console --env=prod cache:clear --no-warmup
[08-Mar-2018 16:54:49 Europe/Paris] PHP Fatal error:  Uncaught RuntimeException: APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file. in /app/bin/console:20
```

I can see that my case is kind of weird and I am going to change the way I ran the command but we could do something to improve DX in that case perhaps ?

sstok:
To be clear phpunit follows the same convention as the DotEnv component, if an env is already set it will not overwrite, unless you provide `force="true"`.

joshlopes:
A DotEnvLoader from @ogizanagi seems what SF needs - why isn't this shown any love? Is there a workaround for this already implemented?

weaverryan:
It’s just waiting on me to find some time to finish it. I have some agreement that this is probably the right approach

phpunit/phpunit/4.7/tests/bootstrap.php
stof: `getenv` is an issue because of not being thread safe (probably less an issue in the PHPUnit bootstrap as this is used only in the CLI)

PR #369, opened 2018-02-19
Tell user which lines can be deleted when using Symfony 4.0+

B-Galati:
| Q             | A
| ------------- | ---
| License       | MIT

I asked myself these questions and was not able to find a quick / clear answer.
If someone have references to add, it would be cool.


symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -36,6 +36,7 @@ public function registerBundles()
 
     protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader)
     {
+        // Feel free to remove the 1st parameter if you are using Symfony 4.0+
```
stof: I would rather put the name instead of saying `1st`. It will be more robust (especially if someone removes it without removing the comment)

B-Galati:
@stof done thank you

symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -36,6 +36,7 @@ public function registerBundles()
 
     protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader)
     {
+        // Feel free to remove the first parameter if you are using Symfony 4.0+ as it's the default behavior
```
Pierstoval: I think @stof was more recommending something like `Feel free to remove the "container.autowiring.strict_mode" parameter (...)` instead of replacing `1st` by `first` :smile:

B-Galati: Hahaha ! Thank you @Pierstoval :-)

symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -36,6 +36,8 @@ public function registerBundles()
 
     protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader)
     {
+        // Feel free to remove the "container.autowiring.strict_mode" parameter
+        // if you are using Symfony 4.0+ as it's the default behavior
```
xabbuh: maybe we should rather talk about `symfony/dependency-injection` than Symfony in general because that's the component that matters for this parameter

B-Galati: Done thank you

PR #371, opened 2018-02-24
Moved the EasyAdmin recipe to its new vendor

javiereguiluz:
| Q             | A
| ------------- | ---
| License       | MIT

We've moved the Composer package from `javiereguiluz/easyadmin-bundle` to `easycorp/easyadmin-bundle` ... so I guess we need to update the recipes too. I don't know if I must also delete the entire `javiereguiluz/` directory.


javiereguiluz:
@Nyholm yes! The old package is abandoned on Packagist.org in favor of the new one (https://packagist.org/packages/javiereguiluz/easyadmin-bundle).

PR #372, opened 2018-02-25
Add some aliases for dependency-injection

chalasr:
| Q             | A
| ------------- | ---
| License       | MIT


fabpot:
I don't see when this would help as the component is part of the base installation.

chalasr:
Right

PR #374, opened 2018-02-26
lexik/jwt-auth adds gitignore to manifest

aeimer:
| Q             | A
| ------------- | ---
| License       | MIT

Adds `/var/jwt/` to gitignore for the lexik/jwt-auth-bundle. This will prevent developers from commiting the certificates, which are placed under `/var/jwt/[private|public].pem`


aeimer:
We may consider ignoring `config/jwt/*.pem` which is the path used in https://github.com/chalasr/lexik-jwt-authentication-sandbox/tree/flex 

Maybe @chalasr can answer this question

Nyholm:
Doing config/jwt/*.pem makes more sense since ‘var’ is already ignored. 

I’m happy with this PR. @chalasr, approve it if you like it.

PR #377, opened 2018-03-08
Fix TypeError argument when APP_DEBUG is set

Plopix:
| Q             | A
| ------------- | ---
| License       | MIT

When you specify APP_DEBUG in the ENV, it is interpreted as a "_string_" which triggers this kind of error:

```
Fatal error: Uncaught TypeError: Argument 2 passed to Symfony\Component\HttpKernel\Kernel::__construct() must be of the type boolean, 
string given, called in /app/public/index.php on line 49 and defined in /app/vendor/symfony/http-kernel/Kernel.php:76
 Stack trace: 
#0 /app/public/index.php(49): Symfony\Component\HttpKernel\Kernel->__construct('prod', '0') 
#1 {main} thrown in /app/vendor/symfony/http-kernel/Kernel.php on line 76
```
This PR aims to solve that issue.

> unless I missed a trick

### DIFF
--- a/symfony/console/3.3/bin/console
+++ b/symfony/console/3.3/bin/console
@@ -24,7 +24,7 @@ if (!isset($_SERVER['APP_ENV'])) {
 
 $input = new ArgvInput();
 $env = $input->getParameterOption(['--env', '-e'], $_SERVER['APP_ENV'] ?? 'dev', true);
-$debug = ($_SERVER['APP_DEBUG'] ?? ('prod' !== $env)) && !$input->hasParameterOption('--no-debug', true);
+$debug = (bool) ($_SERVER['APP_DEBUG'] ?? ('prod' !== $env)) && !$input->hasParameterOption('--no-debug', true);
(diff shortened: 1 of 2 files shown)


weaverryan:
@Plopix are you using strict_mode in `index.php`? I don't get this error, but logically, yea, it makes sense, and I think this PR makes sense.

Plopix:
Hey @weaverryan, yes I forgot to mention that. You are right, I have added the `declare(strict_types=1);` in those 2 files.

Etrimi:
I have the following error (on a new install with : composer create-project symfony/website-skeleton Project)
Notice: Undefined index: APP_DEBUG in /var/www/GeskaDev/public/index.php on line 19

Plopix:
@Etrimi, I don't experience that, it is odd. Can we see the `index.php` that you are using? Did you change anything in it?

Etrimi:
problem solved by #381

PR #368, opened 2018-02-18
Doctrine generated SQL unaffected by recipe dbal charset utf8mb4

felixgomez:
| Q             | A
| ------------- | ---
| License       | MIT

Currently the yaml configuration in the doctrine recipe assigns the charset value to `utf8mb4`:

```yaml
doctrine:
    dbal:
        # configure these for your database server
        driver: 'pdo_mysql'
        server_version: '5.7'
        charset: utf8mb4
```

But this does not affect the configuration of doctrine generated SQL as charset and collation of the tables don't inherit this default charset database value.

So SQL generated through `doctrine:schema:update --dump-sql` keeps adding `DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci` by table.

This is true even if mysql server is configured to use utf8mb4 by default or `?charset=utf8mb4` is added to `DATABASE_URL` env variable. 

IMHO it would be better to change the defaults for doctrine recipe so generated SQL really uses utf8mb4 by table.

```yaml
doctrine:
    dbal:
        # configure these for your database server
        driver: 'pdo_mysql'
        server_version: '5.7'
        charset: utf8mb4
        default_table_options:
            charset: utf8mb4
            collate: utf8mb4_unicode_ci
```

PR #384, opened 2018-03-19
[DX] Set default_table_options to match default utf8mb4 charset

keichinger:
| Q             | A
| ------------- | ---
| License       | MIT

The `default_table_options` options should match the default charset `utf8mb4` so the migrations will generate each table with `utf8mb4` + `utf8mb4_unicode_ci` instead of `utf8` + `utf8_unicode_ci`, so users don't have to specify the `table_options` for each entity.


fabpot:
Closing as a duplicate of #368

apfelbox:
@fabpot would it be possible to merge #368 then?

PR #393, opened 2018-03-24
Add npm and Yarn error logs to .gitignore

msheakoski:
By default, it is a good idea to add the npm/Yarn error logs to .gitignore when installing the Encore package so they don't accidentally get committed. There are many common situations which cause an error log to be generated, whether through running commands, or installing Node.js packages. In my case, it was because the Encore command exits with a code of 1 when `yarn run encore --help` is run.

| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/webpack-encore-pack/1.0/manifest.json
+++ b/symfony/webpack-encore-pack/1.0/manifest.json
@@ -8,7 +8,7 @@
     "gitignore": [
         "/node_modules/",
         "/%PUBLIC_DIR%/build/",
-        "/npm-debug.log",
-        "/yarn-error.log"
+        "npm-debug.log",
+        "yarn-error.log"
     ]
 }
--- a/symfony/webpack-encore-pack/1.0/manifest.json
+++ b/symfony/webpack-encore-pack/1.0/manifest.json
@@ -7,6 +7,8 @@
     "aliases": ["encore", "webpack", "webpack-encore"],
     "gitignore": [
         "/node_modules/",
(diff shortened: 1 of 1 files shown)


Nyholm:
Isnt this something that should live in the global gitignore?

chalasr:
@Nyholm Better in the project gitignore I think, that gives a consistent result across environments

chalasr:
Ah, it merged... hope you agree

msheakoski:
I think that these log files correctly belong in the project .gitignore because Encore is an abstraction of webpack/Node.js, so it has a responsibility to manage artifacts created as a result of installing/using the out-of-the-box configuration of the package.

I am doing this in the same spirit as the php-unit, web-server-bundle, and php-cs-fixer recipes. The goal is to create a "happy path" and reduce the amount of configuration/surprises for the end user.

Ideally, you would want to have node_modules, npm-debug.log, and yarn-error.log in your global gitignore already if you do a lot of work with Node.js. Having them in both the project and global gitignore won't cause any problems however.

Nyholm:
You are correct @msheakoski. There is no harm. And I do not mind having them here. I just wanted to have some reasoning behind what entries goes to the local file and what to the global. 

Ie, I would not approve of .DS_Store 

Thank you for your contribution.

chalasr:
> Ie, I would not approve of .DS_Store

Agreed, OS specific files fit perfectly in the global gitignore only

PR #395, opened 2018-03-27
Add `symfony/messenger` recipe

sroze:
| Q             | A
| ------------- | ---
| License       | MIT

- [x] Depends on https://github.com/symfony/symfony/pull/26632.


sroze:
I guess the review is invalid because obviously the tag does not exists 🤔

symfony/messenger/4.1/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "MESSENGER_ADAPTER_DSN": "amqp://guest:guest@localhost:5672/%2f/messages"
+    },
+    "aliases": ["messenger", "amqp"]
```
Pierstoval: I'd also add "messaging" alias, WDYT?

ogizanagi: What about `command_bus` too? Even if that's a bit reductive considering the wider component scope, that might be a good way to discover it.

symfony/messenger/4.1/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "MESSENGER_ADAPTER_DSN": "amqp://guest:guest@localhost:5672/%2f/messages"
```
ogizanagi: Should we really provide this env var value by default? If there is no need for an adapter at first, can't we set it in the recipe to have `null` as default and add a message explaining how to tweak it in order to use the amqp adapter?

I.e:

```yml
# config/packages/messenger.yaml
parameters:
    env(MESSENGER_ADAPTER_DSN): null
```

or commenting `adapters.default: "%env(MESSENGER_ADAPTER_DSN)%"` 
or providing the `MESSENGER_ADAPTER_DSN="amqp://guest:guest@localhost:5672/%2f/messages"` var commented by default, as in https://github.com/symfony/recipes/blob/bd247214b647c7974bb2e21ac1be8854fdbfdc47/symfony/framework-bundle/3.3/manifest.json#L17.

symfony/messenger/4.1/manifest.json
Pierstoval: I like it: when making presentations about flex & hexagonal architecture, it's always badass to say "We need a command bus, so let's do `composer req command_bus` 😄

symfony/messenger/4.1/manifest.json
sroze: I quite like the idea of commented lines for the adapters and environment variable as well. I've updated the PR accordingly 👍

symfony/messenger/4.1/manifest.json
sroze: I'm not sure about "messaging" tbh. `command-bus` might be better as a meta-package (or other package), which will use the Messenger and auto-enable specific middlewares for command buses (see [here](https://github.com/symfony/symfony/issues/26652#issuecomment-375893089))

symfony/messenger/4.1/config/packages/messenger.yaml
```diff
@@ -0,0 +1,9 @@
+framework:
+    messenger:
+        adapters:
+            # Uncomment the following line to enable an adapter named "amqp"
+            # amqp: "%env(MESSENGER_ADAPTER_DSN)%"
+
+        routing:
+            # Route your messages to the senders...
```
Nyholm: Why the mysterious dots at the end ...?

symfony/messenger/4.1/config/packages/messenger.yaml
```diff
@@ -0,0 +1,9 @@
...
+            # 'App\Message\YourMessage': messenger.sender.amqp
```
weaverryan: We're discussing allowing the short adapter name here - e.g. `amqp` (and also service id's, of course). If we do that (I want to), it might be simpler to do that before merging the recipe.

sroze: Fair point. For the records, it's here: https://github.com/symfony/symfony/pull/26909

symfony/messenger/4.1/config/packages/messenger.yaml
sroze: No reason 😄

symfony/messenger/4.1/config/packages/messenger.yaml
```diff
@@ -0,0 +1,9 @@
+framework:
+    messenger:
+        adapters:
+            # Uncomment the following line to enable an adapter named "amqp"
+            # amqp: "%env(MESSENGER_ADAPTER_DSN)%"
+
+        routing:
+            # Route your messages to the senders.
```
fabpot: I would even remove this last dot.

sroze: Here we go, removed (and actually replaced "senders" with "adapters" as it makes more sense now we the shorter name I believe).

symfony/messenger/4.1/config/packages/messenger.yaml
```diff
@@ -0,0 +1,9 @@
+framework:
+    messenger:
+        adapters:
+            # Uncomment the following line to enable an adapter named "amqp"
+            # amqp: "%env(MESSENGER_ADAPTER_DSN)%"
```
yceruto: Single quote around the parameter?

symfony/messenger/4.1/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "#MESSENGER_ADAPTER_DSN": "amqp://guest:guest@localhost:5672/%2f/messages"
+    },
```
ogizanagi: Perhaps [post-install-output](https://github.com/symfony/recipes#post-install-output-configurator) would be a good fit for mentioning the `symfony/serializer-pack` to start using built-transport easily.

yceruto: Maybe a new `symfony/messenger-pack` would help to install both.

sroze: Any phrasing suggestion? :)

symfony/messenger/4.1/config/packages/messenger.yaml
```diff
@@ -0,0 +1,9 @@
+framework:
+    messenger:
+        adapters:
```
sroze: Should be renamed to `transports`

symfony/messenger/4.1/manifest.json
ogizanagi:
```txt
<bg=blue;fg=white>              </>
<bg=blue;fg=white> What's next? </>
<bg=blue;fg=white>              </>

  * You're ready to use the Messenger component. You can define you message buses
    or start using the default one right now by injecting the `messenger.bus.default` service 
    or typehinting `Symfony\Component\Messenger\MessageBusInterface` in your classes.
  * If you need to send messages to your broker, you can benefit from the built-in
    amqp transport by: 

    1. Installing the <info>symfony/serializer-pack</>
    2. Uncommenting the <comment> MESSENGER_ADAPTER_DSN</> env var 
       and `framework.messanger.transport.amqp` config
    3. Routing your messages to the amqp sender.

  * <fg=blue>Read</> the documentation at <comment>https://symfony.com/doc/master/components/messenger.html</>
```

(something like that)

sroze: That's an amazing idea!

symfony/messenger/4.1/post-install.txt
```diff
@@ -0,0 +1,17 @@
+<bg=blue;fg=white>              </>
+<bg=blue;fg=white> What's next? </>
+<bg=blue;fg=white>              </>
+
+  * You're ready to use the Messenger component. You can define your own message buses
+    or start using the default one right now by injecting the <info>messenger.bus.default</info> service
+    or typehinting <info>Symfony\Component\Messenger\MessageBusInterface</info> in your code.
+
+  * If you need to send messages to your broker, you can benefit from the built-in
+    amqp transport by:
```
fabpot: AMQP

symfony/messenger/4.1/post-install.txt
```diff
@@ -0,0 +1,17 @@
...
+
+    1. Installing the <info>symfony/serializer-pack</>
```
fabpot: missing `;` at the end (same for the next bullet)

symfony/messenger/4.1/post-install.txt
```diff
@@ -0,0 +1,17 @@
...
+    2. Uncommenting the <comment> MESSENGER_ADAPTER_DSN</> env var
+       and `framework.messanger.transports.amqp` config
```
fabpot: You need to remove the backticks. Replace them with `<comment>`?

symfony/messenger/4.1/post-install.txt
```diff
@@ -0,0 +1,17 @@
...
+
+  * You're ready to use the Messenger component. You can define your own message buses
+    or start using the default one right now by injecting the <info>messenger.bus.default</info> service
+    or typehinting <info>Symfony\Component\Messenger\MessageBusInterface</info> in your code.
+
+  * If you need to send messages to your broker, you can benefit from the built-in
+    amqp transport by:
+
+    1. Installing the <info>symfony/serializer-pack</>
+    2. Uncommenting the <comment> MESSENGER_ADAPTER_DSN</> env var
```
fabpot: extra whitespace after `<comment>`

sroze:
@fabpot updated and added the `amqp` alias on the `symfony/amqp-pack`.

symfony/messenger/4.1/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "#MESSENGER_ADAPTER_DSN": "amqp://guest:guest@localhost:5672/%2f/messages"
```
fabpot: The `#` looks weird here. Keep in mind that this is going to be used for more than the `.env` file.

I would do something like: `"#1": "MESSENGER_ADAPTER_DSN amqp://guest:guest@localhost:5672/%2f/messages",` instead.

sroze: Well, [that's what we do in the `framework-bundle`](https://github.com/symfony/recipes/blob/bd247214b647c7974bb2e21ac1be8854fdbfdc47/symfony/framework-bundle/3.3/manifest.json#L17) as well.

sroze: I'm not sure what you mean by `#1` as a key 🤔

ogizanagi: We use the same here: https://github.com/symfony/recipes/blob/bd247214b647c7974bb2e21ac1be8854fdbfdc47/symfony/framework-bundle/3.3/manifest.json#L17

should we change it? Not sure to understand what you mean by
> Keep in mind that this is going to be used for more than the .env file.

?

fabpot: We are using the `env` config to update `phpunit.xml` as well. So, using `#` to mean "comment" is wrong. `#1` is a convention to tell the parser that this is a comment, which is then interpreted depending on where we are using the env var.

sroze: > `#1` is a convention to tell the parser that this is a comment, which is then interpreted depending on where we are using the env var.

Alright, I'm going to change it that way; though... is it documented somewhere? 🤔

fabpot:
@symfony-flex-server review

PR #397, opened 2018-04-12
Make it easier to move config

slootjes:
| Q             | A
| ------------- | ---
| License       | MIT

Small change, only removes some duplicate code to make it easier to have a different config dir.

### DIFF
--- a/symfony/framework-bundle/3.3/src/Kernel.php
+++ b/symfony/framework-bundle/3.3/src/Kernel.php
@@ -25,9 +25,19 @@ class Kernel extends BaseKernel
         return $this->getProjectDir().'/var/log';
     }
 
+    public function getConfigDir()
+    {
+        return $this->getProjectDir().'/config';
+    }
+
+    public function getBundlesFile()
+    {
+        return $this->getConfigDir().'/bundles.php';
+    }
+
     public function registerBundles()
     {
-        $contents = require $this->getProjectDir().'/config/bundles.php';
+        $contents = require $this->getBundlesFile();
         foreach ($contents as $class => $envs) {
             if (isset($envs['all']) || isset($envs[$this->environment])) {
                 yield new $class();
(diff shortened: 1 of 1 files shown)


symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -37,25 +47,22 @@ public function registerBundles()
 
     protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader)
     {
-        $container->addResource(new FileResource($this->getProjectDir().'/config/bundles.php'));
+        $container->addResource(new FileResource($this->getBundlesFile()));
         // Feel free to remove the "container.autowiring.strict_mode" parameter
         // if you are using symfony/dependency-injection 4.0+ as it's the default behavior
         $container->setParameter('container.autowiring.strict_mode', true);
         $container->setParameter('container.dumper.inline_class_loader', true);
-        $confDir = $this->getProjectDir().'/config';
```
Pierstoval: Instead I'd suggest to only replace this change by this:

```diff
-        $confDir = $this->getProjectDir().'/config';
+        $confDir = $this->getConfigDir();
```
Having a single var reduces the amount of calls, and therefore improves performances 😉

symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -37,25 +47,22 @@ public function registerBundles()
...
-        $loader->load($confDir.'/{services}_'.$this->environment.self::CONFIG_EXTS, 'glob');
+        $loader->load($this->getConfigDir().'/{packages}/*'.self::CONFIG_EXTS, 'glob');
+        $loader->load($this->getConfigDir().'/{packages}/'.$this->environment.'/**/*'.self::CONFIG_EXTS, 'glob');
+        $loader->load($this->getConfigDir().'/{services}'.self::CONFIG_EXTS, 'glob');
+        $loader->load($this->getConfigDir().'/{services}_'.$this->environment.self::CONFIG_EXTS, 'glob');
     }
 
     protected function configureRoutes(RouteCollectionBuilder $routes)
     {
-        $confDir = $this->getProjectDir().'/config';
```
Pierstoval: Same here: I'd suggest to keep the var for performances

slootjes:
Thanks for the feedback, if Flex doesn't allow to change these paths then indeed it doesn't make sense to make the methods public. Even though the idea behind the PR can't done I do think it's still a small improvement.

symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -25,9 +25,19 @@ public function getLogDir()
...
 
+    private function getConfigDir()
+    {
+        return $this->getProjectDir().'/config';
+    }
+
+    private function getBundlesFile()
+    {
+        return $this->getConfigDir().'/bundles.php';
+    }
```
fabpot: Can you move these methods at the end of the class (we tend to put private methods after public and protected ones).

symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -52,10 +52,19 @@ protected function configureContainer(ContainerBuilder $container, LoaderInterfa
...
         $routes->import($confDir.'/{routes}/'.$this->environment.'/**/*'.self::CONFIG_EXTS, '/', 'glob');
         $routes->import($confDir.'/{routes}'.self::CONFIG_EXTS, '/', 'glob');
     }
+
+    private function getConfigDir()
+    {
+        return $this->getProjectDir().'/config';
+    }
+
+    private function getBundlesFile()
```
stof: IMO, this method should not exist at all. Changing its implementation must never happen, as Flex will keep using `%CONFIG_DIR%/bundles.php`. So I would keep `$this->getConfigDir().'/bundles.php'` inline in both places

PR #399, opened 2018-04-14
Update Webpack Encore version

codedmonkey:
| Q             | A
| ------------- | ---
| License       | MIT


fabpot:
Do we really want forbid people to use older versions? In any case, composer will always install the latest one. So, what's the point?

weaverryan:
Yea, that's fine with me - there's really no point.

BoShurik:
Version `0.19.0` does not satisfy `^0.17.0`. Every time after applying recipe I should change constraint

fabpot:
Ah, we are at version < 1.0. Reopening then.

PR #400, opened 2018-04-17
removing the Symfony 3.3 warning, because that version is no longer supported

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

Simple: Symfony 3.3 is no longer supported. So, no need for this message anymore.


weaverryan:
The check failure is invalid

PR #405, opened 2018-05-09
Described EasyLog Handler activation after install

gquemener:
| Q             | A
| ------------- | ---
| License       | MIT

I believe this PR improves DX when requiring the `easycorp/easy-log-handler` package (and thus the `debug`symfony pack also).

IMHO, there is no solution which could answer the question asked in the `FIXME` comment of the configuration file:
> How to add this configuration automatically without messing up with the monolog configuration?

The only acceptable solution I see is to NOT automatically add the configuration and provide a quick getting started message which will guide developers.

### DIFF
--- a/easycorp/easy-log-handler/1.0/config/packages/dev/easy_log_handler.yaml
+++ b/easycorp/easy-log-handler/1.0/config/packages/dev/easy_log_handler.yaml
@@ -3,7 +3,6 @@ services:
         public: false
         arguments: ['%kernel.logs_dir%/%kernel.environment%.log']
 
-#// FIXME: How to add this configuration automatically without messing up with the monolog configuration?
 #monolog:
 #    handlers:
 #        buffered:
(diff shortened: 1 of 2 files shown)


weaverryan:
But, will simply uncommenting the code work? Doesn’t this conflict or get overridden by the normal monolog configuration?

gquemener:
I've just checked it and enabled the easy log handler just by uncommenting the configuration.

Here are the steps I've performed:

1.  `composer create-project symfony/skeleton my-project`
2.  `composer require log`
3.  `composer require debug`
4.  Uncomment lines in `config/packages/dev/easy_log_handler.yaml`

I haven't really investigated why the `config/packages/dev/monolog.yaml` configuration is not overriding the monolog configuration (it is loaded after the `config/packages/dev/easy_log_handler.yaml` though), but both default configurations are living peacefully side-by-side!

Finally, an alternative solution would be to NOT provide commented configuration but a `post-install` message refering to online documentation. It all depends on what's the "convention" among the other recipes ([Here](https://github.com/symfony/recipes-contrib/pull/320#discussion_r187789844), it seems that commented configuration is not really recommended).

royklutman:
Is there any update on this?

gquemener:
Hello,

I am closing this PR as it is not really needed/required :)

PR #408, opened 2018-05-15
Bootstrap environment 

joshlopes:
Picking up one of @weaverryan branches and helping out glue everything together having in consideration the discussion on his PR #366 and the DotEnvLoad from @ogizanagi.

Below is the copy of that PR.

| Q             | A
| ------------- | ---
| License       | MIT

Fixes https://github.com/symfony/flex/issues/251

WIP
====

Needs to be reviewed and need tests? 

Problem
======

The problem is that currently, environment variables must be completely duplicated between .env and phpunit.xml.dist

Solution
=======

Extend the Kernel to allow different env to be loaded, falling back to the old behaviour ($_SERVER['APP_ENV']). The env files are overwritten, so we always load `.env` and `.env.test` values will override `.env` ones and so on.

.env is read in the same way as bin/console and public/index.php
Environment variables in phpunit.xml.dist WIN over .env.

### DIFF
--- a/phpunit/phpunit/4.7/phpunit.xml.dist
+++ b/phpunit/phpunit/4.7/phpunit.xml.dist
@@ -5,16 +5,16 @@
          xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.1/phpunit.xsd"
          backupGlobals="false"
          colors="true"
-         bootstrap="vendor/autoload.php"
+         bootstrap="tests/bootstrap.php"
 >
     <php>
         <ini name="error_reporting" value="-1" />
         <env name="KERNEL_CLASS" value="App\Kernel" />
+        <env name="SHELL_VERBOSITY" value="-1" />
+
+        <!-- override or set env variables for the test env here -->
         <env name="APP_ENV" value="test" />
         <env name="APP_DEBUG" value="1" />
-        <env name="APP_SECRET" value="s$cretf0rt3st" />
-        <env name="SHELL_VERBOSITY" value="-1" />
-        <!-- define your env variables for the test env here -->
     </php>
 
     <testsuites>
--- a/phpunit/phpunit/4.7/phpunit.xml.dist
+++ b/phpunit/phpunit/4.7/phpunit.xml.dist
@@ -14,7 +14,6 @@
 
         <!-- override or set env variables for the test env here -->
         <env name="APP_ENV" value="test" />
-        <env name="APP_DEBUG" value="1" />
     </php>
 
     <testsuites>
--- a/phpunit/phpunit/4.7/tests/bootstrap.php
+++ b/phpunit/phpunit/4.7/tests/bootstrap.php
@@ -9,7 +9,7 @@ use App\Kernel;
  * Those variables will override any defined in .env.
  */
 
-Kernel::bootstrapEnvironment();
+Kernel::bootstrapEnvironment($_ENV['APP_ENV'] ?? null);
 
 $debug = $_SERVER['APP_DEBUG'] ?? true;
 
--- a/phpunit/phpunit/4.7/tests/bootstrap.php
+++ b/phpunit/phpunit/4.7/tests/bootstrap.php
@@ -1,0 +1,18 @@
+<?php
+
+require __DIR__.'/../vendor/autoload.php';
+
+use App\Kernel;
+
+/*
+ * Environment variables can also be specified in phpunit.xml.dist.
+ * Those variables will override any defined in .env.
+ */
+
+Kernel::bootstrapEnvironment();
+
+$debug = $_SERVER['APP_DEBUG'] ?? true;
+
+if ($debug) {
+    umask(0000);
+}
--- a/symfony/console/3.3/bin/console
+++ b/symfony/console/3.3/bin/console
@@ -5,7 +5,6 @@ use App\Kernel;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Debug\Debug;
-use Symfony\Component\Dotenv\Dotenv;
 
 set_time_limit(0);
 
--- a/symfony/console/3.3/bin/console
+++ b/symfony/console/3.3/bin/console
@@ -15,7 +15,7 @@ if (!class_exists(Application::class)) {
 }
 
 $input = new ArgvInput();
-$env = $input->getParameterOption(['--env', '-e'], $_SERVER['APP_ENV'] ?? 'dev', true);
+$env = $input->getParameterOption(['--env', '-e'], null, true);
 
 Kernel::bootstrapEnvironment($env);
 
--- a/symfony/console/3.3/bin/console
+++ b/symfony/console/3.3/bin/console
@@ -15,15 +15,11 @@ if (!class_exists(Application::class)) {
     throw new \RuntimeException('You need to add "symfony/framework-bundle" as a Composer dependency.');
 }
 
-if (!isset($_SERVER['APP_ENV'])) {
-    if (!class_exists(Dotenv::class)) {
-        throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
-    }
-    (new Dotenv())->load(__DIR__.'/../.env');
-}
-
 $input = new ArgvInput();
 $env = $input->getParameterOption(['--env', '-e'], $_SERVER['APP_ENV'] ?? 'dev', true);
+
+Kernel::bootstrapEnvironment($env);
+
 $debug = (bool) ($_SERVER['APP_DEBUG'] ?? ('prod' !== $env)) && !$input->hasParameterOption('--no-debug', true);
 
 if ($debug) {
--- a/symfony/framework-bundle/3.3/public/index.php
+++ b/symfony/framework-bundle/3.3/public/index.php
@@ -2,18 +2,11 @@
 
 use App\Kernel;
 use Symfony\Component\Debug\Debug;
-use Symfony\Component\Dotenv\Dotenv;
 use Symfony\Component\HttpFoundation\Request;
 
 require __DIR__.'/../vendor/autoload.php';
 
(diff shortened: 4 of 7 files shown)


phpunit/phpunit/4.7/tests/bootstrap.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+require __DIR__.'/../vendor/autoload.php';
+
+use App\Kernel;
+
+/*
+ * Environment variables can also be specified in phpunit.xml.dist.
```
joshlopes: This was a copy paste, but needs double check / test as i'm not sure if this is completely true.

symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -35,6 +36,27 @@ public function registerBundles()
         }
     }
 
+    public static function bootstrapEnvironment(string $env = null)
+    {
+        $env = $env ?? $_SERVER['APP_ENV'] ?? null;
```
joshlopes: If we decide to include `$env ?? $_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? null` we can remove the tenerary on phpunit bootsrap.

symfony/phpunit-bridge/3.3/phpunit.xml.dist
```diff
@@ -5,16 +5,16 @@
          xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.1/phpunit.xsd"
          backupGlobals="false"
          colors="true"
-         bootstrap="vendor/autoload.php"
+         bootstrap="tests/bootstrap.php"
```
joshlopes: should we update phpunit bridge configuration too ? or should this be done in phpunit bridge script instead?

ogizanagi:
Thanks for taking care of this @joshlopes!
The https://github.com/symfony/recipes/blob/master/symfony/console/3.3/bin/console file misses an update too.
At first I thought about making Flex itself providing the `DotenvLoader` class but I guess it makes sense having it in the fwb recipe.

symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -35,6 +36,27 @@ public function registerBundles()
...
+        }
+
+        // Dotenv is not installed or generic env file is not present
+        // Assume envs are already defined
+        if (!class_exists(Dotenv::class) || !file_exists($projectDir.'/.env')) {
+            return;
+        }
+
+        $envFiles = ["$projectDir/.env"];
+        if (file_exists($file = "$projectDir/.env.$env")) {
```
covex-nn: `$env` can be null here, under windows `file_exists('.env.')` will return `true`. So, if `null === $env`, then `$envFiles` will contain two elements: with `.env` and with `.env.`

symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -35,6 +36,27 @@ public function registerBundles()
         }
     }
 
+    public static function bootstrapEnvironment(string $env = null)
```
covex-nn: I think `$env` should be null here.
You already pass not null value from `bin/console`. You can pass a same value from `public/index.php` (you call `bootstrapEnvironment` and create `$env` var on the next line, just swap that lines and pass `$env` to `bootstrapEnvironment`!). Also you can pass `$_SERVER['APP_ENV'] ?? 'test'` from `tests/bootstrap.php`.

symfony/console/3.3/bin/console
```diff
@@ -15,15 +14,11 @@ if (!class_exists(Application::class)) {
...
 
-if (!isset($_SERVER['APP_ENV'])) {
-    if (!class_exists(Dotenv::class)) {
-        throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
-    }
-    (new Dotenv())->load(__DIR__.'/../.env');
-}
-
 $input = new ArgvInput();
 $env = $input->getParameterOption(['--env', '-e'], $_SERVER['APP_ENV'] ?? 'dev', true);
```
stof: This means that the environment from `.env` will not be considered anymore as fallback

symfony/console/3.3/bin/console
```diff
@@ -15,15 +14,11 @@ if (!class_exists(Application::class)) {
...
+
+Kernel::bootstrapEnvironment($env);
```
stof: I see an issue here: as soon as you use `--env`, the .env file won't be loaded anymore, meaning the flag won't be usable anymore locally.

stof: even when you don't use it btw, due to the default to `dev`

symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -35,6 +36,27 @@ public function registerBundles()
...
+    {
+        $env = $env ?? $_SERVER['APP_ENV'] ?? null;
+        $projectDir = __DIR__.'/..';
```
fabpot: should be `$this->getProjectDir()` instead.

joshlopes: @fabpot the function is static. Creating the AppKernel happens at later point after the env has been bootstrapped.

symfony/console/3.3/bin/console
joshlopes: @stof not sure what you mean, the env will always be loaded if there is a .env file and DotEnv is registered. It will also load both .env and `.env.{ENV}`. The default Server APP_ENV that is for, say, production. Whereas the default would then be `prod` for the cases without the option defined and then bootstrap would not happen.

symfony/console/3.3/bin/console
joshlopes: I think i got you now, i think we should default this to null if `$_SERVER['APP_ENV']` does not exist. That way the .env file will still be loaded and that APP_ENV will be used.

symfony/framework-bundle/3.3/src/Kernel.php
weaverryan: I think we should do this. But, read `$_ENV['APP_ENV']` first - it should take priority. This is consistent with how env vars are read in the container: 

https://github.com/symfony/symfony/blob/4f4c1725f0fcc2bb858ef0e2b5b892c5dac253a8/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php#L71-L74

phpunit/phpunit/4.7/phpunit.xml.dist
```diff
@@ -5,16 +5,16 @@
...
+         bootstrap="tests/bootstrap.php"
 >
     <php>
         <ini name="error_reporting" value="-1" />
         <env name="KERNEL_CLASS" value="App\Kernel" />
+        <env name="SHELL_VERBOSITY" value="-1" />
+
+        <!-- override or set env variables for the test env here -->
         <env name="APP_ENV" value="test" />
         <env name="APP_DEBUG" value="1" />
```
weaverryan: I don't believe this even works. This will be a string "1". I think we should remove it.

symfony/framework-bundle/3.3/public/index.php
```diff
@@ -2,18 +2,11 @@
...
 require __DIR__.'/../vendor/autoload.php';
 
-// The check is to ensure we don't use .env in production
-if (!isset($_SERVER['APP_ENV'])) {
-    if (!class_exists(Dotenv::class)) {
-        throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
-    }
-    (new Dotenv())->load(__DIR__.'/../.env');
-}
+Kernel::bootstrapEnvironment();
```
weaverryan: I'm wondering if we should SET `$_SERVER['APP_ENV']` inside `Kernel::bootstrapEnvironment();`. Basically, move this logic there. We would:

1) Load `.env` when needed
2) If `APP_ENV` is still not set, set `$_SERVER['APP_ENV'] = 'dev'`;
3) Except, if `$env` is passed to `bootstrapEnvironment()` (like for console), set that `$env` value into `$_SERVER['APP_ENV']`.

We could do the same for `$_SERVER['APP_DEBUG']`. The idea would be to simplify / de-duplicate all of the `APP_ENV` and `APP_DEBUG` logic. We could even move the `umask` & Debug code into `bootstrapEnvironment()`

symfony/framework-bundle/3.3/src/Kernel.php
covex-nn: @weaverryan environment can be not set if `'test' === $_SERVER['APP_ENV']`. And if so, then `.env` file(s) should be loaded (or `.env.test`?). And if so, then `$env` should be passed from `tests/bootstrap.php` too

weaverryan:
Ping @joshlopes!

symfony/framework-bundle/3.3/src/Kernel.php
joshlopes: @weaverryan according to with what you just reviewed later:

> So, for me, this $env argument is more of an optional, "override" - for a case where you want to use this $env even IF $_SERVER['APP_ENV'] is already set. So, it should be optional and should not be passed, except from the console.

This is the whole purpose of $env, so i think it should be read first, and fallback to the APP_ENV if we are not overriding it.

symfony/console/3.3/bin/console
joshlopes: @stof i've changed so it loads `.env` and after that load whatever `.env.$serverAppEnv` was set there

symfony/framework-bundle/3.3/src/Kernel.php
joshlopes: if $_SERVER['APP_ENV'] was defined, we assume we already have the enviromnents loaded.

joshlopes:
@weaverryan @dunglas should i just close this PR in favor of your PR's? I don't oppose to make it consistent with how react app is doing now for the env.

weaverryan:
+1 Let's close @joshlopes - but I would love your input on the other PR's that are being made for this :).

PR #409, opened 2018-05-23
Alias "phpunit" to test-pack, remove config when installing v4.1 of phpunit-bridge

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

I'd expect `composer req phpunit` to produce the same as `composer req test`.

This also implements https://github.com/symfony/symfony/issues/27037#issuecomment-387280742
There is no need to provide a default config when using symfony 4.1.

Here is a diff between recipes v3.3 and v4.1 for the phpunit-bridge:
```diff
diff -r symfony/phpunit-bridge/3.3/bin/phpunit symfony/phpunit-bridge/4.1/bin/phpunit
9c9
<     putenv('SYMFONY_PHPUNIT_REMOVE=symfony/yaml');
---
>     putenv('SYMFONY_PHPUNIT_REMOVE=');
Only in symfony/phpunit-bridge/3.3: config/services_test.yaml
diff -r symfony/phpunit-bridge/3.3/manifest.json symfony/phpunit-bridge/4.1/manifest.json
4d3
<         "config/": "%CONFIG_DIR%/",
```

### DIFF
--- a/symfony/phpunit-bridge/4.1/bin/phpunit
+++ b/symfony/phpunit-bridge/4.1/bin/phpunit
@@ -1,0 +1,18 @@
+#!/usr/bin/env php
+<?php
+
+if (!file_exists(dirname(__DIR__).'/vendor/symfony/phpunit-bridge/bin/simple-phpunit')) {
+    echo "Unable to find the `simple-phpunit` script in `vendor/symfony/phpunit-bridge/bin/`.\n";
+    exit(1);
+}
+if (false === getenv('SYMFONY_PHPUNIT_REMOVE')) {
+    putenv('SYMFONY_PHPUNIT_REMOVE=');
+}
+if (false === getenv('SYMFONY_PHPUNIT_VERSION')) {
+    putenv('SYMFONY_PHPUNIT_VERSION=6.5');
+}
+if (false === getenv('SYMFONY_PHPUNIT_DIR')) {
+    putenv('SYMFONY_PHPUNIT_DIR='.__DIR__.'/.phpunit');
+}
+
+require dirname(__DIR__).'/vendor/symfony/phpunit-bridge/bin/simple-phpunit';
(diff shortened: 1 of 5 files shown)


symfony/phpunit-bridge/3.3/manifest.json
```diff
@@ -8,6 +8,5 @@
     "gitignore": [
         ".phpunit",
         "/phpunit.xml"
-    ],
-    "aliases": ["phpunit", "simple-phpunit"]
```
stof: I would still keep the `simple-phpunit` alias

symfony/phpunit-bridge/4.1/bin/phpunit
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env php
+<?php
+
+if (!file_exists(dirname(__DIR__).'/vendor/symfony/phpunit-bridge/bin/simple-phpunit')) {
+    echo "Unable to find the `simple-phpunit` script in `vendor/symfony/phpunit-bridge/bin/`.\n";
+    exit(1);
+}
+if (false === getenv('SYMFONY_PHPUNIT_REMOVE')) {
+    putenv('SYMFONY_PHPUNIT_REMOVE=');
```
stof: is this actually needed ?

symfony/phpunit-bridge/3.3/manifest.json
nicolas-grekas: fixed

symfony/phpunit-bridge/4.1/bin/phpunit
nicolas-grekas: yes: this sets the env vars to the empty string, so that the default behavior is bypassed (when getenv returns false, yaml+prophecy are removed otherwise)

nicolas-grekas: (and symfony/yaml is no longer a phpunit dep in recent version btw if one is wondering why we don't need this anymore)

stof: there was a suggestion in symfony to change the default list of removed package to not remove prophecy (matching what this script would do by default). It may be worth reconsidering it for 4.2 if we decide that the recipe should not remove anything.

PR #410, opened 2018-05-23
[webpack-encore] use versioned assets

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Let's enable versioned assets by default?
Related to https://github.com/symfony/webpack-encore-pack/pull/1

### DIFF
--- a/symfony/webpack-encore-pack/1.0/webpack.config.js
+++ b/symfony/webpack-encore-pack/1.0/webpack.config.js
@@ -5,7 +5,7 @@ Encore
     .setOutputPath('public/build/')
     // the public path used by the web server to access the previous directory
     .setPublicPath('/build')
-    // the path inside the Symfony app, as given to the asset() function
+    // the public path you will use in Symfony's asset() function - e.g. asset('build/some_file.js')
     .setManifestKeyPrefix('build/')
 
     .cleanupOutputBeforeBuild()
--- a/symfony/webpack-encore-pack/1.0/webpack.config.js
+++ b/symfony/webpack-encore-pack/1.0/webpack.config.js
@@ -5,20 +5,27 @@ Encore
     .setOutputPath('public/build/')
     // the public path used by the web server to access the previous directory
     .setPublicPath('/build')
+    // the path inside the Symfony app, as given to the asset() function
+    .setManifestKeyPrefix('build/')
+
     .cleanupOutputBeforeBuild()
     .enableSourceMaps(!Encore.isProduction())
-    // uncomment to create hashed filenames (e.g. app.abc123.css)
-    // .enableVersioning(Encore.isProduction())
+
+    // the following line enables hashed filenames (e.g. app.abc123.css)
+    .enableVersioning(Encore.isProduction())
 
     // uncomment to define the assets of the project
-    // .addEntry('js/app', './assets/js/app.js')
-    // .addStyleEntry('css/app', './assets/css/app.scss')
+    //.addEntry('js/app', './assets/js/app.js')
+    //.addStyleEntry('css/app', './assets/css/app.scss')
+
+    // uncomment if you use TypeScript
+    //.enableTypeScriptLoader()
 
     // uncomment if you use Sass/SCSS files
-    // .enableSassLoader()
+    //.enableSassLoader()
 
     // uncomment for legacy applications that require $/jQuery as a global variable
-    // .autoProvidejQuery()
+    //.autoProvidejQuery()
 ;
 
 module.exports = Encore.getWebpackConfig();
--- a/symfony/webpack-encore-pack/1.0/config/packages/assets.yaml
+++ b/symfony/webpack-encore-pack/1.0/config/packages/assets.yaml
@@ -1,0 +1,3 @@
+framework:
+    assets:
+        json_manifest_path: '%kernel.project_dir%/public/build/manifest.json'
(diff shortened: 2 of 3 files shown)


symfony/webpack-encore-pack/1.0/webpack.config.js
```diff
@@ -4,21 +4,23 @@ Encore
     // the project directory where compiled assets will be stored
     .setOutputPath('public/build/')
     // the public path used by the web server to access the previous directory
-    .setPublicPath('/build')
+    .setPublicPath('build/')
```
nicolas-grekas: This triggers:
> WARNING   The value passed to setPublicPath() should *usually* start with "/" or be a full URL (http://...). If you're not sure, then you should probably change your public path and make this message disappear.

but this is the only way to make assets path-independent. Using the `asset()` function adds the prefix in Symfony.

symfony/webpack-encore-pack/1.0/webpack.config.js
```diff
@@ -4,21 +4,23 @@ Encore
...
     .cleanupOutputBeforeBuild()
     .enableSourceMaps(!Encore.isProduction())
-    // uncomment to create hashed filenames (e.g. app.abc123.css)
```
Nyholm: Please leave a modified comment

symfony/webpack-encore-pack/1.0/webpack.config.js
weaverryan: I think `/build` IS correct. But, even with `/build`, the "keys" in `manifest.json` should be relative - it will generate something like this:

```json
{
  "build/0.b76febcc.js": "/build/0.b76febcc.js",
  "build/admin.css": "/build/admin.fd4fe345.css",
  "build/admin.js": "/build/admin.eb9c5c59.js",
}
```

The keys will never have the opening `/`, which is why you can say `asset('build/admin.js')` and it matches.

weaverryan:
> Im trying to think of this makes deployments more tricky. Could you still (easily) upload assets or build assets on server etc?

I don't think it makes deployment more difficult. It DOES force you to use the `asset()` function in Twig and make sure you deploy the `manifest.json` file, but that should be it.

symfony/webpack-encore-pack/1.0/webpack.config.js
nicolas-grekas: When deploying an app in a subdirectory (like I just did), it breaks...

stof: but if you make it relative, things will break for any path not going through the Asset component, as they would be relative to the current page (which may be nested deeper than your root). and there might be a bunch of cases:

- JS imports handled by webpack when using code splitting (`import()` or the older `require.ensure()`)
- image paths referenced from JS (requiring an image in webpack works, and returns the path for this image in the build, including any versioning)
- image (and font) paths referenced from CSS may be an issue as well, I'm not sure. Webpack *could* generate relative paths there (as such paths are relative to the CSS location, not to the document), but it could as well be using absolute paths to keep things simpler for them

We **cannot** ensure that all assets are going through the Symfony component when dealing with more complex frontend code referencing assets from the frontend too.

nicolas-grekas: Do we need some integration then? It feels odd to have to hardcode the subpath in a webpack config when the Symfony app is just fine wherever we put it.

stof: Which kind of integration ? the Asset component guesses the subfolder being used based on the incoming request. The webpack builds is finished since ages at this point.

nicolas-grekas: Any integration that would make it seamless to put a Symfony app in a subdirectory.

stof: next solution: give up on any webpack feature referencing an asset from another one... Not really an option once you embrace webpack

nicolas-grekas: understood, updated

symfony/webpack-encore-pack/1.0/webpack.config.js
```diff
@@ -5,20 +5,27 @@ Encore
     .setOutputPath('public/build/')
     // the public path used by the web server to access the previous directory
     .setPublicPath('/build')
+    // the path inside the Symfony app, as given to the asset() function
+    .setManifestKeyPrefix('build/')
```
weaverryan: Unless you're deploying to a sub-directory or a CDN, this is not something you need. And if you DO set your publicPath to an absolute URL, we actually tell you that you need this exact method. Actually, if you set your publicPath to a subdirectory (e.g. `/subdirectory/build`) Encore also tells you in that case that you need to add this.

So, in theory, the code walks you through when you need this. So I don't want to include it for everyone else. But obviously, you had a confusing experience - so what could we have done to improve that?

weaverryan: Not that everyone goes looking for docs, but here are the docs about a sub-directory: https://symfony.com/doc/current/frontend/encore/faq.html#my-app-lives-under-a-subdirectory

nicolas-grekas: The experience I had was:
1. Refresh page in subdirectory: asset broken. => Missing public path prefix
2. Rebuild assets => error hinting about setManifestKeyPrefix

So yes, it works. But each step was a small wtf with time spent having to concentrate to understand what was happening (or even just displayed, rtfm ;) ).

Does this added line hurt anything? It would remove one oops if not :)

teohhanhui:
Why not have a `symfony/asset-pack` instead? I think it'll make more sense that the "asset pack" uses Webpack Encore.

weaverryan:
I just approved this. Need one more +1 please :)

PR #412, opened 2018-05-28
API Platform: add a post install message

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

Fixes api-platform/api-platform#657.

### DIFF
--- a/api-platform/core/2.1/post-install.txt
+++ b/api-platform/core/2.1/post-install.txt
@@ -1,0 +1,12 @@
+<bg=blue;fg=white>              </>
+<bg=blue;fg=white> What's next? </>
+<bg=blue;fg=white>              </>
+
+  * Your API is almost ready:
+    1. Create your first API resource in <info>src/Entity</info>;
+    2. Go to <info>/api</info> to browse your API
+
+  * To enable the GraphQL support, run <comment>composer require webonyx/graphql-php</>,
+    then browse <info>/api/graphql</info>.
+
+  * <fg=blue>Read</> the documentation at <comment>https://api-platform.com/docs</>


dunglas:
The error is a false positive.

api-platform/core/2.1/post-install.txt
```diff
@@ -0,0 +1,12 @@
+<bg=blue;fg=white>              </>
+<bg=blue;fg=white> What's next? </>
+<bg=blue;fg=white>              </>
+
+  * Your API is almost ready:
+    1. Create your first API resource in <info>src/Entity</info>;
```
dunglas: What do you thing about adding an example using MakerBundle here? (ping @weaverryan)

weaverryan: Ah, it’s tough because we don’t know if it’s installed. And, maybe the user already has an entity that they want to turn into a Resource? You would know better than me what a typical use-case is :)

Also, a new user won’t really know what to do at this first step - they won’t know how to “create an API” resource. A link to some very direct docs might be better?

dunglas: So maybe:

    Create a class in <info>src/Entity</info> and mark it with the <info>ApiPlatform\Core\Annotation\ApiResource</info> annotation.
    Alternatively, if <info>symfony/maker-bundle</info> is installed, run <info>bin/console make:entity --api-resource</info> to generate an API resource class interactively.

weaverryan: It’s really long, and you’ll normally add a use statement for the annotation anyways.

I think it’s better to describe what to do “mark an entity as an ApiResource. See https://“.

dunglas:
@fabpot can I merge it manually? The error is a false positive.

PR #413, opened 2018-05-28
[messenger] Docker support

maxhelias:
| Q             | A
| ------------- | ---
| License       | MIT

Requires dunglas/symfony-docker-skeleton#5 and symfony/flex#128.

It's like the PR #113 for doctrine-bundle

### DIFF
--- a/symfony/messenger/4.1/manifest.json
+++ b/symfony/messenger/4.1/manifest.json
@@ -5,6 +5,21 @@
     "env": {
         "#1": "MESSENGER_TRANSPORT_DSN=amqp://guest:guest@localhost:5672/%2f/messages"
     },
+    "docker-compose": {
+        "services": [
+            "rabbitmq:",
+            "  image: rabbitmq:management-alpine",
+            "  environment:",
+            "    # You should definitely change the password in production",
+            "    - RABBITMQ_DEFAULT_USER=guest",
+            "    - RABBITMQ_DEFAULT_PASS=guest",
+            "  volumes:",
+            "    - rabbitmq-data:/var/lib/rabbitmq",
+            "  ports:",
+            "    - \"5672\""
+        ],
+        "volumes": ["rabbitmq-data: {}"]
+    },
     "aliases": ["messenger"],
     "conflict": {
         "symfony/framework-bundle": "<4.1"
--- a/symfony/messenger/4.3/manifest.json
+++ b/symfony/messenger/4.3/manifest.json
@@ -8,6 +8,21 @@
         "#3": "MESSENGER_TRANSPORT_DSN=doctrine://default",
         "#4": "MESSENGER_TRANSPORT_DSN=redis://localhost:6379/messages"
     },
+    "docker-compose": {
+        "services": [
+            "rabbitmq:",
+            "  image: rabbitmq:management-alpine",
+            "  environment:",
+            "    # You should definitely change the password in production",
+            "    - RABBITMQ_DEFAULT_USER=guest",
+            "    - RABBITMQ_DEFAULT_PASS=guest",
+            "  volumes:",
+            "    - rabbitmq-data:/var/lib/rabbitmq",
+            "  ports:",
+            "    - \"5672\""
+        ],
+        "volumes": ["rabbitmq-data: {}"]
+    },
     "aliases": ["messenger"],
     "conflict": {
         "symfony/framework-bundle": "<4.3"


maxhelias:
Work in progress to test a new format in recipes. Visible here : https://github.com/maxhelias/flex/tree/docker

symfony/messenger/4.1/manifest.json
```diff
@@ -3,7 +3,23 @@
...
+        "#1": "MESSENGER_TRANSPORT_DSN=amqp://symfony:password@localhost:5672/%2f/messages"
+    },
+    "docker-compose": {
+        "services": {
+            "rabbitmq": {
+                "image": "rabbitmq:3.7.5",
+                "environment": [
+                    "# You should definitely change the password in production",
+                    "- RABBITMQ_DEFAULT_USER=symfony",
+                    "- RABBITMQ_DEFAULT_PASS=password"
```
Koc: consider adding `RABBITMQ_NODENAME` also.

https://docs.docker.com/samples/library/rabbitmq/
> One of the important things to note about RabbitMQ is that it stores data based on what it calls the “Node Name”, which defaults to the hostname. What this means for usage in Docker is that we should specify -h/--hostname explicitly for each daemon so that we don’t get a random hostname and can keep track of our data

symfony/messenger/4.1/manifest.json
```diff
@@ -3,7 +3,22 @@
...
+            "rabbitmq:",
+            "  image: rabbitmq:3.7.14",
+            "  environment:",
+            "    # You should definitely change the password in production",
+            "    - RABBITMQ_DEFAULT_USER=symfony",
+            "    - RABBITMQ_DEFAULT_PASS=password",
+            "  volumes:",
+            "    - rabbitmq-data:/var/lib/rabbitmq",
+            "  ports:",
+            "    - \"5672:5672\""
```
fabpot: Setting a localhost port makes it harder to have several projects running in parallel. I would only expose the container port and let Docker choose a random port for the localhost.

dunglas: Is it possible with Docker or Docker compose? AFAIK there are no built-in ways to do that, you have to rely on external tools such as https://github.com/weaveworks/weave.

Also, it will degrade the DX in some ways. For instance tools like MySQL Workbench ou PHPStorm DB expect the service to run on the default port.

fabpot: It works really well with Docker compose. Harcoding the port means that there is no way to run several projects in parallel on a laptop, that's a DX issue as well :)

dunglas: Do you have an example of the syntax?

B-Galati: An idea, do not expose any port and suggest one to use `docker-compose.override.yml` file to customize his config like exposing a port.

Pierstoval: Random port forces the end user to change any env var related to the configuration urls in `.env.local`, especially if container is recreated with a new random port, because I'm not sure any Docker or Compose configuration could propagate the random value to any env var. I'm not sure it's the best solution for DX.

If anyone is working on multiple projects in parallel, my personal recommendation would be to `docker-compose down/stop` before leaving the previous project, and `docker-compose up -d` on the next project to work on, as a measure of safety for ports.

@B-Galati's idea is better IMO: no default port, and only an override one. This invites the recipe to have a `docker-compose.override.yaml.dist` file, and forces the installation workflow to copy/paste it to `docker-compose.override.yaml` at setup to make sure a port can be publicly used, to improve DX when setting up the project.

fabpot: The Symfony CLI is able to deal with random ports as Docker exposes the ports in its API. So, you always get the right env var for all your services and you are still able to run as many projects in parallel as you want. Best DX possible.

symfony/messenger/4.1/manifest.json
```diff
@@ -3,7 +3,20 @@
         "config/": "%CONFIG_DIR%/"
     },
     "env": {
-        "#1": "MESSENGER_TRANSPORT_DSN=amqp://guest:guest@localhost:5672/%2f/messages"
+        "#1": "MESSENGER_TRANSPORT_DSN=amqp://symfony:password@localhost:5672/%2f/messages"
```
sroze: I'd keep `guest` & `guest` because they are the RabbitMq defaults.

symfony/messenger/4.1/manifest.json
```diff
@@ -3,7 +3,20 @@
...
+    },
+    "docker-compose": {
+        "services": [
+            "rabbitmq:",
+            "  image: rabbitmq:3.7.14",
+            "  environment:",
+            "    # You should definitely change the password in production",
+            "    - RABBITMQ_DEFAULT_USER=symfony",
```
sroze: Keep `guest` & `guest`?

maxhelias:
Now it's ready 😃

fabpot:
As mentioned in symfony/flex, we should expose the RabbitMQ port. See https://github.com/symfony/flex/pull/533#discussion_r309559954

fabpot:
This one is quite opinionated as one might want to use something else. But that's a good start.

PR #415, opened 2018-06-01
Messenger: use Transport instead of Adapter in ENV name

thePanz:
| Q             | A
| ------------- | ---
| License       | MIT


weaverryan:
@sroze wanted to make sure you saw this - obviously, the change is correct, but, do any docs need update to match?

sroze:
Thanks @thePanz.

PR #416, opened 2018-06-05
[symfony/twig-bundle] remove "paths", thats already the default

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

since 3.4 at least, but 3.3 is EOLed.

### DIFF
--- a/symfony/twig-bundle/3.3/config/packages/twig.yaml
+++ b/symfony/twig-bundle/3.3/config/packages/twig.yaml
@@ -1,4 +1,3 @@
 twig:
-    paths: ['%kernel.project_dir%/templates']
     debug: '%kernel.debug%'
     strict_variables: '%kernel.debug%'


xabbuh:
We could copy the existing recipe to make this on only available for 3.4 and up even if 3.3 is not maintained anymore (but we then just wouldn't do any updates anymore for the 3.3 recipe).

nicolas-grekas:
Errors are odd right now: the path is mentioned two times...

xabbuh:
Well, we could also think about improving the error message instead (i.e. checking if the default path is also part of the other configured paths?).

nicolas-grekas:
Maybe enhance the registration logic? The error message is accurate about what happens under the hood so this should remain :)

yceruto:
Maybe Twig shouldn't add the same namespace/path twice.

yceruto:
> I kinda like this config: it’s nice to say “why do templates live here? Check out this config!”.

If we want to keep this config, what about to show the `default_path` then? That's where it really lives.

nicolas-grekas:
Related to https://github.com/symfony/symfony/pull/27728 I suppose?

yceruto:
I think this is a diffent issue, actually we're adding the root `templates` directory twice:
debug_twig_recipe

It's necessary only for 3.3, since 3.4 this path is already added by `twig.default_path` option. I bet for https://github.com/symfony/recipes/pull/416#issuecomment-396481551 + https://github.com/symfony/recipes/pull/416#issuecomment-397808669:
```twig
twig:
    default_path: '%kernel.project_dir%/templates'
```
even if it's the default value.

nicolas-grekas:
(3.3 is EOLed so we don't really care for it)

fabpot:
I agree that having this setting even if that's the default value is nice. Let's close.

fabpot:
@yceruto I know, but that's something we need to fix on the Symfony side.

fabpot:
@yceruto Probably, yes

PR #417, opened 2018-06-05
Fix bug when loading bundles classes by environment

NAYZO:
| Q             | A
| ------------- | ---
| License       | MIT

Bundle classes are not loaded correctly, for example:
```php
// config/bundles.php
<?php

return [
Symfony\Bundle\FrameworkBundle\FrameworkBundle::class => ['dev' => false, 'prod' => true],
```
The FrameworkBundle will be loaded for both **dev** and for **prod** environments even with `'dev' => false` !!
  
Same thing for:
```php
// config/bundles.php
<?php

return [
Symfony\Bundle\FrameworkBundle\FrameworkBundle::class => ['all' => false],
```

### DIFF
--- a/symfony/framework-bundle/3.3/src/Kernel.php
+++ b/symfony/framework-bundle/3.3/src/Kernel.php
@@ -29,7 +29,7 @@ class Kernel extends BaseKernel
     {
         $contents = require $this->getProjectDir().'/config/bundles.php';
         foreach ($contents as $class => $envs) {
-            if (isset($envs['all']) || !empty($envs[$this->environment])) {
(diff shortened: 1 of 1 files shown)


maxhelias:
Good point, I think it could save some performance 👍 
Did you try to profile this feature ?

NAYZO:
@maxhelias no i didn't profile the feature.

maxhelias:
@NAYZO I discussed this correction on this PR #364, you should go see our exchanges

fabpot:
Only keys are relevant. Values are here for performance reasons (checking a key is faster).

PR #422, opened 2018-06-14
base.html.twig: remove optional tags

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

Remove optional tags according to [Google Best Practices](https://google.github.io/styleguide/htmlcssguide.html#Optional_Tags) and [the HTML5 spec](https://html.spec.whatwg.org/multipage/syntax.html#syntax-tag-omission).

Having a minimalist starting file is more in the spirit of Flex.

### DIFF
--- a/symfony/twig-bundle/3.3/templates/base.html.twig
+++ b/symfony/twig-bundle/3.3/templates/base.html.twig
@@ -1,12 +1,6 @@
 <!DOCTYPE html>
-<html>
-    <head>
-        <meta charset="UTF-8">
-        <title>{% block title %}Welcome!{% endblock %}</title>
-        {% block stylesheets %}{% endblock %}
-    </head>
-    <body>
-        {% block body %}{% endblock %}
-        {% block javascripts %}{% endblock %}
-    </body>
-</html>
+<meta charset="UTF-8">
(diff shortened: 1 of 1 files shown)


weaverryan:
I'm worried it'll be confusing for many users. It doesn't *look* like an HTML page anymore - so for many users (me included) I'd be thinking "what is this?" and "where should I put my content?"

So, I'm not sure it's worth it... 🤔

dunglas:
It's our duty to educate our users to the modernized frontend stack, and the new features of HTML, JS, CSS :)
Writing HTML like that is a recommended practice for a while, implemented by all browsers and easier when you get used to it.

stof:
I'm -1 for that, as it makes usage of the layout much harder. Rules about tags being optional heavily depend about what comes after them and what comes as their first child. As soon as you have a block allowing to change this content, you can be sure it is actually optional in the final output.

And btw, this would break the web debug toolbar, which won't be able to detect this as a full page where it should get injected.

stof:
And this would also break testing with the BrowserKit component, as PHP's DomDocument API does not support these rules.

Pierstoval:
This would also break the web debug toolbar which adds the script when it sees `</body>` in a Response.

fabpot:
Let's close.

PR #424, opened 2018-06-29
Disable security on ESI fragments

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Do or don't? Good default or not?

### DIFF
--- a/symfony/security-bundle/3.3/config/packages/security.yaml
+++ b/symfony/security-bundle/3.3/config/packages/security.yaml
@@ -6,6 +6,9 @@ security:
         dev:
             pattern: ^/(_(profiler|wdt)|css|images|js)/
             security: false
+        esi:
+            pattern: ^/_fragment$
+            security: false
         main:
             anonymous: true
 


Nyholm:
I’m okey with this if the /fragment only is accessible from localhost (or
trusted proxies)

nicolas-grekas:
That's not the case AFAIK, but fragments are signed so this has the same result, AFAIK again :)

sroze:
Why is it something we have to do now but we didn't have to do before? 🤔

nicolas-grekas:
My 2cts: because before session tracking was broken so we were sending the "public" headers - but now we're stricter and we notice the issue?

Nyholm:
I consider the case when someone accessing the fragment url directly. If there is no security, then all kind of nasty things may happen. 

Could we have security:false and still have a accesss control list entry?

xabbuh:
This would now indeed make it harder to restrict access to fragments to localhost (or your reverse proxy's IP address) as access control rules now do not apply. What about defining this firewall as stateless? Should have the same effect if I don't miss anything, but you would be able to add access control rules for fragment URLs.

nicolas-grekas:
I'm looking at `ContextListener`: some things happen only on the master request, but the master is on `/_fragment` so that the firewall is actually required to correctly enable actions in fragments to access the user token - which is itself triggered in a subrequest.
This means I'm now considering that the issue is that `ContextListener` is too eager at starting the session: when the user token is not used, we'd better not use the session at all.
Closing here.

chalasr:
@EmmanuelVella could you please submit a bug report on the `symfony/symfony` repository?

PR #425, opened 2018-06-29
[lexik_jwt_authentication] Add recipe for v2.5

chalasr:
| Q             | A
| ------------- | ---
| License       | MIT

The change requiring a new version of the recipe is that the `public_key_path` and `private_key_path` have been renamed to `public_key` and `secret_key`.

Motivation for this naming change is that they now both accept a raw key as value, which allows storing the raw key itself as an env var and configure the bundle like `secret_key: %env(JWT_PRIVATE_KEY)%` or use whatever custom env var processor to resolve the raw key (I need some feedbacks before using this for the recipe config). Additionally HMAC support was added, which requires only `secret_key` to be set with a raw secret string.
Lastly, only one of the key is required by the config tree (think SSO, server needs the secret key and clients need the public one, no side require to hold both, clients must not know about the secret key).

### DIFF
--- a/lexik/jwt-authentication-bundle/2.5/manifest.json
+++ b/lexik/jwt-authentication-bundle/2.5/manifest.json
@@ -1,0 +1,17 @@
+{
+    "bundles": {
+        "Lexik\\Bundle\\JWTAuthenticationBundle\\LexikJWTAuthenticationBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%"
+    },
+    "aliases": ["jwt-auth"],
+    "env": {
+        "JWT_SECRET_KEY": "%kernel.project_dir%/%CONFIG_DIR%/jwt/private.pem",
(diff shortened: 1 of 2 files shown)


lexik/jwt-authentication-bundle/2.5/manifest.json
```diff
@@ -0,0 +1,17 @@
...
+        "config/": "%CONFIG_DIR%"
+    },
+    "aliases": ["jwt-auth"],
+    "env": {
+        "JWT_SECRET_KEY": "%kernel.project_dir%/%CONFIG_DIR%/jwt/private.pem",
+        "JWT_PUBLIC_KEY": "%kernel.project_dir%/%CONFIG_DIR%/jwt/public.pem",
+        "JWT_PASSPHRASE": "%generate(secret)%"
+    },
+    "gitignore": [
+        "/%CONFIG_DIR%/jwt/*.pem"
```
sroze: Shouldn't we recommend to put these keys in `var/jwt` actually?

PR #426, opened 2018-07-01
Add intro note to services.yaml and routes.yaml

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Fixes https://github.com/symfony/symfony-docs/issues/9285


symfony/framework-bundle/3.3/config/services.yaml
```diff
@@ -1,3 +1,6 @@
+# This file is the entry point to configure your own services.
+# Files in the packages/ subdirectory configure your dependencies.
```
fabpot: Yes and no. You can store your own configuration under `packages/` as well. There is no restrictions. People are free to organize their config as they see fit (same for the routing).

PR #429, opened 2018-07-06
Enable utf8 mode for the router on 4.2

nicolas-grekas:
Sibling to https://github.com/symfony/symfony/pull/27774

### DIFF
--- a/symfony/routing/4.2/config/routes.yaml
+++ b/symfony/routing/4.2/config/routes.yaml
@@ -1,0 +1,3 @@
+#index:
+#    path: /
+#    controller: App\Controller\DefaultController::index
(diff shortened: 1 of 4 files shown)


symfony/routing/4.2/config/routes.yaml
```diff
@@ -0,0 +1,3 @@
+#index:
+#    path: /
+#    controller: App\Controller\DefaultController::index
```
xabbuh: Shouldn't this be just one colon?

nicolas-grekas: didn't we deprecate the single colon syntax? this is the same file as for lower versions btw

xabbuh: same as in the other PR: I did confuse what we deprecate 🙈

PR #432, opened 2018-07-10
Fix NelmioCors regex

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

The current regex allows domains like `http://localhost8000` (missing `:`). This PR allows only `localhost` followed by a port.

Reported by @Mathieudewet.


nelmio/cors-bundle/1.5/manifest.json
```diff
@@ -6,7 +6,7 @@
         "config/": "%CONFIG_DIR%/"
     },
     "env": {
-        "CORS_ALLOW_ORIGIN": "^https?://localhost:?[0-9]*$"
+        "CORS_ALLOW_ORIGIN": "^https?://localhost(:[0-9]+)?$"
```
Pierstoval: I'd even make the group non-capturing by adding `?:` after the first parenthesis, WDYT? 🤔

dunglas: It was my first though, but it makes the regex harder to read, and performance doesn't matter here because this value definitely need to be changed in production.

PR #434, opened 2018-07-11
Use default_path instead of paths for Twig and Translator config

yceruto:
| Q             | A
| ------------- | ---
| License       | MIT

Fix https://github.com/symfony/recipes/pull/416#issuecomment-404230767

### DIFF
--- a/symfony/translation/3.3/config/packages/translation.yaml
+++ b/symfony/translation/3.3/config/packages/translation.yaml
@@ -1,7 +1,6 @@
 framework:
     default_locale: '%locale%'
     translator:
-        paths:
-            - '%kernel.project_dir%/translations'
+        default_path: '%kernel.project_dir%/translations'
         fallbacks:
             - '%locale%'
--- a/symfony/twig-bundle/3.3/config/packages/twig.yaml
+++ b/symfony/twig-bundle/3.3/config/packages/twig.yaml
@@ -1,6 +1,4 @@
 twig:
     default_path: '%kernel.project_dir%/templates'
-    # additional template paths
-    #paths: []
     debug: '%kernel.debug%'
     strict_variables: '%kernel.debug%'
--- a/symfony/twig-bundle/3.3/config/packages/twig.yaml
+++ b/symfony/twig-bundle/3.3/config/packages/twig.yaml
@@ -1,4 +1,4 @@
 twig:
-    paths: ['%kernel.project_dir%/templates']
+    default_path: '%kernel.project_dir%/templates'
     debug: '%kernel.debug%'
     strict_variables: '%kernel.debug%'
--- a/symfony/twig-bundle/3.3/config/packages/twig.yaml
+++ b/symfony/twig-bundle/3.3/config/packages/twig.yaml
@@ -1,4 +1,6 @@
 twig:
     default_path: '%kernel.project_dir%/templates'
+    # additional template paths
+    #paths: []
     debug: '%kernel.debug%'
     strict_variables: '%kernel.debug%'


xabbuh:
Is this really a good idea? Showing the `paths` option has the advantage that developers can easily discover where to add new paths. Now they may change the setting here thus overriding the default value coming from Symfony core which would then in turn change the behaviour of the app.

javiereguiluz:
I agree with Christian. I've changed the `paths` option of Twig in lots and lots of projects. Not sure if for others is common too ... but if it is, this change makes it harder to do that.

Pierstoval:
For new projects it might not be _that_ common, as templates should still be stored under `templates/`

fabpot:
I would really be surprised if many people were overriding the paths. I cannot even think of one good reason to do that. But I understand the point. But:

 * if one changes the default_path, they will realized really fast that it breaks everything.

 * we try to not have all possible options in config files, meaning that we don't want to replace documentation here.

yceruto:
Redefining the default path again in `twig.paths` is not a good idea either, people might think that changing it they're moving the default template dir to another place, and that's not true, they might be surprised why `templates/` dir is still working.

stof:
Registering the existing default path in `paths` has another drawbacl: it means the path gets registered twice (as the default path is still registered), which was leading to bug reports about paths appearing twice in error messages

yceruto:
@stof That means you agree with this change, right? Could I get the last approval then?

javiereguiluz:
Related to this, someone asked about this yesterday on Slack:

> What are the Twig class that are in charge to look for templates in directories?
> I'd like to customize them to add some logic (go look inside this folder, then
> another one, then take default behaviour)

They replied him:

```yaml
twig:
    paths: ['%kernel.project_dir%/templates/plymouth', '%kernel.project_dir%/templates']
```

And he said:

> thanks I thought i was going to write a TemplateGuesser

-----

Apart from this, I understand what you say about `default_path` being defined twice. But don't you think the real problem is defining `default_path` itself? Not sure if possible but:

1) We could add a super simple condition that checks if `default_path` is already in `paths` and remove it to avoid duplications.
2) We could remove `default_path` and just use `paths`. If you absolutely need a default path, make it the first (or the last) element of `paths`?

yceruto:
if changing this configuration is really frequent and useful then it has won the place to appear by default.

symfony/twig-bundle/3.3/config/packages/twig.yaml
```diff
@@ -1,4 +1,6 @@
 twig:
+    # additional templates paths
```
xabbuh: additional template paths

symfony/translation/3.3/config/packages/translation.yaml
```diff
@@ -1,6 +1,8 @@
 framework:
     default_locale: '%locale%'
     translator:
+        # additional translations paths
```
xabbuh: additional translation paths

yceruto:
https://github.com/symfony/symfony/pull/27764#issuecomment-418410792 it's another reason to make this change. Setting the default path in Twig `paths` doesn't excludes the `bundles/` directory on cache warmup and the overridden templates are cached using the wrong template name.

symfony/translation/3.3/config/packages/translation.yaml
```diff
@@ -1,7 +1,8 @@
 framework:
     default_locale: '%locale%'
     translator:
-        paths:
-            - '%kernel.project_dir%/translations'
+        # additional translation paths
+        #paths: []
+        default_path: '%kernel.project_dir%/translations'
```
stof: Are you sure `default_path` works here ? IIRC, it is a Symfony 3.4+ feature, while this recipe applies to 3.3+

yceruto: Yes, but I'm following the same approach that @nicolas-grekas in the previous PR:
> https://github.com/symfony/recipes/pull/416#issuecomment-400427622 (3.3 is EOLed so we don't really care for it)

nicolas-grekas: a recipe is for new projects 99%, and no new project will use this recipe ever (or at least we don't care IMHO)

symfony/translation/3.3/config/packages/translation.yaml
```diff
@@ -1,7 +1,8 @@
 framework:
     default_locale: '%locale%'
     translator:
-        paths:
-            - '%kernel.project_dir%/translations'
+        # additional translation paths
```
nicolas-grekas: maybe we should remove this one? I'm not sure it's that common to add more paths, so better keep the config simpler? (same below of course)

yceruto: I think so too, but there is no consensus about that (others member think otherwise) at least for Twig `paths`. See previous comments.

javiereguiluz:
I still think that we should display `paths` (for Twig, not translations) ... but this is based on my own experience. If most people think it's better to not display it, let's do that so we can merge this fast. If we receive complaints from users, we can easily change this later.

yceruto:
Removed translations `paths` only.

symfony/translation/3.3/config/packages/translation.yaml
stof: but then, it would be better to migrate this recipe to be 3.4+ and have no recipe at all for 3.3, than to have a 3.3+ recipe triggering an error on 3.3

symfony/twig-bundle/3.3/config/packages/twig.yaml
```diff
@@ -1,4 +1,6 @@
 twig:
-    paths: ['%kernel.project_dir%/templates']
+    default_path: '%kernel.project_dir%/templates'
+    # additional template paths
+    #paths: []
```
nicolas-grekas: @javiereguiluz can you hint when this is needed? Still feels like something we could clean :)
Any other opinion from anyone else?

javiereguiluz: In real apps I've used it 1) to add another directory where templates were defined (yes, I know, this is the obvious purpose!), 2) on symfony.com app I use it to define Twig namespaces, 3) in some personal apps I use it to add `public/static/images/` as a directory for templates. Why? Because that way I can use the SVG icons as regular images in SCSS and embed them in Twig templates with  `{{ source('icons/-------.svg') }}`

But again, if most people never change this option, let's remove it from the default config file. Thanks!

yceruto: Going back to the simpleness version, anyway it's well documented.

PR #435, opened 2018-07-11
Fix injection of SecurityChecker with Crawler

maxhelias:
| Q             | A
| ------------- | ---
| License       | MIT

Fix the instantiation error of [SecurityChecker](https://github.com/sensiolabs/security-checker/blob/master/SensioLabs/Security/SecurityChecker.php) in the next release.

```
Too few arguments to function SensioLabs\Security\SecurityChecker::__construct(), 0 passed in /Users/mhe/www/perso/my-api/var/cache/dev/ContainerYAbmEjX/getSecurityCheckerCommandService.php on line 13 and exactly 1 expected 
```

### DIFF
--- a/sensiolabs/security-checker/4.2/config/packages/security_checker.yaml
+++ b/sensiolabs/security-checker/4.2/config/packages/security_checker.yaml
@@ -1,0 +1,13 @@
+services:
+    SensioLabs\Security\Crawler:
+        public: false
+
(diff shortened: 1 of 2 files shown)


maxhelias:
In fact it is rather better to fix the instantiation in the repot

xabbuh:
The version would be 5.0 instead of 4.2.

PR #440, opened 2018-07-19
Proofreading webpack.config.js for clarity

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

Re-reading the default webpack.config.js file for clarity. I'm also updating the documentation at the moment, and will base those changes off of this.

### DIFF
--- a/symfony/webpack-encore-pack/1.0/webpack.config.js
+++ b/symfony/webpack-encore-pack/1.0/webpack.config.js
@@ -1,30 +1,46 @@
 var Encore = require('@symfony/webpack-encore');
 
 Encore
-    // the project directory where compiled assets will be stored
+    // directory where compiled assets will be stored
     .setOutputPath('public/build/')
-    // the public path used by the web server to access the previous directory
+    // public path used by the web server to access the output path
     .setPublicPath('/build')
-    // the public path you will use in Symfony's asset() function - e.g. asset('build/some_file.js')
-    .setManifestKeyPrefix('build/')
+    // only needed for CDN's or sub-directory deploy
+    //.setManifestKeyPrefix('build/')
 
+    /*
+     * ENTRY CONFIG
+     *
+     * Add 1 entry for each "page" of your app
+     * (including one that's included on every page - e.g. "app")
+     *
+     * Each entry will result in one JavaScript file (e.g. app.js)
+     * and one CSS file (e.g. app.css) if you JavaScript imports CSS.
+     */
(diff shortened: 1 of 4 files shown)


symfony/webpack-encore-pack/1.0/webpack.config.js
```diff
@@ -1,31 +1,35 @@
...
+    // (including one that's included on every page - e.g. "app")
+    .addEntry('app', './assets/js/app.js')
+    //.addEntry('page1', './assets/js/page1.js')
+    //.addEntry('page2', './assets/js/page2.js')
 
     .cleanupOutputBeforeBuild()
     .enableSourceMaps(!Encore.isProduction())
-
-    // the following line enables hashed filenames (e.g. app.abc123.css)
+    // enables hashed filenames (e.g. app.abc123.css)
```
Pierstoval: Maybe a hint telling that it's a good practice to use it conjointedly with `framework.assets.json_manifest_path`?

symfony/webpack-encore-pack/1.0/webpack.config.js
```diff
@@ -1,31 +1,35 @@
...
     .enableVersioning(Encore.isProduction())
 
-    // uncomment to define the assets of the project
-    //.addEntry('js/app', './assets/js/app.js')
-    //.addStyleEntry('css/app', './assets/css/app.scss')
```
Pierstoval: As you added a default JS entry, why not add one for CSS too?

weaverryan:
Thanks for the reviews! I've got a WIP on this, so that I can create a corresponding PR to the docs so that everything is consistent (and we can merge at nearly the same time). I'll ping back when that's ready :)

symfony/webpack-encore-pack/1.0/webpack.config.js
weaverryan: `addStyleEntry` is kind of an anti-pattern. So, I doc'ed it, but people shouldn't really need to use it, if they want to do things correctly.

weaverryan:
This is ready for review! The big differences are that I created an `app.js` and `app.css` file with basic contents. The idea is that we recommend every app always start with these files, and these will be included in the layout. If it's an SPA or simple app, it's all they'll need. If they want page-specific JavaScript, then they'll add more entry files.

I've updated the docs with this change in mind: https://github.com/symfony/symfony-docs/pull/10128

PR #442, opened 2018-07-23
[phpunit-bridge] default to not failing on deprecations

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Until now, I've been advocating that peoples' CI should fail whenever a deprecation is reported.
But when a vendor throws a deprecation, sometime, it's not your fault and you can't do anything about it (except contributing back to the project, which is good for OSS!)

Talking on Slack, @greg0ire convinced me we can do this change: by setting `SYMFONY_DEPRECATIONS_HELPER` to `999999` by default, we make ppls' CI green (when they don't raise more than 1M deprecations.)
And it still makes the deprecation report visible on the tests' output so that ppl still have some incentive to fix them.

Fixes https://github.com/symfony/symfony/issues/27936

### DIFF
--- a/symfony/phpunit-bridge/3.3/bin/phpunit
+++ b/symfony/phpunit-bridge/3.3/bin/phpunit
@@ -5,6 +5,10 @@ if (!file_exists(dirname(__DIR__).'/vendor/symfony/phpunit-bridge/bin/simple-php
     echo "Unable to find the `simple-phpunit` script in `vendor/symfony/phpunit-bridge/bin/`.\n";
     exit(1);
 }
+if (false === getenv('SYMFONY_DEPRECATIONS_HELPER')) {
+    // see https://symfony.com/doc/current/components/phpunit_bridge.html#making-tests-fail
+    putenv('SYMFONY_DEPRECATIONS_HELPER=999999');
+}
 if (false === getenv('SYMFONY_PHPUNIT_REMOVE')) {
     putenv('SYMFONY_PHPUNIT_REMOVE=symfony/yaml');
 }
--- a/symfony/phpunit-bridge/4.1/bin/phpunit
+++ b/symfony/phpunit-bridge/4.1/bin/phpunit
@@ -5,6 +5,10 @@ if (!file_exists(dirname(__DIR__).'/vendor/symfony/phpunit-bridge/bin/simple-php
     echo "Unable to find the `simple-phpunit` script in `vendor/symfony/phpunit-bridge/bin/`.\n";
     exit(1);
 }
+if (false === getenv('SYMFONY_DEPRECATIONS_HELPER')) {
+    // see https://symfony.com/doc/current/components/phpunit_bridge.html#making-tests-fail
+    putenv('SYMFONY_DEPRECATIONS_HELPER=999999');
+}
 if (false === getenv('SYMFONY_PHPUNIT_REMOVE')) {
     putenv('SYMFONY_PHPUNIT_REMOVE=');
 }


ChangePlaces:
But seeing deprecations reported in methods of _your_ code is scary, and prompts urgent WTF'ing. I'd advocate the average developer doesn't care for deprecations in code that isn't theirs and for that reason, this should be disabled by default, with a note in the testing docs detailing how to opt it.

nicolas-grekas:
Configuring the bridge via `phpunit.xml` is not supported, the bridge runs too early.
See https://github.com/symfony/symfony/issues/28028 btw.

Tobion:
I'd prefer if this new default is handled in https://github.com/symfony/symfony/blob/0eea07752211f464818a54efc7907e8ca4c617a6/src/Symfony/Bridge/PhpUnit/bootstrap.php#L39 so that it also behaves like this when not using the simple-phpunit wrapper.

nicolas-grekas:
But that would be a BC break. I also think having a stricter default is a better practice. And requiring people to opt-in for less strict config (via their own wrapper or a real env var) encourages ownership IMHO.

nicolas-grekas:
ping @symfony/deciders

Tobion:
As the recommended default, wouldn't it be cleaner and more self-explaining to have an explicit config value for this like `pass-and-show` instead of some random number.

nicolas-grekas:
That'd be a new feature, thus not actionable until 4.2. And personally, I prefer the number, because it hints it exists and can be changed: lowering this number up to zero is a best practice IMHO.

ChangePlaces:
Where is this best practice declared so people can read up on it?

nicolas-grekas:
@ChangePlaces I looked at [the doc](https://symfony.com/doc/current/components/phpunit_bridge.html#making-tests-fail), it's not explicit so we could improve (it's implicit because that's the default, but I suppose that's not enough.) Would you mind sending a PR to improve that §?

ChangePlaces:
I don't think I'm the best person to submit a PR as I fail to see it as a best practice myself

alcaeus:
@nicolas-grekas what's needed to get this merged?

nicolas-grekas:
One more vote from any @symfony/deciders

dunglas:
I would prefer a new config option too. This one looks like a hack, and should’nt be in the official recipe if we can provide something nicer.
Also, shouldn’t we use $_SERVER instead of getenv for consistency?

nicolas-grekas:
A new option will not be applicable until 4.2. I can wait personally, and I also prefer strict mode so I'm also fine closing. But that's not what others would prefer... Let's move on ?

nicolas-grekas:
Oh and about `$_SERVER`, the bridge only uses `getenv()`

nicolas-grekas:
About the future, see https://github.com/symfony/symfony/issues/28048

greg0ire:
Also, I think a logical follow-up would be to make a BC-breaking PR to use `s/0/999999/` on this line, correct?

https://github.com/symfony/symfony/blob/f96753b9ab497a42f7a76188e2de4911b50f5407/src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler.php#L40

nicolas-grekas:
Nope, there's no logical BC break.

greg0ire:
Sorry, what do you mean with "logical BC-break"? I think a user having suddenly the bridge no longer make their test suite fail where it used to fail is entitled to call that a BC-break, aren't they? Would you do that PR on the stable branch? Or did I misunderstand your answer?

nicolas-grekas:
I mean we cannot change that line 40 you linked: that'd be a BC break and we just don't allow us to do that.
This PR is on a recipe so that only new project will get the new default config. No BC break either, all is well :)

greg0ire:
Thanks for the clarification, I'm trying to address some concerned voiced on https://github.com/doctrine/doctrine-website/issues/200

PR #447, opened 2018-08-08
Update README.rst

yceruto:
| Q             | A
| ------------- | ---
| License       | MIT

This fix a typo according to https://github.com/symfony/flex/blob/fabf04baa532b8068558d1e54a9358ef26814449/src/Flex.php#L404, though it doesn't work currently, see https://github.com/symfony/demo/issues/842

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -222,10 +222,12 @@ script (``php-script`` for PHP scripts, ``script`` for any shell script and
 .. code-block:: json
 
     {
-        "auto-scripts": {
-            "vendor/bin/security-checker security:check": "php-script",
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -222,7 +222,7 @@ script (``php-script`` for PHP scripts, ``script`` for any shell script and
 .. code-block:: json
 
     {
-        "composer-scripts": {
+        "auto-scripts": {
```
fabpot: We should probably add the wrapping `scripts` entry to make it very clear. WDTY?

yceruto: Sorry for the delay, done.

PR #451, opened 2018-08-15
Allow installation of latest encore version

BackEndTea:
| Q             | A
| ------------- | ---
| License       | MIT

Under yarn/npm versioning rules, the new `0.20.*` versions  weren't installable by default,  see also: 
https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004

Its pretty much the same as composer versioning, as in it wont bump minor versions in pre `1.0.0` versioning.


BackEndTea:
Thank you @fabpot & @maxhelias

PR #452, opened 2018-08-22
[symfony/framework-bundle] Secure session defaults with "cookie_secure: auto" and "cookie_samesite: lax"

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Enabling https://github.com/symfony/symfony/pull/28244 and https://github.com/symfony/symfony/pull/28168 by default.


fabpot:
That only works on 4.2+, so we probably need a new directory for that version.

nicolas-grekas:
> That only works on 4.2+, so we probably need a new directory for that version.

Indeed, updated.
Now also enabled "cookie_samesite: lax" by default.

PR #453, opened 2018-08-27
 [FrameworkBundle] New `parameters.yaml` file + Remove `parameters` key from `services.yaml`

alex2005git:
| Q             | A
| ------------- | ---
| License       | MIT

What this PR hopes to accomplish, is remove the `parameters` key from `services.yaml`, it's a bit counter intuitive for the parameters to be in the `services.yaml` file.  
  
Also the parameters could get quite many, also the services, so the separation would make the files more clear as to what they should do.  
  
Another advantage this has, because the `parameters.yaml` file gets loaded before other configs, an environment could overwrite a parameter (for whatever reason) by providing a `parameters.yaml` file in the `packages/{env}` folder. This is intuitive, you would think that by placing this file in an environment folder it would overwrite the parameters that you set in it.

### DIFF
--- a/symfony/framework-bundle/3.3/config/parameters.yaml
+++ b/symfony/framework-bundle/3.3/config/parameters.yaml
@@ -1,0 +1,3 @@
+# Put parameters here that don't need to change on each machine where the app is deployed
+# https://symfony.com/doc/current/best_practices/configuration.html#application-related-configuration
+parameters:
--- a/symfony/framework-bundle/3.3/config/services.yaml
+++ b/symfony/framework-bundle/3.3/config/services.yaml
@@ -1,10 +1,6 @@
 # This file is the entry point to configure your own services.
 # Files in the packages/ subdirectory configure your dependencies.
 
-# Put parameters here that don't need to change on each machine where the app is deployed
-# https://symfony.com/doc/current/best_practices/configuration.html#application-related-configuration
-parameters:
-
 services:
     # default configuration for services in *this* file
     _defaults:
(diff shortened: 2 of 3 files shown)


BoShurik:
It that case where you suggest to place `parameters.yaml` generated by [incenteev/composer-parameter-handler](https://github.com/Incenteev/ParameterHandler)?

BoShurik:
Also, there is [container configurator](https://github.com/symfony/recipes#container-configurator) that depends on `parameters` key in `services.yaml`

alex2005git:
@BoShurik Does the file have anything to do with symfony parameters?
The file name could be anything you want from what I see in the documentation `"file": "app/config/parameters.yml",` but I don't know how someone uses that file after it is created, it gets loaded somehow?

alex2005git:
The container configuration is an easy enough PR, from just a quick look I had.  
I could go ahead an make it if this might get integrated...

BoShurik:
I use it like this in `services.yaml`
```yaml
imports:
    - { resource: 'parameters.yaml' }
parameters:
    locale: 'en'
#...
```

For me the problem is that this file has been used as a `.env` before symfony 4, so it might be misleading to use another name in case using `incenteev/composer-parameter-handler`

But it's just my opinion, let's wait what others think about it

BoShurik:
I'm still using it because it's harder to manage env variables without docker. You need to set them in two places (web-server and CLI config) instead of one (`parameters.yaml`)

fabpot:
I did include parameters in the main services.yaml file on purpose. The use of parameters is way less important as of Symfony 4, and as such, I don't want to make it front and center. Still here if you need it, but just not first class citizen anymore. You can of course do this change for your project, but it cannot be the default.

PR #457, opened 2018-09-07
revert #447

xabbuh:
| Q             | A
| ------------- | ---
| License       | MIT

As said in https://github.com/symfony/recipes/pull/447#issuecomment-418300427 I think we need to revert this commit. The readme does not describe the format of the resulting `composer.json` section, but describes how you need to define the configurator in the `manifest.json` file of a recipe.


xabbuh:
ping @yceruto

PR #458, opened 2018-09-09
Adding recipe for symfony/webpack-encore-bundle

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

Hi guys!

This is for the new bundle - https://github.com/symfony/webpack-encore-bundle.

This would replace the webpack-encore-pack recipe. Well, we won't delete that, but the idea would be that we would now install this bundle, not the pack.

Cheers!


symfony/webpack-encore-bundle/1.0/config/packages/webpack_encore.yaml
```diff
@@ -0,0 +1,4 @@
+webpack_encore:
+    # The path where Encore is building the assets.
+    # This should match Encore.setOutputPath() in webpack.config.js.
+    output_path: '%kernel.public_dir%/build'
```
Koc: What about using two separate build directories per dev/prod env? It useful when you should testing in both modes in same time.

Pierstoval: I don't think that it's really common to run dev & prod at the same time in the same environment...

weaverryan:
This is ready for review - the bundle 1.0 has been released.

The biggest change is moving the recipe from `webpack-encore-pack` to `webpack-encore-bundle`. The pack was originally needed because there was no PHP package we could attach the recipe to. But, the bundle has the same dependencies as the pack (`symfony/assets`) and so we can fully use it.

Thanks!

weaverryan:
This is ready to go. I just tested that if you remove `symfony/webpack-encore-pack`, it will not remove all your existing files, because it (apparently) uses the most up-to-date recipe version, which contains no files to remove.

PR #459, opened 2018-09-13
Slightly tweak the comment

bocharsky-bw:
| Q             | A
| ------------- | ---
| License       | MIT

The current condition does not guarantee we do not load `.env` in production. We should tweak whether the comment or the condition to `if (!isset($_SERVER['APP_ENV']) || 'prod' === $_SERVER['APP_ENV'])`. Or maybe remove this comment at all? It's pretty obvious.


Nyholm:
Thank you Victor

PR #461, opened 2018-09-13
Proposal to make it possible to load website in different environments with Symfony Web Server

bocharsky-bw:
| Q             | A
| ------------- | ---
| License       | MIT

The problem is that with the current implementation it's impossible to load websites in different environments except `dev` one using Symfony Web Server. That's because Symfony Web Server is registered for `dev` env only by default, i.e: `Symfony\Bundle\WebServerBundle\WebServerBundle::class => ['dev' => true],` in `config/bundles.php` and revealing it for other envs is not good. So, if I want to test my website in `prod`/`test`/etc. envs with Symfony Web Server - I get an error about:

> There are no commands defined in the "server" namespace.

So if I start the server in `dev` mode - the website is also in `dev` mode and I can't change it with env vars, and even modifying `.env` file after the server was started does not help. So the only solution I see is to create a new front controller like `public/index_prod.php`, `public/index_test.php`, etc. where hardcode environment you need. But this takes us back to multiple front controllers.

However, with these changes, it is possible to run Symfony Web Server in `dev` mode but the website in any other environment you want with the next command:

```bash
APP_ENV=prod bin/console s:r
```

You still can load the website in the `dev` mode by default:

```bash
bin/console s:r
```

I think Symfony console is a bit special, so we should not base it on env variables, instead use the old `--env` option to switch environments.

Here's a complete application with these changes that you can see in action:
https://github.com/bocharsky-bw/env_var_reproducer


Pierstoval:
We can't trust `getenv()`, it's not thread safe, so this would lead to errors in some production environments :confused:

bocharsky-bw:
Oh... I changed `$_SERVER` with `getenv()` because when you run `APP_ENV=prod bin/console s:r` command with these changes, in `bin/console` `$_SERVER` and `getenv()` have env vars, however `$_ENV` is empty. But in `public/index.php` `$_ENV` and `getenv()` have env vars, but `$_SERVER` does not. Hm, probably it's possible to pass env vars to $_SERVER with Symfony Web Server? Then we can keep using `$_SERVER` in `index.php`

Pierstoval:
The problem comes from PHP that's not populating `$_SERVER` correctly with env vars in CLI that's exactly why I proposed symfony/symfony#25693, to avoid inconsistencies of this kind...

bocharsky-bw:
Thanks for pointing it! So what if we all `$_SERVER` with `$_ENV` in index.php instead of `getenv()` as I'm doing in this PR? I'd be happy with these changes as well because it will solve my problem as well. I changed `$_SERVER` to `getenv()` just because I think it should be better, but now I see it's not  thread-safe, but $_ENV could work then.

Or are there any reasons why we cannot use `$_ENV` instead of `$_SERVER` in index.php? Probably I'm missing something. And if we can't - then the only solution is to fallback $_SERVER to $_ENV that I think is OK to have in index.php

Pierstoval:
We can't use `$_ENV` only either, because a `php.ini` directive (or other probems) can make it empty...

bocharsky-bw:
OK, thanks for the explanation. So we need to take into account both `$_SERVER` and `$_ENV` in index.php in order to fix the problem, and I suppose not only in index.php probably but in other places where we use only `$_SERVER`. That's why having some kind of API as suggested in https://github.com/symfony/symfony/issues/25693 would make sense I think

stof:
That change is a BC break, as it means that commands will not run in the env specified in the environment. And this can cause issues for instance for any command running in the composer scripts, which would have to run in the prod env on a prod server, because dev requirements are not installed.

I think only the webserver commands need a special case, to be able to start the webserver with a different env than the one running the command.

maxhelias:
Does the PR #466, complete your need ?

bocharsky-bw:
Looks like `LOAD_ENV=prod bin/console` might work for me 👍

maxhelias:

PR #462, opened 2018-09-14
Move more platform-dependent options to the DSN

scaytrase:
| Q             | A
| ------------- | ---
| License       | MIT

https://github.com/doctrine/dbal/blob/a2cdd8a5b6f507e7f0c509fb9b0134561db357c8/lib/Doctrine/DBAL/DriverManager.php#L215-L272

IIUC this code, doctrine/dbal already supports parsing extra options from DSN query string.

I suggest move platform-dependent options to the DSN by default, making it easier, i.e to switch on sqlite for testing just switching the DSN without leaving confusing 5.7 version and driver setting behinf

### DIFF
--- a/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
@@ -8,8 +8,6 @@ parameters:
 doctrine:
     dbal:
         # configure these for your database server
-        driver: 'pdo_mysql'
-        server_version: '5.7'
         charset: utf8mb4
         default_table_options:
             charset: utf8mb4
--- a/doctrine/doctrine-bundle/1.6/manifest.json
+++ b/doctrine/doctrine-bundle/1.6/manifest.json
@@ -9,7 +9,6 @@
     "env": {
         "#1": "Format described at http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url",
         "#2": "For an SQLite database, use: \"sqlite:///%kernel.project_dir%/var/data.db\"",
-        "#3": "Configure your db driver and server_version in config/packages/doctrine.yaml",
-        "DATABASE_URL": "mysql://db_user:db_password@127.0.0.1:3306/db_name"
+        "DATABASE_URL": "mysql://db_user:db_password@127.0.0.1:3306/db_name?server_version=5.7"
     }
 }


scaytrase:
Not sure I'm responsible for the `container` failure

BoShurik:
Without this parameters you can't run cache:warmup
https://github.com/symfony/recipes/blob/master/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml#L2-L5

scaytrase:
@BoShurik I didn't remove the default env which this comment belongs to. If config parameters are requried - could we do something like `DATABASE_URL: mysql://localhost/?server_version=5.7` as env default (or whatever the default values are), which will make this more explicit

Pierstoval:
Why not add two envs `DATABASE_DRIVER` and `DATABASE_VERSION` that would be defaulted to mysql 5.7 in the recipe file?

scaytrase:
Driver is already the part of DATABASE_URL. Version could be the part as suggested, so I really do not see the need of separate env var but considerable. I think not every driver uses version, i.e sqlite

Pierstoval:
Yep but even with an empty `DATABASE_URL`, Doctrine needs driver and version to compile the container.

scaytrase:
But when the database url is empty it is not actually empty

where is default driver 
https://github.com/doctrine/DoctrineBundle/blob/master/DependencyInjection/Configuration.php#L127
default host-port-user etc
https://github.com/doctrine/DoctrineBundle/blob/master/DependencyInjection/Configuration.php#L191-L194

We can just do this deafults being the default URL components alongside with the query and server_version

Pierstoval:
This is not only about splitting vars or database url being empty.

To compile the container, Doctrine needs the version and the driver to know which Driver and Platform service classes will be registered in the container, this is done in a compiler pass, and it doesn't need the host nor the user at this time.

If you don't set the version but you have the driver, Doctrine will make a call to the database to know the version, to know which services need to be compiled, and to make a call, you need proper credentials, which is not always the case when starting the project.

That's one reason why driver & version are set by default in the recipe.

IIRC, the team decided to set mysql as default because it's probably the most used dbms, and 5.7 because it was the latest stable when the recipe was created (we could even set 8.0 as it's the latest stable now). This probably also encourages developers to keep the same driver across environments, which is a good practice IMO.

jvasseur:
@Pierstoval, that's wrong, the Driver and Platform are not compiled into the container but are resolved at runtime in the connection factory (https://github.com/doctrine/DoctrineBundle/blob/master/ConnectionFactory.php#L52)

The "query to get the database version" error that used to append, it append during the warmup and not during container compilation because a some services that depended on the connection were loaded, forcing Doctrine to make a query to get the version to instantiate the connection with the right platform.

This means you have to either set the driver and version in the the config like we currently do our always pass an environment variable that contains at least the driver and the server version (it might work without it if you're lucky and don't depends on Doctrine for warmup).

scaytrase:
@jvasseur DriversManager utilizes the connection params there

https://github.com/doctrine/dbal/blob/a2cdd8a5b6f507e7f0c509fb9b0134561db357c8/lib/Doctrine/DBAL/DriverManager.php#L153

But it parses database URL first to split it into params
https://github.com/doctrine/dbal/blob/a2cdd8a5b6f507e7f0c509fb9b0134561db357c8/lib/Doctrine/DBAL/DriverManager.php#L144

And it could obtain both version
https://github.com/doctrine/dbal/blob/a2cdd8a5b6f507e7f0c509fb9b0134561db357c8/lib/Doctrine/DBAL/DriverManager.php#L144

And any custom parameter from query string of url
https://github.com/doctrine/dbal/blob/a2cdd8a5b6f507e7f0c509fb9b0134561db357c8/lib/Doctrine/DBAL/DriverManager.php#L265-L269

So generally passing the `DATABASE_URL` with defaults in it is the same as leaving default config, so we could use

`DATABASE_URL=mysql://root@localhost/?server_version=5.7` to achieve the same config as we have before in YAML without having any conflicting hardcodes when we use another driver in URL

javiereguiluz:
I disagree about this change. In my experience is not that common to have completely different databases in the different environments: MySQL 5.7 in local an 3.4 in production? PostgreSQL in local and MariaDB in production? 

I'm not saying that nobody does that ... but I doubt is common, so I wouldn't move them to an env var.

scaytrase:
@javiereguiluz most frequent case I've seen is SQLite for tests and *SQL for other

chalasr:
This option has been moved on purpose in https://github.com/symfony/recipes/pull/258, I think it's good as-is. Having the option defined statically sounds better to me as you must change it explicitly if you need to (you generally don't want your code to run under different server version).
As said in the PR body, doctrine just ignores the `server_version` option when sqlite is used as it is serverless. IMHO leaving the option set does not harm, better always reflect production settings than testing ones.

scaytrase:
@chalasr nice link, thank you. But I still don't get why can we move defaults to env url like mentioned in that PR

```
env(DATABASE_URL): 'mysql://?server_version=5.7'
```

instead of 

```yaml
doctrine:	doctrine:
    dbal:	    dbal:
        # configure these for your database server
        driver: 'pdo_mysql'	
        server_version: '5.7'
```

PR #464, opened 2018-09-20
[FWBundle] $_ENV fallback (needed for Panther)

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

When the builtin PHP web server is used, env vars are only available in `$_ENV`, not in `$_SERVER` (see https://bugs.php.net/bug.php?id=67808).

This is especially annoying when testing an app with Panther because, by default, it automatically boots the app using the integrated PHP web server, consequently env vars defined in `phpunit.xml` aren't inherited, and the app executed using the `dev` env instead of the `test` one.

This PR fixes this problem.

### DIFF
--- a/symfony/framework-bundle/3.3/public/index.php
+++ b/symfony/framework-bundle/3.3/public/index.php
@@ -8,15 +8,15 @@ use Symfony\Component\HttpFoundation\Request;
 require __DIR__.'/../vendor/autoload.php';
 
 // The check is to ensure we don't use .env in production
-if (!isset($_SERVER['APP_ENV'])) {
+if (!isset($_SERVER['APP_ENV']) && !isset($_ENV['APP_ENV'])) {
     if (!class_exists(Dotenv::class)) {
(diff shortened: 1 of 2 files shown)


albertboada:
Why not using `getenv()`?

dunglas:
@albertboada because it's not thread safe.

stof:
Does DotEnv checks both `$_SERVER` and `$_ENV` for existing env variables too ?

Pierstoval:
@stof seems it's a handled case:

https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Dotenv/Dotenv.php#L81

albertboada:
> @albertboada because it's not thread safe.

Thanks @dunglas, I see. 

However, we are experimenting a situation where `'APP_ENV'` is neither set in `$_SERVER` nor in `$_ENV`, and the only way of bootstrapping correctly our Symfony 4 application in our production is falling back to `getenv()`. We are deploying on Amazon Elastic Container Service. Any particular reason why this can happen?

Issue #317, opened 2017-12-16
[RFC] Allow a recipe to be merged with other versions


Pierstoval:
For issues like #316, the recipe needs a single file to be updated and the rest is quite similar.

Maybe we could think about a way to "merge" a recipe with the same recipe for another version? This could avoid having duplicate recipes with too few changes.

I'm not sure of any technical way to do it, because there might be several different cases, like merging a list of files, merging manifest rules...

WDYT?

Nyholm:
A recipe is something you use when you install the dependency. You copy some default config down from the internet and then let the user modify it. In my opinion, it would lead to many weird side effects if you changed existing configuration on `composer update`. Even a command like `composer update-recipes` would not really be preferred. 

However, if this was not a "recipe repository" but a "configuration repository", your suggestion would make more sense.

Pierstoval:
I think I haven't expressed myself correctly, I'm not talking about a composer command, but a way to avoid copy/pasting recipes when releasing a new version of a package.

It's just to merge recipes on symfony.sh.

Example:

A package defines a manifest that copies files, sets environment vars, a bundle, and scripts.
But let's imagine this package releases a new major version that optimizes some configuration options (let's say, the routing format, for the case of #316 ), and only one file in the recipe needs to be updated.

In order to avoid duplicating the recipe and changing a single file, we could simply add a new manifest for this new version, and have a rule in it that says something like _"Take this config from the previous version"_, and execute current version rules after that.

Nyholm:
Ah, okey. I understand. That is a nice idea. 

How would the implementation look like? Should a bit complicated to try to figure out how many versions back we should go looking for config files. Also, how about readability? Right now if I want version 3.1 of package `foo` I can look at the recipe repo and find that version and know exactly what I would get.

Pierstoval:
I'd like to have more opinions about the idea, but I'm wondering should we be able to import **only specific rules** or **all rules**?

I'm personally in the second option and override afterwards, therefore the implementation might look like this:

By importing manually all rules we need, and then, all we add will be merged:
```json
{
    "import_rules": {
        "3.1": [
            "copy-from-recipe"
        ],
        "4.0": [
            "bundles"
        ]
    },
    "copy-from-recipes": {
        "new_dir/": "%PROJECT_DIR%/new_dir/"
    }
}
```

Or just specify the recipe versions, everything will be retrieved in the specified order, all we add manually will be overriden:
```json
{
    "from": [
        "3.1", "3.2"
[...]

PR #465, opened 2018-09-20
PHPUnit: use a .env file

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

Having to modify both `.env` and and `phpunit.xml.dist` when adding an environment variable is tedious and error prone.
Storing env vars in different formats (XML and `.env`) also leads to weird bugs such as #317.

With this PR, in test mode `.env` and `.env.test` files are always loaded. `.env.test` allows to override env vars for this env.

Not related with this PR: I don't really understand why we need both `.env` and `.env.dist`. As `.env` file must not be used in production anyway, can't we have only a (versioned) `.env` file and get rid of the `.dist` one?
For people not using Docker/Vagrant/etc, we may allow to override the values of the `.env` file in a `.env.local` file (that will not be versioned).
Even if we want to keep the `.dist` file, cannot we always load it along with `.env`? Then we'll be able to only use `.env` to override some vars (the secrets) instead of having to copy the full content in both. wdyt?

### DIFF
--- a/symfony/phpunit-bridge/4.1/.env.test
+++ b/symfony/phpunit-bridge/4.1/.env.test
@@ -1,0 +1,2 @@
+APP_ENV=test
+APP_SECRET=s$cretf0rt3st
--- a/symfony/phpunit-bridge/4.1/tests/bootstrap.php
+++ b/symfony/phpunit-bridge/4.1/tests/bootstrap.php
@@ -1,0 +1,7 @@
+<?php
+
+require __DIR__.'/../vendor/autoload.php';
+
+use Symfony\Component\Dotenv\Dotenv;
+
+(new Dotenv())->load(__DIR__.'/../.env', __DIR__.'/../.env.test');
--- a/symfony/phpunit-bridge/4.1/manifest.json
+++ b/symfony/phpunit-bridge/4.1/manifest.json
@@ -2,7 +2,8 @@
     "copy-from-recipe": {
         "bin/": "%BIN_DIR%/",
         "phpunit.xml.dist": "phpunit.xml.dist",
-        "tests/": "tests/"
+        "tests/": "tests/",
+        ".env.test": ".env.test"
     },
     "gitignore": [
         ".phpunit",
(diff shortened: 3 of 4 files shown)


ogizanagi:
Have you seen https://github.com/symfony/recipes/pull/408 and related issues?
To me as mentioned in https://github.com/symfony/recipes/pull/366#issuecomment-364778024, another issue to consider with the current situation is we cannot easily execute a sf command in any environment.

> For people not using Docker/Vagrant/etc, we may allow to override the values of the .env file in a .env.local file (that will not be versioned).

👍 I've seen multiple dotenv components using this. E.g: 
- https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use
- https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#what-other-env-files-can-be-used

symfony/phpunit-bridge/4.1/tests/bootstrap.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+require __DIR__.'/../vendor/autoload.php';
+
+use Symfony\Component\Dotenv\Dotenv;
+
+(new Dotenv())->load(__DIR__.'/../.env', __DIR__.'/../.env.test');
```
weaverryan: If an env var is specified in both, `.env.test` wins, right?

Also, if we do this for the `test` environment, people will ask for it for other environments too. And I think they're right - it would be simpler to have a file like this that will load from `.env.{ENV}` (if it exists) and to require that from `index.php` and `console`. `console` is especially important - we don't want someone to run a `bin/console` test with `--env=test` and have different environment variables than their tests.

Check out https://github.com/symfony/recipes/pull/408/files - where we working on something similar to this. We were also putting even some extra code into this central "bootstrap" file to fully remove the duplication between `index.php`, `console` (and now) the test bootstrap.

symfony/phpunit-bridge/4.1/.env.test
```diff
@@ -0,0 +1,2 @@
+APP_ENV=test
+APP_SECRET=s$cretf0rt3st
```
weaverryan: Is there any benefit to defining these? The `APP_ENV` will always already be set. Perhaps just leave it blank, but with a comment that overrides can go here?

dunglas:
I’m with totally for “copying” the behavior of Create React App for several reasons:

- it works smoothly and is easier to use for developers than what we do know
- developers coming from the JS world are used to this behavior
- something more “personal”: api-platform/api-platform comes with both Symfony Flex and Create React App (it’s a monorepo containing the server and the client code), uniformizing will improve our DX a lot

I’ll update the PR accordingly tomorrow

dunglas:
Superseded by #466.

PR #466, opened 2018-09-21
Allow to override .env files for specific environments

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

As discussed in #465 and symfony/symfony#28533, this PR makes Symfony behaving like Create React App, Rails or Sinatra: the following files, starting from the bottom. The first value set (or those already defined in the environment) take precedence:

- `.env` - The Original®
- `.env.dev`, `.env.test`, `.env.prod`... - Environment-specific settings.
- `.env.local` - Local overrides. This file is loaded for all environments _except_ `test`.
- `.env.dev.local`, `.env.test.local`, `.env.prod.local`... - Local overrides of environment-specific settings.

The default envs are `env` and `test`.

It also fixes https://github.com/symfony/recipes/pull/366#issuecomment-364778024: when running `bin/console -e=prod`, the `.env.prod` file will be loaded (if `APP_ENV` is not already defined).

Needs:

- symfony/symfony#28533
- symfony/flex#423

Closes #465, #366, #408.

### DIFF
--- a/symfony/console/4.2/bin/console
+++ b/symfony/console/4.2/bin/console
@@ -1,0 +1,41 @@
+#!/usr/bin/env php
+<?php
+
+use App\Kernel;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Debug\Debug;
+use Symfony\Component\Dotenv\Dotenv;
+
+set_time_limit(0);
+
+require __DIR__.'/../vendor/autoload.php';
+
+if (!class_exists(Application::class)) {
+    throw new \RuntimeException('You need to add "symfony/framework-bundle" as a Composer dependency.');
+}
+
+$input = new ArgvInput();
+$env = $input->getParameterOption(['--env', '-e'], null, true);
+
+if (!isset($_SERVER['APP_ENV'])) {
+    if (!class_exists(Dotenv::class)) {
+        throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
+    }
+    (new Dotenv())->loadForEnv($env ?? 'dev', __DIR__.'/../.env');
+}
+
+$env = $env ?? $_SERVER['APP_ENV'] ?? 'dev';
+$debug = (bool) ($_SERVER['APP_DEBUG'] ?? ('prod' !== $env)) && !$input->hasParameterOption('--no-debug', true);
+
+if ($debug) {
+    umask(0000);
+
+    if (class_exists(Debug::class)) {
+        Debug::enable();
+    }
+}
+
+$kernel = new Kernel($env, $debug);
+$application = new Application($kernel);
+$application->run($input);
--- a/symfony/console/4.2/bin/console
+++ b/symfony/console/4.2/bin/console
@@ -3,22 +3,14 @@
 
 use App\Kernel;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
-use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Debug\Debug;
-use Symfony\Component\Dotenv\Dotenv;
 
 set_time_limit(0);
 
-require __DIR__.'/../vendor/autoload.php';
+require __DIR__.'/../src/bootstrap.php';
 
 if (!class_exists(Application::class)) {
     throw new \RuntimeException('You need to add "symfony/framework-bundle" as a Composer dependency.');
-}
-
-if (class_exists(Dotenv::class)) {
-    (new Dotenv())->loadForEnv($_SERVER['APP_ENV'] ?? 'dev', __DIR__.'/../.env');
-} elseif (!isset($_SERVER['APP_ENV']) && !isset($_ENV['APP_ENV'])) {
-    throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
 }
 
 $env = $_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'dev';
--- a/symfony/console/4.2/bin/console
+++ b/symfony/console/4.2/bin/console
@@ -7,14 +7,11 @@ use Symfony\Component\Debug\Debug;
 
 set_time_limit(0);
 
-require __DIR__.'/../src/bootstrap.php';
+[$env, $debug] = require __DIR__.'/../src/bootstrap.php';
 
 if (!class_exists(Application::class)) {
     throw new \RuntimeException('You need to add "symfony/framework-bundle" as a Composer dependency.');
 }
-
-$env = $_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'dev';
-$debug = (bool) ($_SERVER['APP_DEBUG'] ?? $_ENV['APP_DEBUG'] ?? ('prod' !== $env)) && !$input->hasParameterOption('--no-debug', true);
 
 if ($debug) {
     umask(0000);
--- a/symfony/console/4.2/bin/console
+++ b/symfony/console/4.2/bin/console
@@ -15,18 +15,14 @@ if (!class_exists(Application::class)) {
     throw new \RuntimeException('You need to add "symfony/framework-bundle" as a Composer dependency.');
 }
 
-$input = new ArgvInput();
-$env = $input->getParameterOption(['--env', '-e'], null, true);
-
-if (!isset($_SERVER['APP_ENV'])) {
-    if (!class_exists(Dotenv::class)) {
-        throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
-    }
-    (new Dotenv())->loadForEnv($env ?? 'dev', __DIR__.'/../.env');
+if (class_exists(Dotenv::class)) {
+    (new Dotenv())->loadForEnv($_SERVER['APP_ENV'] ?? 'dev', __DIR__.'/../.env');
+} elseif (!isset($_SERVER['APP_ENV']) && !isset($_ENV['APP_ENV'])) {
+    throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
 }
 
-$env = $env ?? $_SERVER['APP_ENV'] ?? 'dev';
-$debug = (bool) ($_SERVER['APP_DEBUG'] ?? ('prod' !== $env)) && !$input->hasParameterOption('--no-debug', true);
+$env = $_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'dev';
+$debug = (bool) ($_SERVER['APP_DEBUG'] ?? $_ENV['APP_DEBUG'] ?? ('prod' !== $env)) && !$input->hasParameterOption('--no-debug', true);
 
 if ($debug) {
     umask(0000);
--- a/symfony/console/4.2/bin/console
+++ b/symfony/console/4.2/bin/console
@@ -1,26 +1,14 @@
 #!/usr/bin/env php
 <?php
 
-use App\Kernel;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
-use Symfony\Component\Debug\Debug;
-
-set_time_limit(0);
-
-[$env, $debug] = require __DIR__.'/../src/bootstrap.php';
 
 if (!class_exists(Application::class)) {
     throw new \RuntimeException('You need to add "symfony/framework-bundle" as a Composer dependency.');
 }
 
-if ($debug) {
-    umask(0000);
+set_time_limit(0);
 
-    if (class_exists(Debug::class)) {
-        Debug::enable();
-    }
-}
-
-$kernel = new Kernel($env, $debug);
+$kernel = require __DIR__.'/../src/bootstrap.php';
 $application = new Application($kernel);
 $application->run($input);
--- a/symfony/flex/1.0/.env
+++ b/symfony/flex/1.0/.env
@@ -1,3 +1,3 @@
 # This file is a "template" of which env vars need to be defined for your application
-# Override the variabled you want in .env.local for development, create environment variables when deploying to production
+# Override the variables you want in .env.local for development, create environment variables when deploying to production
 # https://symfony.com/doc/current/best_practices/configuration.html#infrastructure-related-configuration
--- a/symfony/flex/1.0/.env
+++ b/symfony/flex/1.0/.env
@@ -1,3 +1,3 @@
 # This file is a "template" of which env vars need to be defined for your application
-# Copy this file to .env file for development, create environment variables when deploying to production
+# Override the variabled you want in .env.local for development, create environment variables when deploying to production
 # https://symfony.com/doc/current/best_practices/configuration.html#infrastructure-related-configuration
--- a/symfony/framework-bundle/4.2/public/index.php
+++ b/symfony/framework-bundle/4.2/public/index.php
@@ -12,7 +12,7 @@ if (!isset($_SERVER['APP_ENV']) && !isset($_ENV['APP_ENV'])) {
     if (!class_exists(Dotenv::class)) {
         throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
     }
-    (new Dotenv())->load(__DIR__.'/../.env');
+    (new Dotenv())->loadForEnv('dev', __DIR__.'/../.env');
 }
 
 $env = $_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'dev';
--- a/symfony/framework-bundle/4.2/public/index.php
+++ b/symfony/framework-bundle/4.2/public/index.php
@@ -4,10 +4,7 @@ use App\Kernel;
 use Symfony\Component\Debug\Debug;
 use Symfony\Component\HttpFoundation\Request;
 
-require __DIR__.'/../src/bootstrap.php';
-
-$env = $_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'dev';
-$debug = (bool) ($_SERVER['APP_DEBUG'] ?? $_ENV['APP_DEBUG'] ?? ('prod' !== $env));
+[$env, $debug] = require __DIR__.'/../src/bootstrap.php';
 
 if ($debug) {
     umask(0000);
--- a/symfony/framework-bundle/4.2/public/index.php
+++ b/symfony/framework-bundle/4.2/public/index.php
@@ -2,18 +2,9 @@
 
 use App\Kernel;
 use Symfony\Component\Debug\Debug;
-use Symfony\Component\Dotenv\Dotenv;
 use Symfony\Component\HttpFoundation\Request;
 
-require __DIR__.'/../vendor/autoload.php';
-
-// The check is to ensure we don't use .env if APP_ENV is defined
-if (!isset($_SERVER['APP_ENV']) && !isset($_ENV['APP_ENV'])) {
-    if (!class_exists(Dotenv::class)) {
-        throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
-    }
-    (new Dotenv())->loadForEnv('dev', __DIR__.'/../.env');
-}
+require __DIR__.'/../src/bootstrap.php';
 
 $env = $_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'dev';
 $debug = (bool) ($_SERVER['APP_DEBUG'] ?? $_ENV['APP_DEBUG'] ?? ('prod' !== $env));
--- a/symfony/framework-bundle/4.2/public/index.php
+++ b/symfony/framework-bundle/4.2/public/index.php
@@ -1,27 +1,17 @@
 <?php
 
-use App\Kernel;
-use Symfony\Component\Debug\Debug;
 use Symfony\Component\HttpFoundation\Request;
 
-[$env, $debug] = require __DIR__.'/../src/bootstrap.php';
-
-if ($debug) {
-    umask(0000);
-
-    Debug::enable();
-}
+$kernel = require __DIR__.'/../src/bootstrap.php';
 
 if ($trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? $_ENV['TRUSTED_PROXIES'] ?? false) {
     Request::setTrustedProxies(explode(',', $trustedProxies), Request::HEADER_X_FORWARDED_ALL ^ Request::HEADER_X_FORWARDED_HOST);
 }
-
 if ($trustedHosts = $_SERVER['TRUSTED_HOSTS'] ?? $_ENV['TRUSTED_HOSTS'] ?? false) {
     Request::setTrustedHosts(explode(',', $trustedHosts));
 }
-
-$kernel = new Kernel($env, $debug);
 $request = Request::createFromGlobals();
+
 $response = $kernel->handle($request);
 $response->send();
 $kernel->terminate($request, $response);
--- a/symfony/framework-bundle/4.2/src/bootstrap.php
+++ b/symfony/framework-bundle/4.2/src/bootstrap.php
@@ -4,8 +4,14 @@ require __DIR__.'/../vendor/autoload.php';
 
 use Symfony\Component\Dotenv\Dotenv;
 
-if (class_exists(Dotenv::class)) {
-    (new Dotenv())->loadForEnv($_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'dev', __DIR__.'/../.env');
-} elseif (!isset($_SERVER['APP_ENV']) && !isset($_ENV['APP_ENV'])) {
+$envFromEnv = $_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? null;
+$env = $envFromEnv ?? 'dev';
+$prod = 'prod' === $env;
+
+if (!$prod && class_exists(Dotenv::class)) {
+    (new Dotenv())->loadForEnv($env, __DIR__.'/../.env');
+} elseif (null === $envFromEnv) {
     throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
 }
+
+return [$env, (bool) ($_SERVER['APP_DEBUG'] ?? $_ENV['APP_DEBUG'] ?? !$prod)];
--- a/symfony/framework-bundle/4.2/src/bootstrap.php
+++ b/symfony/framework-bundle/4.2/src/bootstrap.php
@@ -12,6 +12,7 @@ $prod = 'prod' === $env;
 
 if (!$prod && class_exists(Dotenv::class)) {
     (new Dotenv())->loadForEnv($env, __DIR__.'/../.env');
+    $env = $_SERVER['APP_ENV'] ?? $env;
 } elseif (null === $envFromEnv) {
     throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
 }
--- a/symfony/framework-bundle/4.2/src/bootstrap.php
+++ b/symfony/framework-bundle/4.2/src/bootstrap.php
@@ -5,7 +5,7 @@ require __DIR__.'/../vendor/autoload.php';
 use Symfony\Component\Dotenv\Dotenv;
 
 if (class_exists(Dotenv::class)) {
-    (new Dotenv())->loadForEnv($_SERVER['APP_ENV'] ?? 'test', __DIR__.'/../.env');
+    (new Dotenv())->loadForEnv($_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'test', __DIR__.'/../.env');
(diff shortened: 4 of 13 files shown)


symfony/framework-bundle/4.2/public/index.php
```diff
@@ -12,7 +12,7 @@
     if (!class_exists(Dotenv::class)) {
         throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
     }
-    (new Dotenv())->load(__DIR__.'/../.env');
+    (new Dotenv())->loadForEnv('dev', __DIR__.'/../.env');
```
maxhelias: The first parameter should not be dynamic ?

dunglas: not by default

symfony/console/4.2/bin/console
```diff
@@ -0,0 +1,41 @@
...
+set_time_limit(0);
+
+require __DIR__.'/../vendor/autoload.php';
+
+if (!class_exists(Application::class)) {
+    throw new \RuntimeException('You need to add "symfony/framework-bundle" as a Composer dependency.');
+}
+
+$input = new ArgvInput();
+$env = $input->getParameterOption(['--env', '-e'], null, true);
```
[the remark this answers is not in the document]
dunglas: We need way to know which env we want. I thought about adding a new env var, but as this option already exist I prefer to reuse it.

The fallback is done later.

symfony/framework-bundle/4.2/public/index.php
maxhelias: I understand but some people still use the Dotenv for production and it could lead to errors

symfony/console/4.2/bin/console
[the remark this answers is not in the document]
ogizanagi: Sure, but I would have kept `$env = $input->getParameterOption(['--env', '-e'], $_SERVER['APP_ENV'] ?? 'dev');` as in https://gist.github.com/ogizanagi/84f6e9c306e1dfac7c66ecffce06b858#file-console-diff-L28

dunglas:
@maxhelias @ogizanagi I was thinking about introducing an env var simply called `LOAD_ENV`. When defined it will always load the given env file. It would resolve both of your comments, and can replace the `-e` argument in console.

To run the command in prod env: `LOAD_ENV=prod bin/console`
To un the integrated web server in prod env: `LOAD_ENV=prod ^hp -S 127.0.0.1:8000 -t public`

WDYT?

symfony/console/4.2/bin/console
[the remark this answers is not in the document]
dunglas: It would make the code harder to read because the call `loadForEnv($env)` can change the value `$_SERVER['APP_ENV']`. And if the value change because of the env file, it must be taken into account.

With the current approach, it's explicit that the value can come from the env file.

symfony/flex/1.0/.env
```diff
@@ -1,3 +1,3 @@
 # This file is a "template" of which env vars need to be defined for your application
-# Copy this file to .env file for development, create environment variables when deploying to production
+# Override the variabled you want in .env.local for development, create environment variables when deploying to production
```
sroze: `variables`

sroze: Also, this would only be correct with FWB > 4.2. I don't really know how, but ideally, we want this change to only be there for FWB > 4.2.

dunglas: Yes I don't know how to handle this properly for old versions. The only way I see is creating a Flex 2.0 version.

symfony/console/4.2/bin/console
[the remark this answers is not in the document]
ogizanagi: If `$_SERVER['APP_ENV']` is set, then the DotEnv component will never override it, right?

I'm not convinced about the `LOAD_ENV` suggestion, as to me the `APP_ENV` is really meant to replace the `--env` option and nothing else.
The fact the `.env` file also sets the `APP_ENV` is precisely a fallback for when the `APP_ENV` is not explicitly provided. Like any other env var.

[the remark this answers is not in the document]
dunglas: > If $_SERVER['APP_ENV'] is set, then the DotEnv component will never override it, right?

Well, it's the problem. If `APP_ENV` is defined, DotEnv will never be called. It's intended (in prod real env var should be used). But we assume that if `APP_ENV` is defined, all other mandatory environment variables are defined. So just running `APP_ENV=prod bin/console` will most likely not work, you'll have to run `APP_ENV=prod FOO=BAR FOO2=BAR2 [...] bin/console`. While it's easy in production, and when using Docker or Kubernetes, for most user its not a good a DX.

As we cannot use `APP_ENV` to hint that we need to load a specific env file (because it means, the environnement is already properly configured, you don't need DotEnv), the idea is to introduce `LOAD_ENV` to be this hint.
`LOAD_ENV=prod bin/console` means "load .env.prod" and use it.

`LOAD_ENV` will be a debug tool (that can also be leveraged for people not able to use K8S/Docker etc: just set LOAD_ENV=production in the Apache config, and then you can manage your config using .env.production directly, just like you used to do with parameters.yml).

Another benefit:

`LOAD_ENV=panther bin/phpunit`

```
# .env.panther

DATABASE_URL=mysql://test
```

And you can use a specific DB just for Panther.

wdyt @ogizanagi?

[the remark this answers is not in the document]
ogizanagi: > Well, it's the problem. If APP_ENV is defined, DotEnv will never be called. It's intended (in prod real env vars should be used). But we assume that if APP_ENV is defined, all other mandatory environment variables are defined. 

Sure. That's why in [the `DotEnvLoader` I'm using](https://gist.github.com/ogizanagi/84f6e9c306e1dfac7c66ecffce06b858#file-dotenvloader-php-L14-L15), I do only throw when **no `APP_ENV` is provided** and **no DotEnv component is available**, as the DotEnv component won't even be installed on production (`composer install --no-dev`).
Then, even if the `APP_ENV` env var is provided, as soon as the DotEnv component is available, I use it to fulfill other env vars from `.env` files.

So, in production without DotEnv component, `.env` file won't be loaded (_In our case, it these files aren't even part of the production release_).
During development, one can provide the `APP_ENV` env var to choose environment, as well as any other env var if they want, and appropriate `.env.*` files will be loaded, fulfilling but not overriding env vars.

Does it make sense?

I don't have counter-argument about the other benefit you mentioned with `LOAD_ENV`, but I don't know if it'll be that much common.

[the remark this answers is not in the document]
dunglas: Got it! Let me try.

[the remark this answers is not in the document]
dunglas: @ogizanagi here we are! Very good idea.
And it even allows to do `APP_ENV=panther bin/phpunit` now.

symfony/framework-bundle/4.2/public/index.php
```diff
@@ -7,28 +7,26 @@
...
 require __DIR__.'/../vendor/autoload.php';
 
-// The check is to ensure we don't use .env if APP_ENV is defined
-if (!isset($_SERVER['APP_ENV'])) {
-    if (!class_exists(Dotenv::class)) {
-        throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
-    }
-    (new Dotenv())->load(__DIR__.'/../.env');
+if (class_exists(Dotenv::class)) {
+    (new Dotenv())->loadForEnv($env ?? 'dev', __DIR__.'/../.env');
```
ogizanagi: `$env` is never defined here :)

symfony/console/4.2/bin/console
```diff
@@ -0,0 +1,37 @@
...
+set_time_limit(0);
+
+require __DIR__.'/../vendor/autoload.php';
+
+if (!class_exists(Application::class)) {
+    throw new \RuntimeException('You need to add "symfony/framework-bundle" as a Composer dependency.');
+}
+
+if (class_exists(Dotenv::class)) {
+    (new Dotenv())->loadForEnv($_SERVER['APP_ENV'] ?? 'dev', __DIR__.'/../.env');
```
joshlopes: what would be the outcome here if my `.env` had APP_ENV set to, say `STAGING` would `.env.dev` be loaded?

symfony/framework-bundle/4.2/public/index.php
joshlopes: same here, if .env loads other env that not dev you will try to load .dev at all times, and it may not be the case / what we want. That would not have happened before.

symfony/console/4.2/bin/console
dunglas: No. .env and .env.STAGING will.

dunglas:
To upgrade, users will have to use a script like this one:

```bash
cat composer.json | jq --indent 4 '.require["symfony/flex"] = "^2.0"' > composer.tmp
mv composer.tmp composer.json
cp .env .env.local
cp .env.dist .env
sed -i'' -e 's/\/.env/\/.env.local\
\/.env.*.local/' .gitignore
```

symfony/framework-bundle/4.2/public/index.php
```diff
@@ -7,28 +7,26 @@
...
+} elseif (!isset($_SERVER['APP_ENV']) && !isset($_ENV['APP_ENV'])) {
+    throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
 }
 
-$env = $_SERVER['APP_ENV'] ?? 'dev';
-$debug = (bool) ($_SERVER['APP_DEBUG'] ?? ('prod' !== $env));
+$env = $_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'dev';
+$debug = (bool) ($_SERVER['APP_DEBUG'] ?? $_ENV['APP_DEBUG'] ?? ('prod' !== $env));
```
weaverryan: Could we centralize some of this repeated logic between `index.php` and `console? Specifically, the loading of the DotEnv stuff and this logic. It *must* be consistent between console/index.php & testing. So I think it's important.

symfony/framework-bundle/4.2/public/index.php
dunglas: @ogizanagi good catch, fixed.
@joshlopes no, because it uses https://github.com/symfony/symfony/pull/28533

symfony/framework-bundle/4.2/public/index.php
dunglas: done, I just moved `bootstrap.php` in `src/` and used it in `index.php` and `console`.

symfony/framework-bundle/4.2/src/bootstrap.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+require __DIR__.'/../vendor/autoload.php';
+
+use Symfony\Component\Dotenv\Dotenv;
+
+if (class_exists(Dotenv::class)) {
+    (new Dotenv())->loadForEnv($_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'test', __DIR__.'/../.env');
```
bocharsky-bw: Why not `dev` as a fallback if both `$_SERVER['APP_ENV']` and `$_ENV['APP_ENV']` are not set?

dunglas: You caught a bug! Thanks. Fixed in the last commit.

dunglas:
@bocharsky-bw
> Great idea with LOAD_ENV=prod bin/console 👍

Thanks to the suggestion of @ogizanagi, this var isn't necessary anymore. You can just run `APP_ENV=prod bin/console` and it will work.

bocharsky-bw:
Hm, I'm not sure... the idea is to be able to load websites in `prod` using Symfony web server. This web server should be registered in `dev` only, that means if I'll try to run:
```bash
APP_ENV=prod bin/console server:run
```

I'll see an error about missing command, because we're loading Symfony console in `prod` env that does not have `server:run` command registered (only `dev` has).

So I think we still need `LOAD_ENV`. If I understand it right, the command:
```bash
LOAD_ENV=prod bin/console server:run
```

Will allow us to run `bin/console` in `dev` mode, however, index.php will be loaded in `prod`, right?

dunglas:
No, if the bundle isn't registered for the given environment, it will never work. With `LOAD_ENV` the problem would be the same, the if the env is `prod` and the bundle is not registered for `prod`, it will not be loaded.

symfony/framework-bundle/4.2/src/bootstrap.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+require __DIR__.'/../vendor/autoload.php';
+
+use Symfony\Component\Dotenv\Dotenv;
+
+if (class_exists(Dotenv::class)) {
+    (new Dotenv())->loadForEnv($_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'dev', __DIR__.'/../.env');
```
iltar: This change will always load the .env if I understand it correctly. Now if I make a mistake in composer.json or a composer command where I forget `--dev`, it will move `Dotenv` from `require-dev` to `require` and suddenly my production server is using the wrong variables. The same goes for installation when missing `--no-dev`, you will deploy _with_ Dotenv and break your production, possibly connecting to the wrong database.

I have not followed all discussions, I came here from https://github.com/symfony/symfony/issues/25693#issuecomment-424329159

But this sounds like a very dangerous change that has the potential to break things _by default_. You have to change something in order to _not_ load the wrong data (`--no-dev`) and it's unforgiving when you make a mistaken with your dependencies. Both of those which I see happen all the time with developers new to composer.

dunglas: @iltar yes it's 100% intended (see the proposal of @ogizanagi in the old comments of this PR).

It allows user that can't (or don't want to) properly setup environment variables to use `.env.local` as they was using `parameters.yml` with Symfony 3, while still following (by default and by design) 12factor methodology/cloud native.

It also fixes a lot of usability problems discussed here.

iltar: I get that this is the intention, but it would make it really easy to critically break production environments if you're not paying attention, which sounds too much of a risk for the possible gain.

To me that sounds like a no-go.

Pierstoval: @iltar If the env var is already set, DotEnv won't override it anyway, maybe this could comfort you a bit :D

ogizanagi: By default, the skeleton has [`symfony/dotenv` under the require-dev section](https://github.com/symfony/skeleton/blob/master/composer.json).
The [documentation](https://symfony.com/doc/current/components/dotenv.html) mentions clearly the Dotenv component is not recommended to be used on production. But we must probably update the installation section to add the `--dev` option.

Worst case indeed is the wrong env value will be used if the Dotenv is installed on production, but only if the env var is actually missing on the server. That should be prevented by learning developers new to composer how to take advantage of its features and ensuring your deployment properly uses the `--no-dev` option. When deploying you could also remove the `.env` file for safety. 
IMHO that shouldn't prevent us from doing the changes in this PR.

However, if anyone else is reluctant to these changes as is, we can just add an extra `'prod' !== $env` condition before actually calling `Dotenv:: loadForEnv()`, as a tradeoff. Developers knowing what it implied can still remove it easily.

maxhelias: Maybe add `\trigger_error('You should not use Dotenv in production', \E_USER_WARNING);` here or inthe Dotenv component to warn users ?

dunglas: One of the goal is to allow users to use Dotenv in production (even if it's not recommended, it's not risked, it's just slower and not a good practice). I don't think we should add this warning. We should just document it properly.

And as pointed by @Pierstoval and @ogizanagi there are no security implication at all because if an env var is already set, it will never be overrode.

xabbuh: There is still some risk here: Some code may only be executed if a certain environment variable is set. And the Dotenv component may not be installed accidentally by the developer themself, but because some third-party package has it in its `require` section.

symfony/flex/1.0/.env
dunglas: I don't think so. And it doesn't help, you can have FWB 2.2 with Flex 1.0, or more likely Flex 1.1 with Symfony 3.4.

symfony/framework-bundle/4.2/src/bootstrap.php
dunglas: I'm not sure to understand, can you give an example?

> some third-party package has it in its require section

It would be a super bad practice.

ogizanagi: I agree. I can't see any good reason for a third-party bundle to require the Dotenv component. The component targets end-projects.

iltar: We are all developers and we know that people apply bad practices all the time. Sometimes developers need to be protected and this is one of those scenarios where the little bit of protection can prevent a catastrophic failure.

dunglas: How can it be catastrophic? Existing env vars are never overrode.

stof: this change does not allow using `.env` to specify the value for `APP_ENV`, while it is the current documented way to run the app in a different environment easily locally today.

AFAICt, changing the env locally now requires using a real env variable.

symfony/framework-bundle/4.2/src/bootstrap.php
```diff
@@ -0,0 +1,11 @@
...
+} elseif (!isset($_SERVER['APP_ENV']) && !isset($_ENV['APP_ENV'])) {
+    throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
+}
```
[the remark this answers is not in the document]
stof: code is not the same, as the console file supports using `--env` to provide the environment, and to pass `--no-debug` to force no-debug mode.

symfony/framework-bundle/4.2/src/bootstrap.php
iltar: If I have no APP_ENV set (or any others), and the package ended up in the require, in this example it will try to load the DEV env. If for some reason the development database etc can be reached (which isn't unlikely), or a development micro-service, all requests in production will start talking to this environment. This can lead to incorrect payments, sensitive data leaking because of user ids being different and for some reason test data is using semi-real production data etc.

In this scenario, I rather have my application crash into a 500 than having it connect to the wrong services. By making it the default behavior, you'll open up a can of worms that can harm companies by making an innocent mistake. I'd rather see the development fail on first try, than production, even if this hurts DX in the process. 

```php
(new Dotenv())->loadForEnv($_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'dev', __DIR__.'/../.env')
```

ogizanagi: You are referring to https://symfony.com/doc/current/configuration/environments.html#executing-an-application-in-different-environments, right?

AFAIU, this still works (`$env` still is obtained in front controllers by executing: 
```php
$env = $_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'dev'
```
after loading the `.env` file). 

But for convenience, if you really need to access often your app in another env but cannot provide it in your server configuration, I'd rather recommend creating a dedicated front controller. Not to be commit/deployed of course.

Just changing the `APP_ENV` value in `.env` will change the environment used by the app, but will keep all other env vars dev values, which seems a bit weird.

symfony/framework-bundle/4.2/src/bootstrap.php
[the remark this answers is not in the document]
ogizanagi: Sure but checking if `$input` is defined still is a possibility and anyway would be necessary regarding https://github.com/symfony/recipes/pull/466/files#r219814650

symfony/console/4.2/bin/console
```diff
@@ -0,0 +1,29 @@
...
+
+set_time_limit(0);
+
+require __DIR__.'/../src/bootstrap.php';
+
+if (!class_exists(Application::class)) {
+    throw new \RuntimeException('You need to add "symfony/framework-bundle" as a Composer dependency.');
+}
+
+$env = $_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'dev';
```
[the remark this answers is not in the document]
ogizanagi: This comment can be ignored now since https://github.com/symfony/symfony/pull/28653 is merged :) (thanks @chalasr)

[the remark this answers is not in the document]
chalasr: I think we should still support it until 5.0 :)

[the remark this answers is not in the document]
ogizanagi: I wonder if this is really necessary. This recipe will automatically be used for new projects only. Deprecations & BC layers are only meant for existing projects & libraries. A new project should be free of deprecated features.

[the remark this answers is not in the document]
chalasr: Alright, upgrading the recipe involves to update the code, fine for me as-is. 
The `--no-debug` fallback below can be removed as well

symfony/framework-bundle/4.2/src/bootstrap.php
Tobion: I also feel loading env file just because DotEnv is installed is too risky as the default. The compromise suggested by @ogizanagi would be my preferred solution.

> However, if anyone else is reluctant to these changes as is, we can just add an extra 'prod' !== $env condition before actually calling Dotenv:: loadForEnv(), as a tradeoff. Developers knowing what it implies can still remove it easily.

Tobion: > this change does not allow using .env to specify the value for APP_ENV, while it is the current documented way to run the app in a different environment easily locally today.

This to me is a no-go as it has been teached like this since introducing the .env in sf. Breaking this behavior would catch alot of people offguard.

ogizanagi: As I said above, unless I’m missing something, this still works. But I really don’t think that’s the easiest way.

dunglas: @Tobion this isn't the case, as explainded by @ogizanagi: https://github.com/symfony/recipes/pull/466/files#diff-7d73eabd1e5eb7d969ddf9a7ce94f954R9

weaverryan:
@Jean85 Which part don't you like specifically? Is it the filenames are being renamed (`.env.dist` to `.env` and committed + `.env` to `.env.local` and not committed)? Or some other part? There are several moving pieces with this change - I'm trying to see where the negative feelings are coming from specifically.

Jean85:
The filenames being changed are a big pain point IMHO. It's a practical BC and it will mess with current git repos of applications that uses that convention: migrating from one to the other will be a mess on each local environment, and sometimes on a CI setup too.

I understand wanting to minimize the diff/override in different envs, but I think we should add different files, not change the existing ones; that would also mimic the previous behavior of `parameters.yaml`, which would ease the migration.

dunglas:
@Jean85 I'll change the Flex path to support both the old (`.env.dist`)  and the new flavor.

symfony/console/4.2/bin/console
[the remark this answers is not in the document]
dunglas: done

symfony/framework-bundle/4.2/src/bootstrap.php
dunglas: done

dunglas: I updated the PR, can you check again please?

symfony/framework-bundle/4.2/src/bootstrap.php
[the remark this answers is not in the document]
dunglas: Done in the last commit

B-Galati:
Should we also deprecate the default value notation from services files?
i.e.:

```yaml
# config/services.yaml
parameters:
    env(DATABASE_HOST): localhost
```

https://symfony.com/doc/current/configuration/external_parameters.html#environment-variables

symfony/framework-bundle/4.2/src/bootstrap.php
```diff
@@ -4,8 +4,14 @@
...
 use Symfony\Component\Dotenv\Dotenv;
 
-if (class_exists(Dotenv::class)) {
-    (new Dotenv())->loadForEnv($_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'dev', __DIR__.'/../.env');
-} elseif (!isset($_SERVER['APP_ENV']) && !isset($_ENV['APP_ENV'])) {
+$envFromEnv = $_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? null;
+$env = $envFromEnv ?? 'dev';
+$prod = 'prod' === $env;
+
+if (!$env && class_exists(Dotenv::class)) {
```
ogizanagi: `$env` cannot be null at this point. Did you mean `$prod` ?

dunglas: indeed, fixed

symfony/framework-bundle/4.2/src/bootstrap.php
```diff
@@ -0,0 +1,17 @@
...
+require __DIR__.'/../vendor/autoload.php';
+
+use Symfony\Component\Dotenv\Dotenv;
+
+$envFromEnv = $_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? null;
+$env = $envFromEnv ?? 'dev';
+$prod = 'prod' === $env;
+
+if (!$prod && class_exists(Dotenv::class)) {
+    (new Dotenv())->loadForEnv($env, __DIR__.'/../.env');
```
ogizanagi: Should we add `$env = $_SERVER['APP_ENV'] ?? $env` after this line, so https://github.com/symfony/recipes/pull/466#discussion_r220616063 still works? I.e changing `APP_ENV` in .env (and no APP_ENV env var provided) would still allow running in a different env?

dunglas: Good idea, I also factorized most of the code in `console` and `index.php`

symfony/framework-bundle/4.2/public/index.php
```diff
@@ -1,39 +1,17 @@
...
-
-$env = $_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'dev';
-$debug = (bool) ($_SERVER['APP_DEBUG'] ?? $_ENV['APP_DEBUG'] ?? ('prod' !== $env));
-
-if ($debug) {
-    umask(0000);
-
-    Debug::enable();
-}
+$kernel = require __DIR__.'/../src/bootstrap.php';
```
fabpot: I'm not a big fan of having a `bootstrap.php` file. what about moving the logic to the `Kernel` class instead?

B-Galati:
@ogizanagi it does not relate sorry. I will open an issue on Symfony repository.

symfony/framework-bundle/4.2/src/bootstrap.php
```diff
@@ -0,0 +1,29 @@
...
+
+require __DIR__.'/../vendor/autoload.php';
+
+use App\Kernel;
+use Symfony\Component\Debug\Debug;
+use Symfony\Component\Dotenv\Dotenv;
+
+$envFromEnv = $_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? null;
+$env = $envFromEnv ?? 'dev';
+$prod = 'prod' === $env;
```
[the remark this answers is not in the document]
ogizanagi: My opinion is that it complicates everything to introduce more and more exceptions and would make it far from intuitive. The prod env restriction has a reason, it's a safeguard for real production environments. 
For other env, I don't understand the reasons yet. The main reason for this feature to exist is `env.test` and that's how it'll be used 95% of the time. But some applications defines their own specific symfony env(s) and may also benefit from this. And if you don't, then you shouldn't care putting arbitrary restrictions on a feature already based on conventions :)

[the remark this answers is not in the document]
dunglas: I’m fine with this

symfony/console/4.2/bin/console
```diff
@@ -0,0 +1,14 @@
...
+
+if (!class_exists(Application::class)) {
+    throw new \RuntimeException('You need to add "symfony/framework-bundle" as a Composer dependency.');
+}
+
+set_time_limit(0);
+
+$kernel = require __DIR__.'/../src/bootstrap.php';
+$application = new Application($kernel);
+$application->run($input);
```
XavierVallot: this variable `$input` is not defined though the code cannot work

symfony/console/4.2/bin/console
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env php
+<?php
+
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+
+if (!class_exists(Application::class)) {
```
XavierVallot: the autoload has not yet been called at this point so this condition is always false and always throw the exception even if `framework-bundle` is installed

Kocal: Maybe because there is no `require __DIR__.'/../vendor/autoload.php';` in this file

symfony/framework-bundle/4.2/src/bootstrap.php
```diff
@@ -0,0 +1,29 @@
+<?php
```
XavierVallot: this class lives under `src/` which is autoloaded, so composer excepts a class which this file is not. It gives the following error : 
```
In FileLoader.php line 168:
                                                                                                                                                        
  The autoloader expected class "App\bootstrap" to be defined in file "/srv/app/vendor/composer/../../src/bootstrap.php". The file was found but the c  
  lass was not in it, the class name or namespace probably has a typo in /srv/app/config/services.yaml (which is loaded in resource "/srv/app/config/s  
  ervices.yaml").                                                                                                                                       
                                                                                                                                                        

In DebugClassLoader.php line 204:
                                                                                                                                                        
  The autoloader expected class "App\bootstrap" to be defined in file "/srv/app/vendor/composer/../../src/bootstrap.php". The file was found but the c  
  lass was not in it, the class name or namespace probably has a typo.   
```

dunglas:
Closing in favor of #481

PR #469, opened 2018-09-30
Proposal: Check for CONSOLE_APP_ENV first before APP_ENV in bin/console

bocharsky-bw:
| Q             | A
| ------------- | ---
| License       | MIT

Probably the most elegant alternative solution of both #467 and #461

With these changes we can load websites in any environment using Symfony Web Server having WebServerBundle registered for dev environment only.

For example, with the next command I can load website in `prod` mode:
```bash
CONSOLE_APP_ENV=dev APP_ENV=prod bin/console server:run
```


maxhelias:

PR #471, opened 2018-10-06
Fix routes support when using the PHP built-in web server

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

Currently, when using the built-in PHP web server (as suggested when installing the skeleton), if a route looks like a static file, Symfony isn't even triggered:

Example with the default installation of API Platform:

    api_doc              ANY      ANY      ANY    /api/docs.{_format}

    $ curl http://127.0.0.1:8000/api/docs.jsonld
    <!doctype html><html><head><title>404 Not Found</title><style>

(This 404 error is generated by the built-in web server, not by Symfony because it is not even triggered).

With this fix, it works, and real static files (assets...) are still served correctly.

Note: this error doesn't happen when using Web Server Bundle because it already contains a similar fix.

### DIFF
--- a/symfony/framework-bundle/3.3/public/index.php
+++ b/symfony/framework-bundle/3.3/public/index.php
@@ -24,7 +24,7 @@ if ($debug) {
     Debug::enable();
 }
 
-if (PHP_SAPI === 'cli-server' && '/index.php' !== $_SERVER['SCRIPT_NAME'] && is_file(__DIR__.                 $_SERVER['SCRIPT_NAME'])) {
+if (PHP_SAPI === 'cli-server' && '/index.php' !== $_SERVER['SCRIPT_NAME'] && is_file(__DIR__.$_SERVER['SCRIPT_NAME'])) {
     return false;
 }
 
--- a/symfony/framework-bundle/3.3/public/index.php
+++ b/symfony/framework-bundle/3.3/public/index.php
@@ -24,10 +24,6 @@ if ($debug) {
     Debug::enable();
 }
 
-if ('cli-server' === PHP_SAPI && '/index.php' !== $_SERVER['SCRIPT_NAME'] && is_file(__DIR__.$_SERVER['SCRIPT_NAME'])) {
-    return false;
-}
-
 if ($trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? $_ENV['TRUSTED_PROXIES'] ?? false) {
     Request::setTrustedProxies(explode(',', $trustedProxies), Request::HEADER_X_FORWARDED_ALL ^ Request::HEADER_X_FORWARDED_HOST);
 }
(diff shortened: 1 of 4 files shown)


symfony/framework-bundle/3.3/public/index.php
```diff
@@ -24,6 +24,10 @@
     Debug::enable();
 }
 
+if ('cli-server' === PHP_SAPI && '/index.php' !== $_SERVER['SCRIPT_NAME'] && is_file(__DIR__.$_SERVER['SCRIPT_NAME'])) {
+    return false;
+}
```
fabpot: Not a big fan of adding something like this for all requests.

dunglas: Me neither, my first attempt was to put if inside the `if ($debug)` block. Would you prefer this?
I changed this behavior because sometimes it's convenient to use the built-in web server in "prod" mode to debug, but it can be considered an edge case.

weaverryan: Yea, this block looks confusing... just to fix an edge case situation (where people should be using the web server bundle)

fabpot: Indeed, @weaverryan is right.

teohhanhui: > (where people should be using the web server bundle)

Doesn't it suffer from this same issue, since it's a wrapper around the PHP built-in web server?

dunglas: @weaverryan @fabpot in this case we should not explain how to use the built-in web server in the post install message and require web server bundle as a dev dependency of the skeleton, because it currently it’s very disappointing for the end user. Here is how I figured the problem:

1. During a hackathon, a  newcomer installs Symfony through composer, then API Platform using composer req api
2. It starts the built-in PHP server using the displayed post install command
3. It tries to generate a client app using the client generator of API Platform

It fails because the generator try to access the /docs.jsonld route, and this bug prevent it to work properly.

It’s happening with the default integration of API Platform with Symfony, but it can also happen in a lot of unrelated use cases, and it’s almost impossible to figure what ‘s going on for a newcomer.

@teohhanhui no it doesn’t, because it comes with a router script fixing this problem.

soyuka: > we should not explain how to use the built-in web server in the post install message and require web server bundle as a dev dependency of the skeleton

:+1:

fabpot: @dunglas I agree with you, let's remove the reco from the post install message and recommend to use the web server bundle instead.

dunglas: @fabpot done

soufianZantar: @dunglas and @fabpot i haven't seen this PR before adding my PR #475 .
Adding **public/index.php** ( in server command  `php -S 127.0.0.1:8000 -t public public/index.php` ) it works fine for me and we can keep how to use the built-in web server in the post install  .
it's a good solution ?

dunglas: Not it's not enough (see previous commits).

PR #475, opened 2018-10-14
fix php built-in server command in what's next

soufianZantar:
| Q             | A
| ------------- | ---
| License       | MIT

Fix php built-in server command in **what's next** when creating a new symfony project .
Issues Ticket in Symfony Repo [#28781](https://github.com/symfony/symfony/issues/28781)

When someone run php built-in server with the old command and in the routing parameters must add dot '.', the server don't redirect the request to front controller index.php.

> for example: /page/{name} => /page/one.pl

the server will search for a file named  page/one and with extension .pl in the public folder.

adding public/index.php to the command fix that bug and all requests will passed by index.php

### DIFF
--- a/symfony/framework-bundle/4.2/post-install.txt
+++ b/symfony/framework-bundle/4.2/post-install.txt
@@ -5,7 +5,7 @@
   * <fg=blue>Run</> your application:
     1. Change to the project directory
     2. Create your code repository with the <comment>git init</comment> command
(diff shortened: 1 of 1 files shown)


xabbuh:
We recently decided to drop this entire paragraph in favour of advising to use the WebServerBundle (see #471). That's why I am going to close here. Thank you for proposing anyway.

PR #477, opened 2018-10-26
Add MercureBundle recipe

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/mercure-bundle/0.1/manifest.json
+++ b/symfony/mercure-bundle/0.1/manifest.json
@@ -7,11 +7,11 @@
     },
     "container": {
         "env(MERCURE_PUBLISH_URL)": "",
-        "env(MERCURE_JWT)": ""
+        "env(MERCURE_JWT_SECRET)": ""
     },
     "env": {
         "MERCURE_PUBLISH_URL": "https://demo.mercure.rocks/hub",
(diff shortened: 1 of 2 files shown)


symfony/mercure-bundle/0.1/manifest.json
```diff
@@ -0,0 +1,17 @@
+{
+    "bundles": {
+        "Symfony\\Bundle\\MercureBundle\\MercureBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "container": {
+        "env(MERCURE_PUBLISH_URL)": "",
```
Pierstoval: Why add them as defaults in the container when you have them set up in env? 🤔

dunglas: To prevent an error when env var are not defined (when building a Docker container for instance).

sroze: You can instead define the `env(MERCURE_PUBLISH_URL)` parameter in the bundle.

dunglas: It’s not the bundle’s responsibility: you may use it without env cars.
Also, the use may choose to use other names of env vars.

Issue #476, opened 2018-10-16
symfony/phpunit-bridge overwrite some phpunit.xml parameters at runtime


MichaelDemeyer:
bin/phpunit file check following env parameters at runtime :
- SYMFONY_DEPRECATIONS_HELPER
- SYMFONY_PHPUNIT_REMOVE
- SYMFONY_PHPUNIT_VERSION
- SYMFONY_PHPUNIT_DIR
and set a default value if they're not already defined.
Values defined in phpunit.xml are not defined as env values at this point, so default values will apply.

vendor/symfony/phpunit-bridge/bin/simple-phpunit file do not apply phpunit.xml values if env variables are already set.

So basically it is impossible to set the 4 env variables using phpunit.xml

Step to reproduce :
Add 
        `<env name="SYMFONY_DEPRECATIONS_HELPER" value="disabled" force="true"/>`
to phpunit.xml
run bin/phpunit
SYMFONY_DEPRECATIONS_HELPER value will be 999999

leevigraham:
See: https://github.com/symfony/symfony/issues/28726

nicolas-grekas:
FIxed at the recipe level in https://github.com/symfony/recipes/pull/481
setting there vars in the phpunit.xml files is too late. They should be set in .env.test instead (see recipe).

PR #478, opened 2018-10-26
Fixed symfony env variables in phpunit.xml configuration

alexander-schranz:
| Q             | A
| ------------- | ---
| License       | MIT

fixes symfony/symfony#28726, symfony/recipes#476

With symfony/symfony#28995 its not longer needed that the env variables are needed to be set in the phpunit script for parallel test runs. Also without the patch this putenv make I think more problems as nobody opens the `bin/phpunit` file and it will ignore the phpunit.xml.dist configuration.

### DIFF
--- a/symfony/phpunit-bridge/4.1/bin/phpunit
+++ b/symfony/phpunit-bridge/4.1/bin/phpunit
@@ -5,16 +5,6 @@ if (!file_exists(dirname(__DIR__).'/vendor/symfony/phpunit-bridge/bin/simple-php
     echo "Unable to find the `simple-phpunit` script in `vendor/symfony/phpunit-bridge/bin/`.\n";
     exit(1);
 }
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Sorry but that cannot work: these env vars must be read before phpunit starts - it before phpunit.xml files are opened.

alexander-schranz:
@nicolas-grekas that is be fixed in symfony/symfony#28995 that the env variables need to be set before.

nicolas-grekas:
Oh ok, I'll have a look. Note that I might have another plan. Give me a few days to explain it.

alexander-schranz:
@nicolas-grekas did rebase the other pull request to master and as the other dev suggest I did avoid using global.

In the [symfony pull request](https://github.com/symfony/symfony/pull/28995) a test setup can be found similar to https://github.com/alexander-schranz/symfony-phpunit-test.

PR #479, opened 2018-10-28
Remove deprecated "--env" and "--no-debug" console options

Tobion:
| Q             | A
| ------------- | ---
| License       | MIT

Continuation of https://github.com/symfony/symfony/pull/28653
Since the recipes are only used for new projects this seems safe to do. We do not want to promote using these options anymore.

### DIFF
--- a/symfony/console/3.3/bin/console
+++ b/symfony/console/3.3/bin/console
@@ -3,7 +3,6 @@
 
 use App\Kernel;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
-use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Debug\Debug;
 use Symfony\Component\Dotenv\Dotenv;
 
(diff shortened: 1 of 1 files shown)


Tobion:
Documentation wise https://symfony.com/doc/3.4/console/usage.html still mentions these options for symfony < 4. Maybe we should remove the page for 3.4 as well and add this part instead: https://symfony.com/doc/current/console.html#the-console-app-env-app-debug

chalasr:
@Tobion It's similar to the new directory structure: the old translations/templates locations are still part of the 3.4 docs but the new ones are not, while they are actually supported since 3.4.
I believe we consider `APP_ENV` a Symfony 4 thing "officially", even if we started experimenting it in 3.4

ogizanagi:
#466 will contain these changes in a 4.2 recipe. I'd suggest either to wait for it to be merged, or create the missing 4.2 recipe for console instead.

nicolas-grekas:
Should we really do this? In https://github.com/symfony/symfony/issues/28984, people are complaining removing these attributes make the console harder to use for them. We did deprecate the options from core - and this was the thing to do to me also. BUT shouldn't we continue to allow using them here?

ogizanagi:
That's for new projects or people manually upgrading to new versions of these recipes. A new project should be free of deprecations by default, that's my main argument here. No stronger opinion, but that's what we used to do.

nicolas-grekas:
There is another way to be free from the deprecation: extract these from $argv right into bin/console.

chalasr:
I think we can close, #481

PR #481, opened 2018-11-08
Allow to override .env files per env

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

This PR is an alternative to #466 and replaces #479.
It needs https://github.com/symfony/symfony/pull/29129 to work with 4.2.

What we have here:
- Dotenv is configured to load `.env` < `.env.local` < `.env.$env` < `.env.$env.local` (only the 1st is required, see PR linked above)
- this works starting with Symfony 3.4 and uses an optimized code starting with 4.2 (removing the doc-split hell)
- the loading logic is factorized in a new static `Kernel::bootstrapEnv()` method
- the CLI accepts and *consumes* `--env` & `--no-debug` (see `Kernel::bootstrapCliEnv()`, fixes https://github.com/symfony/symfony/issues/23343)
- `.env` files are loaded for tests (fixes https://github.com/symfony/flex/issues/317, https://github.com/symfony/symfony/issues/28895 and https://github.com/symfony/symfony/issues/28726)
- config for the test env is moved from `phpunit.xml.dist` to `.env.test`
- `.env.dist` is not created anymore (commit `.env` instead and use `.env.local` for local overrides)
- `.gitignore` is updated accordingly

Regarding flex integration:
- this PR doesn't change anything for existing projects - they won't notice
- when flex sees a `.env.test` file, it should skip updating `phpunit.xml` files
- that's it - all the rest should remain identical, isn't it?
- (people that still want to use `.env.dist` just need to create the file and flex will populate it as it does today)

### DIFF
--- a/symfony/console/3.3/bin/console
+++ b/symfony/console/3.3/bin/console
@@ -3,30 +3,20 @@
 
 use App\Kernel;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
-use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Debug\Debug;
-use Symfony\Component\Dotenv\Dotenv;
 
 set_time_limit(0);
 
-require __DIR__.'/../vendor/autoload.php';
+require dirname(__DIR__).'/vendor/autoload.php';
 
 if (!class_exists(Application::class)) {
     throw new \RuntimeException('You need to add "symfony/framework-bundle" as a Composer dependency.');
 }
 
-if (!isset($_SERVER['APP_ENV'])) {
-    if (!class_exists(Dotenv::class)) {
-        throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
-    }
-    (new Dotenv())->load(__DIR__.'/../.env');
-}
+Kernel::bootstrapCli($_SERVER['argv']);
+Kernel::bootstrapEnv();
 
-$input = new ArgvInput();
-$env = $input->getParameterOption(['--env', '-e'], $_SERVER['APP_ENV'] ?? 'dev', true);
-$debug = (bool) ($_SERVER['APP_DEBUG'] ?? ('prod' !== $env)) && !$input->hasParameterOption('--no-debug', true);
-
-if ($debug) {
+if ($_SERVER['APP_DEBUG']) {
     umask(0000);
 
     if (class_exists(Debug::class)) {
@@ -34,6 +24,6 @@ if ($debug) {
     }
 }
 
-$kernel = new Kernel($env, $debug);
+$kernel = new Kernel($_SERVER['APP_ENV'], $_SERVER['APP_DEBUG']);
 $application = new Application($kernel);
-$application->run($input);
+$application->run();
--- a/symfony/framework-bundle/3.3/public/index.php
+++ b/symfony/framework-bundle/3.3/public/index.php
@@ -2,23 +2,13 @@
 
 use App\Kernel;
 use Symfony\Component\Debug\Debug;
-use Symfony\Component\Dotenv\Dotenv;
 use Symfony\Component\HttpFoundation\Request;
 
-require __DIR__.'/../vendor/autoload.php';
+require dirname(__DIR__).'/vendor/autoload.php';
 
-// The check is to ensure we don't use .env in production
-if (!isset($_SERVER['APP_ENV']) && !isset($_ENV['APP_ENV'])) {
-    if (!class_exists(Dotenv::class)) {
-        throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
-    }
-    (new Dotenv())->load(__DIR__.'/../.env');
-}
+Kernel::bootstrapEnv();
 
-$env = $_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? 'dev';
-$debug = (bool) ($_SERVER['APP_DEBUG'] ?? $_ENV['APP_DEBUG'] ?? ('prod' !== $env));
-
-if ($debug) {
+if ($_SERVER['APP_DEBUG']) {
     umask(0000);
 
     Debug::enable();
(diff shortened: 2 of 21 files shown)


symfony/console/3.3/bin/console
```diff
@@ -3,37 +3,27 @@
...
-    }
-    (new Dotenv())->load(__DIR__.'/../.env');
-}
-
-$input = new ArgvInput();
-$env = $input->getParameterOption(['--env', '-e'], $_SERVER['APP_ENV'] ?? 'dev', true);
-$debug = (bool) ($_SERVER['APP_DEBUG'] ?? ('prod' !== $env)) && !$input->hasParameterOption('--no-debug', true);
-
-if ($debug) {
+if ($_SERVER['APP_DEBUG']) {
```
ogizanagi: Should be moved after `Kernel::bootstrapEnv()`, right?

symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -58,4 +60,71 @@ protected function configureRoutes(RouteCollectionBuilder $routes)
         $routes->import($confDir.'/{routes}/'.$this->environment.'/**/*'.self::CONFIG_EXTS, '/', 'glob');
         $routes->import($confDir.'/{routes}'.self::CONFIG_EXTS, '/', 'glob');
     }
+
+    public static function bootstrapCli(array &$argv)
+    {
+        // consume --env and --no-debug from the command line
+
+        for ($i = 0; $i < \count($argv) && '--' !== $v = $argv[$i]; ++$i) {
```
dunglas: It would be nice if it can be moved in `symfony/symfony` directly, it's the last "tricky" part that is generated.

nicolas-grekas: makes sense! suggested in https://github.com/symfony/symfony/pull/29126#issuecomment-437092754

symfony/console/3.3/bin/console
nicolas-grekas: right, fixed thanks!

symfony/framework-bundle/3.3/public/index.php
```diff
@@ -32,7 +22,7 @@
     Request::setTrustedHosts(explode(',', $trustedHosts));
 }
 
-$kernel = new Kernel($env, $debug);
+$kernel = new Kernel($_SERVER['APP_ENV'], $_SERVER['APP_DEBUG']);
```
Tobion: considering env variables can only be string, it feels wrong to use them in boolean context here. (even if it is casted to boolean in bootstrapEnv().

Tobion: I'm afraid is might teach bad practices when people just copy-paste such things for other variables without thinking about it. That is why the env processors got introduced. But not a big deal to me.

nicolas-grekas: works for me: APP_DEBUG is now normalized to the '1' or '0' string

nicolas-grekas:
I improved the phpunit-bridge integration: it now uses `.env` files *before* anything else.
This means `.env.test` works now to configure the bridge also.
This fixes https://github.com/symfony/symfony/issues/28726

nicolas-grekas:
PR updated to leverage https://github.com/symfony/symfony/pull/29154

symfony/framework-bundle/3.3/src/Kernel.php
nicolas-grekas: and done in https://github.com/symfony/symfony/pull/29154

symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -58,4 +59,78 @@ protected function configureRoutes(RouteCollectionBuilder $routes)
...
+        }
+    }
+
+    public static function bootstrapEnv($env = null)
+    {
+        if (null !== $env) {
+            putenv('APP_ENV='.$_SERVER['APP_ENV'] = $env);
+        }
+
+        if ('prod' !== $_SERVER['APP_ENV'] = isset($_SERVER['APP_ENV']) ? $_SERVER['APP_ENV'] : (isset($_ENV['APP_ENV']) ? $_ENV['APP_ENV'] : null)) {
```
chalasr: undefined `APP_ENV` in dev mode generally means you forgot to create your local `.env` file from the dist one. Now it defaults to `dev` so the app may run while it misses project env vars, and you have to figure it out yourself. Not sure it's better DX wise?

chalasr: hmm forgot about it, it still throws

symfony/console/3.3/bin/console
```diff
@@ -3,37 +3,27 @@
...
     umask(0000);
 
     if (class_exists(Debug::class)) {
         Debug::enable();
     }
 }
 
-$kernel = new Kernel($env, $debug);
+$kernel = new Kernel($_SERVER['APP_ENV'], $_SERVER['APP_DEBUG']);
```
Tobion: You need to cast $_SERVER['APP_DEBUG'] to boolean to not rely on implicit typecast.

bartvanraaij:
It looks like the "env Configurator" is broken since this PR?

I try to set up a new API Platform project via:
```
composer create-project symfony/skeleton test-api
cd test-api
composer req api
```

At this point, the console reads `Modify your DATABASE_URL config in .env`.
However, the .env-file is completely empty (apart from the default comment regarding real env-vars).
Also, the `CORS_ALLOW_ORIGIN` var which is required by NelmioCorsBundle (installed by API Platform) is not added.

If I manually create a `.env.dist` file before doing `composer req api` it does work. But that's a workaround to trigger the "old way" I guess.

Another use case:
```
composer create-project symfony/website-skeleton test-site
```
This also generates an empty .env file.

nicolas-grekas:
(fixed by bumping flex to v1.1.7 in skeletons)

bendavies:
To make others aware, there is a breaking change here **if** your coding standard adds `declare(strict_types=1);` to `index.php` as the cast of `APP_DEBUG` to a bool is no longer done, which will result in:
> PHP Fatal error:  Uncaught TypeError: Argument 2 passed to Symfony\Component\HttpKernel\Kernel::__construct() must be of the type boolean, string given`

So, exclude these file from your cs fixer config.

ciaranmcnulty:
@bendavies or add an explicit `(bool)` cast to your index.php whilst adding the `declare`

bendavies:
well yes, but I'd rather keep inline with upstream so I can just copy and paste updates from recipes in future without having to diff my own changes.

This was a good bug, as our integration tests don't hit `index.php`, so this nice bug made it to a deployed environment before being spotted (testing env, thankfully).

PR #482, opened 2018-11-09
Fix recipe: make consistent with symfony/flex 1.0 new env handling

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

In #481, the symfony/flex 1.0 recipe was changed so that the user gets a `.env` file and NOT a `.env.dist`. The gitignore recipe for FWBundle 4.2 was updated, but not 3.3.

Cheers!

### DIFF
--- a/symfony/flex/1.0/.env.dist
+++ b/symfony/flex/1.0/.env.dist
@@ -1,0 +1,3 @@
+# This file is a "template" of which env vars need to be defined for your application
+# Copy this file to .env file for development, create environment variables when deploying to production
+# https://symfony.com/doc/current/best_practices/configuration.html#infrastructure-related-configuration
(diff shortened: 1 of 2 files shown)


symfony/flex/1.0/.env.dist
```diff
@@ -0,0 +1,3 @@
+# This file is a "template" of which env vars need to be defined for your application
+# Copy this file to .env file for development, create environment variables when deploying to production
+# https://symfony.com/doc/current/best_practices/configuration.html#infrastructure-related-configuration
```
weaverryan: Re-added until symfony/flex#434 is merged. Until then, if you don't have a `.env.dist` file, the "env" configurator won't work - i.e. it won't change your `.env` file when you install things.

nicolas-grekas:
Flex PR is https://github.com/symfony/flex/pull/434

PR #485, opened 2018-11-09
Handle disabling bundles when config set to false

### LINKED ISSUE #383: Register bundles not working as expected (or not well documented)
Copy of https://github.com/symfony/symfony/issues/26526

| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | yes
| Symfony version  | 3.4, 4.0

Im facing currently the problem that the condtion to load a bundle in the kernel method registerBundles not work as the bundles.php maybe intends it.

Example:

```php
// CASE 1: this loads the bundle every time
Symfony\Bundle\DebugBundle\DebugBundle::class => ['all' => true]

// CASE 2: this loads the bundle on dev and test 
Symfony\Bundle\DebugBundle\DebugBundle::class => ['dev' => true, 'test' =>true]

// CASE 3: this should not load for test, but it's loaded, since the isset condition gets false, and false !== null
Symfony\Bundle\DebugBundle\DebugBundle::class => ['dev' => true, 'test' => false, 'prod' => true]

// CASE 3.5: this should load for all, except for prod, but it's loaded since the or conditons is always true
Symfony\Bundle\DebugBundle\DebugBundle::class => ['all' => true, 'prod' => false]
```

In my opinion it should looks like this (i know, this implementation it can throw notices)

```php
foreach ($contents as $class => $envs) {
    if ($envs[$this->environment] === true || ($envs['all'] === true && (!isset($envs[$this->environment]) || $envs[$this->environment] !== false))) {
        yield new $class();
    }
}
```

to "fix" the cases 3. In addition it's also allow blacklist a bundle for specific environments (case 3.5).

Tobion:
Fixes #383

| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/framework-bundle/4.2/src/Kernel.php
+++ b/symfony/framework-bundle/4.2/src/Kernel.php
@@ -31,7 +31,7 @@ class Kernel extends BaseKernel
     {
         $contents = require $this->getProjectDir().'/config/bundles.php';
         foreach ($contents as $class => $envs) {
-            if (isset($envs['all']) || isset($envs[$this->environment])) {
+            if (($envs['all'] ?? false) && ($envs[$this->environment] ?? true) || ($envs[$this->environment] ?? false)) {
                 yield new $class();
             }
         }


symfony/framework-bundle/4.2/src/Kernel.php
```diff
@@ -31,7 +31,7 @@ public function registerBundles()
     {
         $contents = require $this->getProjectDir().'/config/bundles.php';
         foreach ($contents as $class => $envs) {
-            if (isset($envs['all']) || isset($envs[$this->environment])) {
+            if (($envs['all'] ?? false) && ($envs[$this->environment] ?? true) || ($envs[$this->environment] ?? false)) {
```
guilliamxavier: For the record: this was rewritten as `if ($envs[$this->environment] ?? $envs['all'] ?? false) {` and backported to the 3.3 recipe in #491

PR #487, opened 2018-11-10
nelmio/cors-bundle: update CORS_ALLOW_ORIGIN

ro0NL:
| Q             | A
| ------------- | ---
| License       | MIT

fixes

```
.env: line 48: syntax error near unexpected token `('
.env: line 48: `CORS_ALLOW_ORIGIN=^https?://localhost(:[0-9]+)?$'
```

with `source .env` 

after

```
echo $CORS_ALLOW_ORIGIN
^https?://localhost(:[0-9]+)?$
```

let me know if im missing something :thinking:

### DIFF
--- a/nelmio/cors-bundle/1.5/manifest.json
+++ b/nelmio/cors-bundle/1.5/manifest.json
@@ -6,7 +6,7 @@
         "config/": "%CONFIG_DIR%/"
     },
     "env": {
-        "CORS_ALLOW_ORIGIN": "^https?://localhost(:[0-9]+)?$"
+        "CORS_ALLOW_ORIGIN": "^https?://localhost\(:[0-9]+\)?$"
     },
     "aliases": ["cors"]
 }
--- a/nelmio/cors-bundle/1.5/manifest.json
+++ b/nelmio/cors-bundle/1.5/manifest.json
@@ -6,7 +6,7 @@
         "config/": "%CONFIG_DIR%/"
     },
     "env": {
-        "CORS_ALLOW_ORIGIN": "^https?://localhost\\\(:[0-9]+\\\)?$"
+        "CORS_ALLOW_ORIGIN": "^https?://localhost\\\\(:[0-9]+\\\\)?$"
     },
     "aliases": ["cors"]
 }
--- a/nelmio/cors-bundle/1.5/manifest.json
+++ b/nelmio/cors-bundle/1.5/manifest.json
@@ -6,7 +6,7 @@
         "config/": "%CONFIG_DIR%/"
     },
     "env": {
-        "CORS_ALLOW_ORIGIN": "^https?://localhost\\(:[0-9]+\\)?$"
+        "CORS_ALLOW_ORIGIN": "^https?://localhost\\\(:[0-9]+\\\)?$"
     },
     "aliases": ["cors"]
 }
--- a/nelmio/cors-bundle/1.5/manifest.json
+++ b/nelmio/cors-bundle/1.5/manifest.json
@@ -6,7 +6,7 @@
         "config/": "%CONFIG_DIR%/"
     },
     "env": {
-        "CORS_ALLOW_ORIGIN": "^https?://localhost\(:[0-9]+\)?$"
+        "CORS_ALLOW_ORIGIN": "^https?://localhost\\(:[0-9]+\\)?$"
     },
     "aliases": ["cors"]
 }
--- a/nelmio/cors-bundle/1.5/manifest.json
+++ b/nelmio/cors-bundle/1.5/manifest.json
@@ -6,7 +6,7 @@
         "config/": "%CONFIG_DIR%/"
     },
     "env": {
-        "CORS_ALLOW_ORIGIN": "^https?://localhost\\\\(:[0-9]+\\\\)?$"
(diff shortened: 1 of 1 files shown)


ro0NL:
see https://3v4l.org/rP8Cj

```
$ FOO=^https?://localhost\(:[0-9]+\)?$
$ echo $FOO
^https?://localhost(:[0-9]+)?$
```

ro0NL:
but it's not generated like that, so default code is invalid currently.

ro0NL:
wait, we can put the single quotes in the recipe yes. Not sure we prefer one or another actually..

ro0NL:
works :ok_hand:

ro0NL:
Hm, not sure actually. Im using https://github.com/lando/lando and i  see it generates the following

```
version: '3.2'
services:
  appserver:
    image: 'devwithlando/php:7.1-fpm'
    environment:
      TERM: xterm
      COMPOSER_ALLOW_SUPERUSER: 1
      PATH: >-
        /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/var/www/.composer/vendor/bin
      LANDO_WEBROOT: /app/public
      LANDO_SERVICE_NAME: appserver
      LANDO_SERVICE_TYPE: php
      LANDO_MOUNT: /app
      COLUMNS: 256
      LANDO: 'ON'
      LANDO_CONFIG_DIR: $LANDO_ENGINE_CONF
      LANDO_DOMAIN: lndo.site
      LANDO_APP_NAME: myproject
      LANDO_APP_ROOT: /home/roland/dev/my-project
      LANDO_APP_ROOT_BIND: /home/roland/dev/my-project
      LANDO_HOST_OS: linux
      LANDO_HOST_UID: '1001'
      LANDO_HOST_GID: '1001'
      LANDO_HOST_IP: 192.168.178.18
      LANDO_WEBROOT_USER: www-data
      LANDO_WEBROOT_GROUP: www-data
      LANDO_WEBROOT_UID: '33'
      LANDO_WEBROOT_GID: '33'
      APP_ENV: dev
      APP_SECRET: c017ebf223052640d2eb426b8cb0d35c
      CORS_ALLOW_ORIGIN: '^https?://localhost(:[0-9]+)?$'
```

Starting it already screams:

```
ERROR: Invalid interpolation format for "environment" option in service "appserver": "^https?://localhost(:[0-9]+)?$"
```

But other values are also wrapped in single quotes :thinking: 

`echo $CORS_ALLOW_ORIGIN` in the container gives an empty string now :(

`source .env` in the container works as expected though,

With single quotes it seems the `$` needs double escaping: `$$`.

nicolas-grekas:
There's only one valid and supported way to parse a .env file, it's with a shell - ie `source .env`.
If other tools didn't implement proper parsing rules, that's their issue IMHO.

ro0NL:
Tend to agree.. not sure if this is tied to Docker or Lando though.

So @dunglas can you confirm it still works for you? I can only say sourcing the current value doesnt.

ro0NL:
```bash
$ docker run -it --rm --env-file .env bash:4.4
bash-4.4# echo $CORS_ALLOW_ORIGIN
'^h2ttps?://localhost(:[0-9]+)?$'
```

That's wrong at least.

nicolas-grekas:
Does `\` escaping work better?

ro0NL:
Not really :(

```bash
$ source .env
$ echo $CORS_ALLOW_ORIGIN
^https?://localhost(:[0-9]+)?$
$ docker run -it --rm --env-file .env bash:4.4
bash-4.4# echo $CORS_ALLOW_ORIGIN
^https?://localhost\(:[0-9]+\)?$
```

ro0NL:
Forwarding env does work :man_shrugging: 

```bash
export CORS_ALLOW_ORIGIN
$ echo $CORS_ALLOW_ORIGIN
^https?://localhost(:[0-9]+)?$
$ docker run -it --rm --env CORS_ALLOW_ORIGIN bash:4.4
bash-4.4# echo $CORS_ALLOW_ORIGIN
^https?://localhost(:[0-9]+)?$
```

nicolas-grekas:
Sure: the issue is the .env parser they use. Once the env var is set, it is not parsed anymore.

ro0NL:
Think so, per https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file

> to set simple (non-array) environment variables

I think "simple" is key here...

ro0NL:
To move forward.. big :+1: for being compatible with sourcing manually.

We can keep the default in the recipe "simple" maybe, to serve both users.

More structurally, perhaps clarify this with a comment on top in the generated env files.

nicolas-grekas:
I'm wondering: how do rails' and nodes' dotenv parse .env files? Do they all parse shell-like or are we the only ones?

nicolas-grekas:
It looks like both nodejs and rails supports single and double quoted strings:
- https://github.com/motdotla/dotenv#rules
- https://github.com/bkeepers/dotenv
- docker-compose is incompatible and there's nothing we can do about it: https://docs.docker.com/compose/env-file/

ro0NL:
i've concluded sharing .env between different concepts (Symfony and Docker) is a bad practice, or impractical at least.

For anyone coming here from google&co, i advise to move docker-compose out of the project root to a dedicated scoped devops/docker folder.

Closing as such.

PR #488, opened 2018-11-11
Load .env.dist when it exists and .env is not found

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Matches https://github.com/symfony/symfony/pull/29171

### DIFF
--- a/phpunit/phpunit/4.7/phpunit.xml.dist
+++ b/phpunit/phpunit/4.7/phpunit.xml.dist
@@ -2,7 +2,7 @@
 
 <!-- https://phpunit.de/manual/current/en/appendixes.configuration.html -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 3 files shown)


phpunit/phpunit/4.7/phpunit.xml.dist
```diff
@@ -2,7 +2,7 @@
 
 <!-- https://phpunit.de/manual/current/en/appendixes.configuration.html -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.7/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.5/phpunit.xsd"
```
nicolas-grekas: version bump related to discussion in https://github.com/symfony/symfony/pull/29174
tl;dr: on the very latest phpunit version validates these, so bumping doesn't anything for older versions, and latest version need recent xsd to pass validation.

PR #490, opened 2018-11-12
Fix loading the autoloader in tests

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

As reported in https://github.com/symfony/recipes/commit/95bb54353eb404a22e365fef62e937a371833af9#r31264986
phpunit uses `include_once` so referencing `vendor/autoload.php` is a no-op.

Let's forget about the potential autoload stack reordering issue, it shouldn't have any effect in practice.


nicolas-grekas:
Closed in favor of https://github.com/symfony/recipes/pull/491

PR #491, opened 2018-11-12
Move loading .env files to bootstrap.php

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

In #481, we completely forgot about apps that *migrate* to flex.
#481 relies on the Kernel being created at the same time as eg `bin/console`; but apps migrating to flex already have a Kernel, and it won't be updated automatically.

Here is a radically simpler approach: let's factorize loading `.env` files in a new `bootstrap.php` file that is used as instead of `vendor/autoload.php` to bootstrap the app.


nicolas-grekas:
Actually, neither skeletons not migrating apps should be upgrade to list config/bootstrap.php: Flex should to it when installing the frameworkbundle recipe!

behat/symfony2-extension/2.1/features/bootstrap/bootstrap.php
```diff
@@ -1,3 +0,0 @@
-<?php
-
-App\Kernel::bootstrapEnv('test');
```
nicolas-grekas: Should be restored to load the "test" env config.

symfony/framework-bundle/3.3/config/bootstrap.php
```diff
@@ -0,0 +1,48 @@
...
+
+use Symfony\Component\Dotenv\Dotenv;
+
+if (!isset($_SERVER['APP_ENV'])) {
+    $_SERVER['APP_ENV'] = isset($_ENV['APP_ENV']) ? $_ENV['APP_ENV'] : null;
+}
+
+if ('prod' !== $_SERVER['APP_ENV']) {
+    if (!class_exists(Dotenv::class)) {
+        throw new \RuntimeException('The "APP_ENV" environment variable is not defined. You need to set it or run "composer require symfony/dotenv" to load it from a ".env" file.');
```
nicolas-grekas: `\` can be removed

symfony/framework-bundle/3.3/config/bootstrap.php
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    $path = \dirname(__DIR__).'/.env';
```
nicolas-grekas: `\`

symfony/framework-bundle/3.3/src/Kernel.php
```diff
@@ -59,82 +59,4 @@ protected function configureRoutes(RouteCollectionBuilder $routes)
...
-    }
-
-    public static function bootstrapEnv($env = null)
-    {
-        if (null !== $env) {
-            putenv('APP_ENV='.$_SERVER['APP_ENV'] = $env);
-        }
-
-        if ('prod' !== $_SERVER['APP_ENV'] = isset($_SERVER['APP_ENV']) ? $_SERVER['APP_ENV'] : (isset($_ENV['APP_ENV']) ? $_ENV['APP_ENV'] : null)) {
-            if (!class_exists(Dotenv::class)) {
```
nicolas-grekas: Dotenv to be removed from use statements

symfony/framework-bundle/4.2/config/bootstrap.php
```diff
@@ -0,0 +1,21 @@
...
+
+use Symfony\Component\Dotenv\Dotenv;
+
+if (!isset($_SERVER['APP_ENV'])) {
+    $_SERVER['APP_ENV'] = $_ENV['APP_ENV'] ?? null;
+}
+
+if ('prod' !== $_SERVER['APP_ENV']) {
+    if (!class_exists(Dotenv::class)) {
+        throw new \RuntimeException('The "APP_ENV" environment variable is not defined. You need to set it or run "composer require symfony/dotenv" to load it from a ".env" file.');
```
nicolas-grekas: `\`

symfony/framework-bundle/4.2/config/bootstrap.php
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    (new Dotenv())->loadEnv(\dirname(__DIR__).'/.env');
```
nicolas-grekas: `\`

symfony/framework-bundle/4.2/src/Kernel.php
```diff
@@ -57,29 +56,4 @@ protected function configureRoutes(RouteCollectionBuilder $routes)
         $routes->import($confDir.'/{routes}/'.$this->environment.'/**/*'.self::CONFIG_EXTS, '/', 'glob');
         $routes->import($confDir.'/{routes}'.self::CONFIG_EXTS, '/', 'glob');
     }
-
-    public static function bootstrapCli(array &$argv)
-    {
-        // consume --env and --no-debug from the command line
-        Application::bootstrapEnv($argv);
```
nicolas-grekas: To be removed from use statements.

nicolas-grekas:
PR tested locally, works like a charm. The critical feature this PR provides is decoupling recipes. As explained in the description, the current recipes are broken because they make too many assumptions.

Note that this PR introduced a new `bootstrap.php` file that should be used to setup both the autoload and load `.env` files.
In order to reduce duplication while still providing decoupling of recipes, this PR uses symbolic links to ship the `bootstrap.php` in several recipes.
Works transparently with the flex server.

PR ready. Votes pending.

symfony/framework-bundle/4.2/src/Kernel.php
```diff
@@ -31,7 +29,7 @@ public function registerBundles()
     {
         $contents = require $this->getProjectDir().'/config/bundles.php';
         foreach ($contents as $class => $envs) {
-            if (($envs['all'] ?? false) && ($envs[$this->environment] ?? true) || ($envs[$this->environment] ?? false)) {
+            if (isset($envs['all']) || isset($envs[$this->environment])) {
```
Pierstoval: Why did you revert this change?

nicolas-grekas: to make the code the same in the v3.3 vs v4.2 recipes - and because the code with isset is more readable.

B-Galati: But wasn't it fixing an actual bug? 

On my end, I don't even understand the version with `??`, too tricky for me, so big :+1: for readability.

nicolas-grekas: if it was a fix, it as not applied to v3.3 :)

B-Galati: Of course, the issue was #383

nicolas-grekas: OK, fixed

B-Galati: That's more readable than the previous solution :+1: Thx!

Pierstoval: Yep, new solution is much better 👍

symfony/framework-bundle/3.3/bootstrap.php
```diff
@@ -0,0 +1,48 @@
...
+
+require __DIR__.'/vendor/autoload.php';
+
+if (!isset($_SERVER['APP_ENV'])) {
+    $_SERVER['APP_ENV'] = isset($_ENV['APP_ENV']) ? $_ENV['APP_ENV'] : null;
+}
+
+if ('prod' !== $_SERVER['APP_ENV']) {
+    if (!class_exists(Dotenv::class)) {
+        throw new RuntimeException('The "APP_ENV" environment variable is set not to "prod". Please run "composer require symfony/dotenv" to load the ".env" files configuring the application.');
```
ogizanagi:
```diff
- The "APP_ENV" environment variable is set not to "prod"
+ The "APP_ENV" environment variable is not set to "prod"

Comment (ogizanagi) [reply symfony/framework-bundle/3.3/bootstrap.php]: Previously, it was: `'The "APP_ENV" environment variable is not defined. You need to set it or run "composer require symfony/dotenv" to load it from a ".env" file.'`

Now, we're always requiring the Dotenv component in non prod env. Is that expected?

Comment (nicolas-grekas) [reply symfony/framework-bundle/3.3/bootstrap.php]: fixed thanks. about non-prod envs: yes to me, that's conscious change in previous PR.

File: symfony/framework-bundle/3.3/bootstrap.php [hunk 10]
```diff
@@ -0,0 +1,48 @@
...
+        throw new RuntimeException('The "APP_ENV" environment variable is not set to "prod". Please run "composer require symfony/dotenv" to load the ".env" files configuring the application.');
+    }
+
+    $path = __DIR__.'/.env';
+
+    if (method_exists(Dotenv::class, 'loadEnv')) {
+        (new Dotenv())->loadEnv($path);
+    } else {
+        if (file_exists($path) || !file_exists($p = "$path.dist")) {
+            $dotenv->load($path);
```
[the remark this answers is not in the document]
nicolas-grekas: fixed thanks

symfony/framework-bundle/3.3/bootstrap.php
```diff
@@ -0,0 +1,51 @@
...
+
+use Symfony\Component\Dotenv\Dotenv;
+
+require __DIR__.'/vendor/autoload.php';
+
+if (!isset($_SERVER['APP_ENV'])) {
+    $_SERVER['APP_ENV'] = isset($_ENV['APP_ENV']) ? $_ENV['APP_ENV'] : null;
+}
+
+if ('prod' !== $_SERVER['APP_ENV']) {
```
weaverryan: There's an edge-case here. With no "real" env vars set, if I run

```php
php bin/console --env=prod some:command
```

None of the `.env `files (including `.env.prod.local`) are loaded. The intention of this line I think is "to avoid loading the `.env` files if real env vars are set". But passing `--env=prod` isn't a real env var. It's annoying, but should we cover that case - *yes* load the `.env` files if `--env=prod` is passed via the command line?

symfony/phpunit-bridge/4.1/manifest.json
```diff
@@ -2,6 +2,7 @@
     "copy-from-recipe": {
         ".env.test": ".env.test",
```
weaverryan: It's not part of this PR, but the `.env.test` file contains:

```
KERNEL_CLASS=App\\Kernel
```

I believe we need a single slash - tests fail currently with `Class "App\\Kernel" doesn't exist`

symfony/framework-bundle/3.3/bootstrap.php
nicolas-grekas: works for me, that's how [it used to work actually](https://github.com/symfony/recipes/pull/481/files#diff-2af50efd729ff8e61dcbd936cf2b114b).
updated

symfony/phpunit-bridge/4.1/manifest.json
nicolas-grekas: that's a BUG of Dotenv: this should work! (`source .env.test` to prove it)
worked around by using single quotes.

nicolas-grekas: that's a BUG of Dotenv: this should work! (`source .env.test` to prove it)
worked around by using single quotes.

phpunit/phpunit/4.7/phpunit.xml.dist
```diff
@@ -5,10 +5,11 @@
          xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.5/phpunit.xsd"
          backupGlobals="false"
          colors="true"
-         bootstrap="tests/bootstrap.php"
+         bootstrap="bootstrap.php"
 >
     <php>
         <ini name="error_reporting" value="-1" />
+        <env name="APP_ENV" value="test" />
```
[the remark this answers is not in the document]
nicolas-grekas: > APP_ENV is already set in bin/phpunit

APP_ENV is *explicitly* set to "test" in `bin/phpunit` - so that `.env.test` is loaded in both cases.
*Not* adding this line is what would create the inconsistency, by not loading `.env.test` when using `vendor/bin/simple-phpunit` (or `vendor/bin/phpunit` btw if installed.)

Or did I miss your case?

nicolas-grekas:
> With these changes, it's now possible to start the built-in web server in the test environment, which is great for Panther:

Not anymore, because you asked something conflicting in your first comment above.

symfony/framework-bundle/3.3/bootstrap.php
weaverryan: It almost works now. When I run `php bin/console --env=prod some:command`, it DOES load `.env`, but it does not load `.env.prod.local` because it thinks (at that moment) that the `$env` is still `dev`.

phpunit/phpunit/4.7/phpunit.xml.dist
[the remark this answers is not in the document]
nicolas-grekas: OK, fixed by *not* loading .env files in bin/phpunit, so that only phpunit.xml drives them.

symfony/framework-bundle/3.3/bootstrap.php
nicolas-grekas: Fixed

symfony/console/3.3/bin/console
```diff
@@ -26,4 +37,4 @@ if ($_SERVER['APP_DEBUG']) {
 
 $kernel = new Kernel($_SERVER['APP_ENV'], $_SERVER['APP_DEBUG']);
```
weaverryan: Edge case: if you use `--env=prod` (or any environment), when you get to this spot, `$_SERVER['APP_ENV']` is always `dev`. So, the `bin/console` kernel always loads in dev.

nicolas-grekas: fixed!

weaverryan:
This works great! I've tested many different situations:

* A) the `.env` `.env.local` and `.env.{env}.local` files all load correctly with DotEnv less than 4.2
* B) PHPUnit loads ENV vars from the `.env` and `.env.test` files correctly, regardless of running `bin/phpunit` or `vendor/bin/phpunit`
* C) env vars in `phpunit.xml.dist` can still be used if you want, and they override values from the `.env` files
* D) Passing the `--env` flag works correctly: the Kernel for the Application is booted into the correct environment and all the proper `.env` files are loaded, even if you use `--env=prod`.

Big +1 from me!

symfony/framework-bundle/4.2/src/Kernel.php
```diff
@@ -31,7 +29,7 @@ public function registerBundles()
     {
         $contents = require $this->getProjectDir().'/config/bundles.php';
         foreach ($contents as $class => $envs) {
-            if (($envs['all'] ?? false) && ($envs[$this->environment] ?? true) || ($envs[$this->environment] ?? false)) {
+            if (!empty($envs['all']) || !empty($envs[$this->environment])) {
```
Tobion: this does not do the same as before for all => true, test => false

nicolas-grekas: thanks, fixed by using `if ($envs[$this->environment] ?? $envs['all'] ?? false) {`

nicolas-grekas:
https://github.com/symfony/symfony/pull/29204 now being merged, PR is ready to go :)

symfony/flex/1.0/.env
```diff
@@ -1,4 +1,5 @@
 # This file defines all environment variables that the application needs.
+# DO NOT PUT PRODUCTION SECRETS IN THIS FILE.
```
fabpot: DO NOT DEFINE?

symfony/framework-bundle/3.3/bootstrap.php
```diff
@@ -0,0 +1,51 @@
+<?php
```
fabpot: Why not store the `bootstrap.php` file in the `src/` directory instead?

nicolas-grekas: because src/ is configurable, so it cannot be safely referenced from other recipes.

symfony/phpunit-bridge/3.3/.env.test
```diff
@@ -1,6 +1,4 @@
 # define your env variables for the test env here
-KERNEL_CLASS=App\\Kernel
+KERNEL_CLASS='App\Kernel'
 APP_SECRET='s$cretf0rt3st'
 SHELL_VERBOSITY=-1
-SYMFONY_DEPRECATIONS_HELPER=999999
```
Tobion: Why not keep this here? This would solve the problem, that bin/phpunit sets the SYMFONY_DEPRECATIONS_HELPER variable while vendor/bin/phpunit does not set it. This was my major pain point in https://github.com/symfony/recipes/pull/442#issuecomment-407994227

symfony/console/3.3/bin/console
```diff
@@ -3,18 +3,29 @@
...
+    throw new RuntimeException('You need to add "symfony/framework-bundle" as a Composer dependency.');
 }
 
-Kernel::bootstrapCli($_SERVER['argv']);
-Kernel::bootstrapEnv();
+$input = new ArgvInput();
+if (null !== $_ENV['APP_ENV'] = $input->getParameterOption(['--env', '-e'], null, true)) {
+    putenv('APP_ENV='.$_ENV['APP_ENV']);
+    // force loading .env files when --env is defined
+    $_SERVER['APP_ENV'] = null;
```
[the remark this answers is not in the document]
nicolas-grekas: See previous comment from @weaverryan: we want .env files to be loaded also when --env is used, even for prod.
There is only one way to opt-out from .env files, which is to set a real APP_ENV var to prod.

symfony/phpunit-bridge/3.3/.env.test
nicolas-grekas: I'm not sure how that's related: defining this var in phpunit.xml.dist does work AFAIK, *except* from the "disabled" value (which has to be set in `bin/phpunit`. The reason is another comment from @weaverryan: loading earlier creates an inconsistency in terms of precedence we'd prefer avoiding.

symfony/console/3.3/bin/console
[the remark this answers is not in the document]
Tobion: I don't see the use-case for this to add such an inconsistency. People will run bin/console --env=prod locally which might work. Then they run exactly the same command when deployed, it it might not work because it tries to load env files but the DotEnv is not installed. Thus they get an exception.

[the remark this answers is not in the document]
nicolas-grekas: that's the opposite: Dotenv is *not* loaded in prod, when APP_ENV is set.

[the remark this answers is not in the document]
Tobion: People that do not define a real APP_ENV but run commands using --env=prod (like sf 3 style) will think it works because it works locally. But it will not work when deployed without the DotEnv.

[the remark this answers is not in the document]
nicolas-grekas: https://github.com/symfony/skeleton/pull/122 should address the concern.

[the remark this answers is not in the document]
nicolas-grekas: if people defined APP_ENV on their prod, they opt-ed in for using real env vars.
Then it's their job to make the process robust.

symfony/phpunit-bridge/3.3/.env.test
Tobion: I don't see how ryans comment disagrees with my proposal. Keep this line here instead of bin/phpunit.

This way `bin/phpunit` and `vendor/bin/phpunit` behaves the same:
- both set the APP_ENV to test
- both load the boostrap file
- both load .env.test file
- both behave the same in regard to SYMFONY_DEPRECATIONS_HELPER becaues it's part of .env.test

symfony/console/3.3/bin/console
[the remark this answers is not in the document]
Tobion: So there are two camps:
1. people not using real env vars who want to use DotEnv also for prod
    - they just need to make DotEnv a real dependency
    - and remove `if ('prod' !== $_SERVER['APP_ENV']) {` from bootstrap
2. people using real env vars
    - `if ('prod' !== $_SERVER['APP_ENV']) {` and `require-dev=DotEnv` secures them against 
      - forgetting to define env vars and accidentally loading .env without realizing it
      - defining APP_ENV but some dependency moved DotEnv to require without realizing it, and thus loading .env params accidentally

symfony/skeleton#122 will break the safeguard for 2)
And `$_SERVER['APP_ENV'] = null;` makes the bahavior inconsistent for 1)

symfony/phpunit-bridge/3.3/.env.test
nicolas-grekas: OK, got it, updated.

symfony/console/3.3/bin/console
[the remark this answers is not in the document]
nicolas-grekas: I'm not sure I understand what you're proposing to change?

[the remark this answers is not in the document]
Tobion: What I'm proposing is not do symfony/skeleton#122 and remove `$_SERVER['APP_ENV'] = null;` so that `APP_ENV=prod bin/console` and `bin/console --env=prod` behaves the same.

>  A fatal error is no better - that didn't save anything really. Only proper deployment process can fail proof a prod. 

Any decent deployment will test if the app runs correctly before switching it live. So yes, throwing an exception when env vars are missing is exactly part of the deployment process you describe. If you remove that behavior, only complicated functional tests can cover loading the wrong environment.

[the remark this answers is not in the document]
weaverryan: > remove $_SERVER['APP_ENV'] = null; so that APP_ENV=prod bin/console and bin/console --env=prod behaves the same.

I disagree with this part - this would make things more confusing. Passing the env flag changes the environment only, and then all the normal rules should be followed. Loading the env vars for -env=dev but suddenly not for -env=prod is weird behavior.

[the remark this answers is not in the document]
Tobion: To me --env=... is just an alias for APP_ENV=... 
This is what people wanted in https://github.com/symfony/symfony/issues/28984 because it's sometimes hard to set env variables. If those two things are not doing the same thing, it's confusing and can cause unexpected behavior in my opinions. But I guess it depends on the perspective.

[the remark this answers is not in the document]
nicolas-grekas: When first submitting the PR it behaved as you describe Tobion, but now I understand the rationale of what Ryan's perspective and it makes sense to me.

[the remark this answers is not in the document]
nicolas-grekas: In case one is wondering, Laravel loads .env.prod when --env=prod is used ;)

symfony/framework-bundle/4.2/public/index.php
```diff
@@ -4,9 +4,7 @@
 use Symfony\Component\Debug\Debug;
 use Symfony\Component\HttpFoundation\Request;
 
-require dirname(__DIR__).'/vendor/autoload.php';
-
-Kernel::bootstrapEnv();
+require dirname(__DIR__).'/src/.bootstrap.php';
 
 if ($_SERVER['APP_DEBUG']) {
```
Tobion: How about moving this block to the bootstrap file as it is currently duplicated in public/index.php and bin/console

Tobion: This would also enable the debug stuff for phpunit when enabled. Is that a problem?

symfony/console/3.3/bin/console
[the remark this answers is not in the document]
Tobion: I still don't see the use-case. If I use real env vars, why would I use .env.prod at all? If I don't use real env, loading .env.prod for--env=prod is not gonna work when the app is deployed by default.
But in the end, as long as people understand the differences between --env and APP_ENV, this is not a problem but more a doc issue.

symfony/framework-bundle/4.2/public/index.php
nicolas-grekas: Yes, that'd be a problem, so we cannot do that (I did it in an earlier version and reverted for this reason.)

bendavies:
for anyone running paratest, this will break your tests.
`src/.bootstrap.php` is loaded by paratest before `<env name="APP_ENV" value="test" />` is read in `phpunit.xml.dist`, so the wrong `.env` files are read.

Still looking for a solution.

nicolas-grekas:
You need to define `APP_ENV=test` before running paratest.
Dunno how this can be achieved when using paratest.
A different bootstrap file would do it of course:
```php
// tests/bootstrap.php
putenv('APP_ENV='.$_SERVER['APP_ENV'] = $_ENV['APP_ENV'] = 'test');
require dirname(__DIR__).'/src/.bootstrap.php');
```

bendavies:
Yep - lovely, thankyou. Have submitted an issue to paratest.

bendavies:
@nicolas-grekas is there a bug in `bin/console` here?

These two should be equivalent, shouldn't they?

`APP_ENV=prod APP_DEBUG=0 bin/console cache:clear --no-warmup`
`bin/console cache:clear --no-warmup --env=prod --no-debug`

The 2nd will result in

```
Fatal error: Uncaught RuntimeException: The "APP_ENV" environment variable is not set to "prod". Please run "composer require symfony/dotenv" to load the ".env" files configuring the application. in project/src/.bootstrap.php on line 13
```

PR #492, opened 2018-11-13
Nelmio Alice: blacklist PHP native functions that match formatters

jmsche:
| Q             | A
| ------------- | ---
| License       | MIT

Hi, I added a few PHP native functions to blacklist, as they should be overridden by fzaninotto/faker.


jmsche:
CI Failure should not block this PR as bundles are included in the package, but can't be detected

jmsche:
Thanks @stof :)

PR #494, opened 2018-11-17
TRUSTED_HOSTS is a regexp

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/framework-bundle/3.3/manifest.json
+++ b/symfony/framework-bundle/3.3/manifest.json
@@ -15,7 +15,7 @@
         "APP_ENV": "dev",
         "APP_SECRET": "%generate(secret)%",
         "#TRUSTED_PROXIES": "127.0.0.1,127.0.0.2",
-        "#TRUSTED_HOSTS": "localhost,example.com"
+        "#TRUSTED_HOSTS": "'^localhost|example\\.com$'"
     },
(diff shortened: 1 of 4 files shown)


symfony/framework-bundle/3.3/manifest.json
```diff
@@ -15,7 +15,7 @@
         "APP_ENV": "dev",
         "APP_SECRET": "%generate(secret)%",
         "#TRUSTED_PROXIES": "127.0.0.1,127.0.0.2",
-        "#TRUSTED_HOSTS": "localhost,example.com"
+        "#TRUSTED_HOSTS": "'^localhost|example\\.com$'"
```
guilliamxavier: Isn't it missing braces i.e. `"'^(localhost|example\\.com)$'"` (or alternatively `"'^localhost$|^example\\.com$'"`)?

nicolas-grekas: please try locally and send a PR if this doesn't work as expected

guilliamxavier: As is it would match `localhost` and `example.com` but also e.g. `localhostfoo` and `fooexample.com`. Proposed #728 as a fix

PR #499, opened 2018-11-22
Move src/.bootstrap.php to config/bootstrap.php

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

I feel like this might be more natural. And we already have bundles.php there :)


weaverryan:
I like this, but I dislike changes :). Let’s do this, but quickly

fabpot:
Actually, is really worth it?

nicolas-grekas:
The change is still cheap now, but won't be soon. Better do it and free us from explaining why this file (that's how this PR started :) )

PR #501, opened 2018-12-03
Always load vars defined in .env files, providing .env.local.php for prod

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

This PR provides a nice experience where `.env` files are "always" loaded.
That means their purpose is easy to reason about: the committed files define the defaults (`.env` especially) - and `.env.local` defines local overrides.

A new `.env.local.php` is provided for prod, so that performance is not affected.
When `.env.local.php` is found, other .env files are ignored.

On the deployement side, this means it should be much easier to deal with `.env` files for people that can't easilly deal with real env vars:
- if you use real env vars they always win
- logically, `.env` files are always loaded (to define defaults)
- on prod, one can dump a `.env.local.php` file for max performance

The plan is to add a new flex command to create this file and ease deployments:
`composer dump-env prod`

### DIFF
--- a/symfony/flex/1.0/.env
+++ b/symfony/flex/1.0/.env
@@ -1,5 +1,14 @@
-# This file defines all environment variables that the application needs.
-# DO NOT DEFINE PRODUCTION SECRETS IN THIS FILE.
-# Use ".env.local" for local overrides during development.
-# Use real environment variables when deploying to production.
+# In all environments, the following files are loaded if they exist,
+# the later taking precedence over the former:
+#
+#  * .env                contains default values for the environment variables needed by the app
+#  * .env.local          uncommitted file with local overrides
+#  * .env.$APP_ENV       committed environment-specific defaults
+#  * .env.$APP_ENV.local uncommitted environment-specific overrides
+#
+# Real environment variables win over .env files.
+#
+# DO NOT DEFINE PRODUCTION SECRETS IN THIS FILE NOR IN ANY OTHER COMMITTED FILES.
+#
+# Run "composer dump-env prod" to compile .env files for production use (requires symfony/flex >=1.2).
 # https://symfony.com/doc/current/best_practices/configuration.html#infrastructure-related-configuration
--- a/symfony/framework-bundle/3.3/config/bootstrap.php
+++ b/symfony/framework-bundle/3.3/config/bootstrap.php
@@ -4,18 +4,18 @@ use Symfony\Component\Dotenv\Dotenv;
 
 require dirname(__DIR__).'/vendor/autoload.php';
 
-if (!array_key_exists('APP_ENV', $_SERVER)) {
-    $_SERVER['APP_ENV'] = $_ENV['APP_ENV'] ?? null;
-}
-
-if ('prod' !== $_SERVER['APP_ENV']) {
-    if (!class_exists(Dotenv::class)) {
-        throw new RuntimeException('The "APP_ENV" environment variable is not set to "prod". Please run "composer require symfony/dotenv" to load the ".env" files configuring the application.');
-    }
-
+// Load cached env vars if the .env.local.php file exists
+// Run "composer dump-env prod" to create it (requires symfony/flex >=1.2)
+if (is_array($env = @include dirname(__DIR__).'/.env.local.php')) {
+    $_SERVER += $env;
+    $_ENV += $env;
+} elseif (!class_exists(Dotenv::class)) {
+    throw new RuntimeException('Please run "composer require symfony/dotenv" to load the ".env" files configuring the application.');
+} else {
     $path = dirname(__DIR__).'/.env';
     $dotenv = new Dotenv();
 
+    // load all the .env files
     if (method_exists($dotenv, 'loadEnv')) {
         $dotenv->loadEnv($path);
     } else {
--- a/symfony/framework-bundle/4.2/config/bootstrap.php
+++ b/symfony/framework-bundle/4.2/config/bootstrap.php
@@ -4,15 +4,15 @@ use Symfony\Component\Dotenv\Dotenv;
 
 require dirname(__DIR__).'/vendor/autoload.php';
 
-if (!array_key_exists('APP_ENV', $_SERVER)) {
-    $_SERVER['APP_ENV'] = $_ENV['APP_ENV'] ?? null;
-}
-
-if ('prod' !== $_SERVER['APP_ENV']) {
-    if (!class_exists(Dotenv::class)) {
-        throw new RuntimeException('The "APP_ENV" environment variable is not set to "prod". Please run "composer require symfony/dotenv" to load the ".env" files configuring the application.');
-    }
-
+// Load cached env vars if the .env.local.php file exists
+// Run "composer dump-env prod" to create it (requires symfony/flex >=1.2)
+if (is_array($env = @include dirname(__DIR__).'/.env.local.php')) {
+    $_SERVER += $env;
+    $_ENV += $env;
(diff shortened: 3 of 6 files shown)


symfony/framework-bundle/3.3/config/bootstrap.php
```diff
@@ -4,13 +4,20 @@
 
 require dirname(__DIR__).'/vendor/autoload.php';
 
-if (!array_key_exists('APP_ENV', $_SERVER)) {
-    $_SERVER['APP_ENV'] = $_ENV['APP_ENV'] ?? null;
+if (is_array($env = @include dirname(__DIR__).'/.env.local.php')) {
+    if ('prod' === ($_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? null)) {
```
nicolas-grekas: `?? $env['APP_ENV']`

nicolas-grekas:
@symfony/deciders votes welcome.

This PR provides a nice experience where `.env` files are "always" loaded.
That means their purpose is easy to reason about: the committed files define the defaults (`.env` especially) - and `.env.local` defines local overrides.

Technically, a new `.env.local.php` is loaded *for the prod env only*, so that performance is affected.
When `.env.local.php` is missing, it is created automatically (will happen eg when `bin/console cache:clear --env=prod` runs.)

On the deployement side, this means it should be much easier to deal with `.env` files for people that can't easilly deal with real env vars:
- if you use real envs vars they always win (which you should - see https://12factor.net/)
- logically, .env files are always loaded (to define defaults)
- technically, one can also dump a `.env.local.php` file in their deployement script if they want
- and when this file is missing, it is auto-created from `.env` files, so that defining `.env.local` on prod can be enough to define "env vars" for simple host needs.

symfony/framework-bundle/4.2/config/bootstrap.php
```diff
@@ -4,16 +4,21 @@
...
+    $_SERVER += $env;
+    $_ENV += $env;
+} elseif (!class_exists(Dotenv::class)) {
+    throw new RuntimeException('Please run "composer require symfony/dotenv" to load the ".env" files configuring the application.');
+} else {
     (new Dotenv())->loadEnv(dirname(__DIR__).'/.env');
+
+    if ('prod' === ($_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? null)) {
+        $env = array_intersect_key($_SERVER, array_flip(explode(',', $_SERVER['SYMFONY_DOTENV_VARS'])));
+        $env = "<?php // This file has been generated by config/bootstrap.php\n\nreturn ".var_export($env, true).";\n";
```
weaverryan: > This file was generated by config/bootstrap.php - update your .env files and re-build the cache to update it

Btw, is this accurate? Will re-running `cache:warmup --env=prod` cause this to be overwritten?

nicolas-grekas: message updated

lyrixx:
Things a getting really complicated IMHO. We have so many `.env` files. 
It was much simpler when we were using .yaml file.
I totally understand why we needed to switch to .env file, but IMHO we should rethink everything.

For exemple why do we need a committed .env file while we say we need to defined a default value of each env var in the services.yaml ? (or another file). We need a default value for the default value ? 
(Actually, this is a rhetorical question. I know the reason: because if we want to use env var (not resolved I mean) even in the DIC we need this .env). 

But how many people really need really dynamic values ? a very few.
We are getting things really complicated for few people while we let so many people behind. 
There are plenty question like "how does env var work", "what should I put in my env var", "how does I change my DB in test" etc etc. Theses question comes on the table almost everyday on slacks (differents slack / different channel).
Most of people I work with / meet / train or discuss with does not understand where we are going and think it's now more complicated than before. 

Symfony 4 + Flex + MakerBundle is really cool and easy. But theses env var really need to be re-thought

nicolas-grekas:
@lyrixx :100: - that's what this PR provides :)

lyrixx:
BTW, this PR is a very good move :+1: 

About this point (the part "which you should"):
> if you use real env vars they always win (which you should - see https://12factor.net/)

I totally disagree. It's like always : **it depend**. Many applications are hosted on bare metal, and will not move. Using env var make things more complicated because you need to load them for FPM + CLI + CRON + Worker. Using a simple `.env` file  (like you porpose) is much more simple.

So we should not say : 

> you should use env var.

Declaring parameter in `.env` for just fine and easy for many!

nicolas-grekas:
After a pretty long discussion with @weaverryan on Slack, we figured out a very simple way to make all this work:
- logically load .env files all the time
- provide a `.env.local.php` file for prod - when found it replaces all .env files
- add a new dump-env command to flex so that deploying would be just `composer dump-env prod` (PR to come)

Simperfit:
👍

symfony/flex/1.0/.env
```diff
@@ -1,5 +1,13 @@
-# This file defines all environment variables that the application needs.
-# DO NOT DEFINE PRODUCTION SECRETS IN THIS FILE.
-# Use ".env.local" for local overrides during development.
-# Use real environment variables when deploying to production.
+# In all environments, the following files are loaded
+# if they exist:
```
lyrixx: Can you can talk about var precedence?

lyrixx: No. Like variable definer in `.env` are overrided with var in `.env.local` etc.

nicolas-grekas: updated

mhujer:
@nicolas-grekas This PR probably causes PHPUnit to fail:

```
Notice: Undefined index: APP_ENV in C:\dev\htdocs\sf-rec\config\bootstrap.php on line 19
```

Steps to 
```
composer create-project symfony/skeleton sf-rec
cd sf-rec
composer require --dev phpunit/phpunit
vendor\bin\phpunit # or vendor/bin/phpunit on Linux

Notice: Undefined index: APP_ENV in C:\dev\htdocs\sf-rec\config\bootstrap.php on line 19
PHPUnit 7.5.1 by Sebastian Bergmann and contributors.


Time: 93 ms, Memory: 2.00MB

No tests executed!
```

mhujer:
Nevermind, just found https://github.com/symfony/recipes/issues/512

eschultz-magix:
> For example why do we need a committed .env file while we say that we need to defined a default value for each env var in the services.yaml ? (or another file). We need a default value for the default value ?
(Actually, this is a rhetorical question. I know the reason: because if we want to use env var (not resolved I mean) even in the DIC we need this .env).

That's a good point and I also see a duplication in here. From the documentation (https://symfony.com/doc/current/configuration/external_parameters.html):

> You can also give the env() parameters a default value: the default value will be used whenever the corresponding environment variable is not found:

`# config/services.yaml
parameters:
    env(DATABASE_HOST): localhost`

By default, the kernel loads the services.yaml and a services_<env>.yaml for each environment as an extension to the services.yaml. Both files allow to set default values like the .env file. So why is the .env file REALLY necessary? Or on the other hand: Wouldn't it be better to deprecate the "env()" parameter feature then?

Pierstoval:
`Default value` and `Necessary env var` are not the same thing. Default value means "If we don't have it, use this". Necessary env var means "We need this env var, and possibly with a valid value".

A `.env` file is here to specify the necessary values. What it contains "by default" is just values "that should work", but sometimes it does not work.

For instance, check the Recaptcha recipe: Google credentials are mandatory, but they won't break the app if they're not valid. They will break only the portions that _actually_ use a captcha.

PR #503, opened 2018-12-08
Add a recipe for API Platform Admin

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

admin-demo-97cd2738071d63989db0bbcb6ba85a25

This PR makes it easy to use [API Platform Admin](https://api-platform.com/docs/admin) in a traditional Symfony app (an app using Twig).

The recipe installs Webpack Encore and API Platform (using their respective packs) and generate a Twig template, a route and the appropriate code to load the admin in a standard Symfony page. 

It works in a different way than in the "official" distribution of API Platform:

* in API Platform, the admin is a standalone Progressive Web App that must be hosted as a static website (using GitHub Page, Netlify or similar solutions)
* when using this recipe, the React component is loaded directly inside the existing Symfony application, through a Twig template (so it is not installable as a PWA, but doesn't require to setup a different server, CORS headers...).

### DIFF
--- a/api-platform/admin-pack/1.0/post-install.txt
+++ b/api-platform/admin-pack/1.0/post-install.txt
@@ -4,13 +4,13 @@
 
   * Your admin is almost ready:
     1. Install the JavaScript dependencies by running <comment>yarn add @api-platform/admin</comment>
-    2. Go in <comment>webpack.config.js</comment> and <info>uncomment</info> the following lines:
+    2. Edit <comment>webpack.config.js</comment> and <info>uncomment</info> the following lines:
 
         Encore
             // ...
             .enableReactPreset()
             .addEntry('admin', './assets/js/admin.js')
 
-  * Mark any entity with the <comment>@ApiResource</comment> annotation, it appears in the admin!
+  * Mark entities you want in the admin with the <comment>@ApiResource</comment> annotation, it appears in the admin!
 
   * Using MakerBundle? Try <info>php bin/console make:entity --api-resource</info>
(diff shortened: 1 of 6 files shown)


api-platform/admin-pack/1.0/post-install.txt
```diff
@@ -0,0 +1,16 @@
...
+  * Your admin is almost ready:
+    1. Install the JavaScript dependencies by running <comment>yarn add @api-platform/admin</comment>
+    2. Go in <comment>webpack.config.js</comment> and <info>uncomment</info> the following lines:
+
+        Encore
+            // ...
+            .enableReactPreset()
+            .addEntry('admin', './assets/js/admin.js')
+
+  * Mark any entity with the <comment>@ApiResource</comment> annotation, it appears in the admin!
```
fabpot: Mark entities you want in the admin with the <comment>@ApiResource</comment> annotation

api-platform/admin-pack/1.0/post-install.txt
```diff
@@ -0,0 +1,16 @@
+<bg=blue;fg=white>                       </>
+<bg=blue;fg=white> To finish the install </>
+<bg=blue;fg=white>                       </>
+
+  * Your admin is almost ready:
+    1. Install the JavaScript dependencies by running <comment>yarn add @api-platform/admin</comment>
+    2. Go in <comment>webpack.config.js</comment> and <info>uncomment</info> the following lines:
```
fabpot: Edit webpack...

rvdavid-rbi:
Whoa... Is this a realy thing still? I've been so buried and committed into the EasyAdmin track that I missed this.

PR #504, opened 2018-12-09
[FrameworkBundle] split cache.yaml out of framework.yaml

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Might be clearer.


weaverryan:
Needs an issue in symfony-docs please. The docs are very sensitive to recipe changes :)

weaverryan:
On this page, for example: https://symfony.com/doc/current/configuration.html#configuration-config-packages

More largely the point is that if we don't create docs issues, the docs will get left behind. In this case, there is only one minor code block that needs to be updated :)

nicolas-grekas:
Doc PR sent: https://github.com/symfony/symfony-docs/pull/10778

PR #507, opened 2018-12-22
symfony/framework-bundle 4.2 can use PHP 7.1 features

mhujer:
| Q             | A
| ------------- | ---
| License       | MIT

Symfony 4 requires PHP 7.1 so I think that the recipe can use features from 7.1. Otherwise I was doing those changes manually when creating a new SF project.

What do you think?

### DIFF
--- a/symfony/framework-bundle/4.2/src/Kernel.php
+++ b/symfony/framework-bundle/4.2/src/Kernel.php
@@ -13,9 +13,9 @@ class Kernel extends BaseKernel
 {
     use MicroKernelTrait;
 
-    const CONFIG_EXTS = '.{php,xml,yaml,yml}';
+    private const CONFIG_EXTS = '.{php,xml,yaml,yml}';
 
-    public function registerBundles()
+    public function registerBundles(): iterable
     {
         $contents = require $this->getProjectDir().'/config/bundles.php';
         foreach ($contents as $class => $envs) {
@@ -25,7 +25,7 @@ class Kernel extends BaseKernel
         }
     }
 
-    protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader)
+    protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader): void
     {
         $container->addResource(new FileResource($this->getProjectDir().'/config/bundles.php'));
         $container->setParameter('container.dumper.inline_class_loader', true);
@@ -37,7 +37,7 @@ class Kernel extends BaseKernel
         $loader->load($confDir.'/{services}_'.$this->environment.self::CONFIG_EXTS, 'glob');
     }
 
-    protected function configureRoutes(RouteCollectionBuilder $routes)
+    protected function configureRoutes(RouteCollectionBuilder $routes): void
     {
         $confDir = $this->getProjectDir().'/config';
 


symfony/framework-bundle/4.2/src/Kernel.php
```diff
@@ -13,19 +14,19 @@ class Kernel extends BaseKernel
 {
     use MicroKernelTrait;
 
-    const CONFIG_EXTS = '.{php,xml,yaml,yml}';
+    private const CONFIG_EXTS = '.{php,xml,yaml,yml}';
```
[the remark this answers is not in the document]
Pierstoval: As it's in recipes, people would only see this change in new projects or when running `sync:recipes --force`, which I think is good to keep track of best practices. I don't find this as a problem.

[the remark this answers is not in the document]
mhujer: Exactly! That's why I think that even `declare(strict_types = 1);` is not a BC break in this case.

symfony/framework-bundle/4.2/src/Kernel.php
```diff
@@ -13,19 +14,19 @@ class Kernel extends BaseKernel
...
     }
 
-    public function getLogDir()
+    public function getLogDir(): string
     {
         return $this->getProjectDir().'/var/log';
     }
 
-    public function registerBundles()
+    public function registerBundles(): Generator
```
xabbuh: should be `\Generator` to comply with Symfony's code style

nicolas-grekas: or better: iterable

symfony/framework-bundle/4.2/config/bootstrap.php
```diff
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types = 1);
```
chalasr: spaces should be removed `declare(strict_types=1);`

symfony/framework-bundle/4.2/config/bootstrap.php
```diff
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
```
ro0NL:
```php
<?php

declare(strict_types=1);

use ...
```

is the usual code style AFAIK

ro0NL: (blank_line_after_opening_tag)

mhujer: fixed

mhujer:
Rebased to resolve conflicts with #521.

Is there anything else that needs to be done before merging?

symfony/framework-bundle/4.2/config/bootstrap.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
```
fabpot: This should be removed in all files as we are not using strict_types in Symfony.

mhujer: @fabpot removed.

I would be happy to know the reasons. There is an issue (https://github.com/symfony/symfony/issues/28423) discussing that it can't be used for parts of the framework which are backported to older versions. But it suggests that it may be used for new components, which won't be backported to <4.0 versions.

PR #508, opened 2018-12-22
Optimize Encore Init

matthieumota:
| Q             | A
| ------------- | ---
| License       | MIT

Some optimization when we init encore with Flex. We can use ```let``` instruction because webpack transpile that. I think that gitignore is not needed because css and js folder exists at creation. If the developer change that folder, it's his responsability to create this file.

### DIFF
--- a/symfony/webpack-encore-bundle/1.0/assets/js/app.js
+++ b/symfony/webpack-encore-bundle/1.0/assets/js/app.js
@@ -9,6 +9,6 @@
 require('../css/app.css');
 
 // Need jQuery? Install it with "yarn add jquery", then uncomment to require it.
-// var $ = require('jquery');
+// let $ = require('jquery');
 
 console.log('Hello Webpack Encore! Edit me in assets/js/app.js');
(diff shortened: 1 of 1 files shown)


symfony/webpack-encore-bundle/1.0/assets/js/app.js
```diff
@@ -9,6 +9,6 @@
 require('../css/app.css');
 
 // Need jQuery? Install it with "yarn add jquery", then uncomment to require it.
-// var $ = require('jquery');
+// let $ = require('jquery');
```
bocharsky-bw: Hm, why not `const $ = require('jquery');`? I think it would be even better because we do not want to override `$`

matthieumota: Agree with that. No need to override $ at any moment. I wait @weaverryan review that and I can commit.

weaverryan: Yea, const is better - good idea - can you make that update?

matthieumota: Done !

Nyholm:
I will try to "dismiss" Ryans review.

PR #509, opened 2018-12-22
Add .splitEntryChunks() to webpack.config.js

bocharsky-bw:
| Q             | A
| ------------- | ---
| License       | MIT

Don't we need adding .`.splitEntryChunks()` as pointed in docs? See https://github.com/symfony/webpack-encore-bundle#usage

### DIFF
--- a/symfony/webpack-encore-bundle/1.0/webpack.config.js
+++ b/symfony/webpack-encore-bundle/1.0/webpack.config.js
@@ -21,6 +21,9 @@ Encore
     //.addEntry('page1', './assets/js/page1.js')
     //.addEntry('page2', './assets/js/page2.js')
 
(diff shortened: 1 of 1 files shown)


weaverryan:
Yea, I kinda think we should. Everyone should now be using the Twig helpers, so you get this optimization “for free” by enabling this.

bocharsky-bw:
@weaverryan Can take a look at docs later, but what do you mean about different output files in the “first example” page? We still have runtime.js, app.css, and app.js. We also mention entrypoints.json on that page too, though I don't see any mention for manifest.json there

weaverryan:
Hmm, you might be right that no changes are needed. I was thinking that some files might be “split” already - and so we might have a vendor~...js type of file. But probably there isn’t enough to cause splitting in that first chapter.

+1 for this PR

PR #510, opened 2018-12-23
PHPUnit: use XSD schema from vendor

mhujer:
| Q             | A
| ------------- | ---
| License       | MIT

Instead of having to manually downloading the XSD to get completion (It available in Alt+Enter menu in PhpStorm), we can use the XSD directly from the installed package.


nicolas-grekas:
Does this work when the simple-phpunit script from the bridge is used?

mhujer:
Ups, maybe not. I'm personally not using it, so I will try it and verify this.

Pierstoval:
By the way, integration of the phpunit bridge in PHPStorm is really cumbersome, so at least keeping this canonical xsd reference is the best option IMO

mhujer:
My change does not work with simple-phpunit, because it installs PHPUnit separately into 
`bin/.phpunit/phpunit-6.5/`, so I'm closing this PR.

PR #518, opened 2019-01-09
Always load .env.local even in test environment

bocharsky-bw:
| Q             | A
| ------------- | ---
| License       | MIT

I'm not sure about the original purpose of this, but does anybody have a good example when it's useful?

Because for me it does not seem useful. For example, I may have some paths to binaries that are different on different systems, so I'd like to overwrite these paths with my `.env.local` that are defaulted in `.env`. And I do want to use those new paths in every environment, not only in `dev`/`prod` skipping `test`. And I don't see a value in duplicating them one more time in `.env.test.local`. Thoughts?


bocharsky-bw:
Oh, looks like on purpose. OK, thanks for the explanation

weaverryan:
We should document this - it would bite me too, honestly.

PR #519, opened 2019-01-09
Remove apostrophes from trusted hosts

ChristianVermeulen:
| Q             | A
| ------------- | ---
| License       | MIT

When activating the `TRUSTED_PROXIES` in the `.env` the enclosing apostrophes mess up the regex. The `Request::SetTrustedProxies` turns this into `{'^localhost|example\\.com$'}i`. In this form the preg_match can not match the hosts.

On top of that, the double `\` also messes it up because this escapes the backslash itself instead of the dot. Therefore the regex will be matching _any charachter except linebreaks_ in the url, while it should be matching the dot.

### DIFF
--- a/symfony/framework-bundle/3.3/manifest.json
+++ b/symfony/framework-bundle/3.3/manifest.json
@@ -15,7 +15,7 @@
         "APP_ENV": "dev",
         "APP_SECRET": "%generate(secret)%",
         "#TRUSTED_PROXIES": "127.0.0.1,127.0.0.2",
-        "#TRUSTED_HOSTS": "^localhost|example\\.com$"
+        "#TRUSTED_HOSTS": "^localhost|example\.com$"
     },
     "gitignore": [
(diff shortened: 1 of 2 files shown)


ChristianVermeulen:
Hmm, this is interesting, apparently the flex tester does not accept the dot escaping. So presumably the extra slash is to satisfy the tests.

However, if that is the case, then it should be factored in when executing the configuration so the double backslash does not end up in the actual `.env` file.

🤔

nicolas-grekas:
Are you using docker-composer? It has a buggy .env parser. Otherwise, the quotes are correctly parsed by all loaders.
Nothing to fix here to me.

ChristianVermeulen:
@nicolas-grekas why yes I do! That sucks... Never crossed my mind it could've been that one. Thanks for the heads up! Hopefully this (closed) PR will be useful to googlers in the future!

PR #524, opened 2019-01-24
adjust recipe configuration

dbu:
preserving headers on redirects includes a potential security risk because on redirect to a different domain credential headers would be leaked.

| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/php-http/httplug-bundle/1.6/config/packages/httplug.yaml
+++ b/php-http/httplug-bundle/1.6/config/packages/httplug.yaml
@@ -1,7 +1,7 @@
 httplug:
     plugins:
-        redirect:
-            preserve_header: true
+        retry:
+            retry: 1
 
     discovery:
         client: 'auto'
@@ -12,3 +12,7 @@ httplug:
             plugins:
                 - 'httplug.plugin.content_length'
                 - 'httplug.plugin.redirect'
+                - header_defaults:
+                    headers:
+                        "X-Httplug": "This is just an example".
+                - 'httplug.plugin.retry'
--- a/php-http/httplug-bundle/1.6/config/packages/httplug.yaml
+++ b/php-http/httplug-bundle/1.6/config/packages/httplug.yaml
@@ -14,5 +14,5 @@ httplug:
                 - 'httplug.plugin.redirect'
                 - header_defaults:
                     headers:
-                        "X-Httplug": "This is just an example".
+                        "X-Httplug": "This is just an example"
                 - 'httplug.plugin.retry'


php-http/httplug-bundle/1.6/config/packages/httplug.yaml
```diff
@@ -12,3 +12,7 @@ httplug:
             plugins:
                 - 'httplug.plugin.content_length'
                 - 'httplug.plugin.redirect'
+                - header_defaults:
+                    headers:
```
dbu: i don't see what would be wrong with this line, its allowed by the configuration: https://github.com/php-http/HttplugBundle/blob/v1.6.0/DependencyInjection/Configuration.php#L319

Nyholm:
I had a similar issue in https://github.com/symfony/recipes-contrib/pull/591/files

You could update the recipe with the super expressive syntax and you will be fine. 
Sorry for not catching this before.

dbu:
i don't see the connection with the buzz pull request. for buzz you are adding a service definition file. or what do you mean with "super expressive syntax"? i tried adding a newline when we have a nested definition after a `-` but that seems not to help.

Nyholm:
> i tried adding a newline when we have a nested definition after a - but that seems not to help.

Yeah, that is what i meant. Hm.

php-http/httplug-bundle/1.6/config/packages/httplug.yaml
xabbuh: What happens with the following?

```yaml
httplug:
    plugins:
        redirect:
            preserve_header: true

    discovery:
        client: 'auto'

    clients:
        app:
            http_methods_client: true
            plugins:
                - 'httplug.plugin.content_length'
                - 'httplug.plugin.redirect'
                - header_defaults:
                      headers:
                          "X-Httplug": "This is just an example."
                - 'httplug.plugin.retry'
```

Nyholm: That’s is what we started with. Right? @dbu

xabbuh: I moved the dot of the header value into the quotes.

dbu: ah damn, i did not see that dot :man_facepalming: 
yep, that was the problem

xabbuh: The error message was indeed a bit misleading. http://yaml-online-parser.appspot.com/ is a good place to quickly verify some YAML string.

Nyholm:
This recipe is moved to contrib.

dbu:
@Nyholm as you also contribute to httplug: do you know if this is already committed in contrib or does this mean i should re-create the change against contrib?

Nyholm:
I know the recipe was just moved. So yes. Recreate the issue. 

Sorry for being short when closing.

PR #525, opened 2019-01-28
fix mercure environment variable name in parameters

GregoireHebert:
| Q             | A
| ------------- | ---
| License       | MIT

The default environment variable set when installing is `MERCURE_JWT ` not `MERCURE_JWT_SECRET`, It contains the JWT token not the secret of it.
The envvar is correctly set but the parameter used is defined with the wrong one.

### DIFF
--- a/symfony/mercure-bundle/0.1/config/packages/mercure.yaml
+++ b/symfony/mercure-bundle/0.1/config/packages/mercure.yaml
@@ -2,4 +2,4 @@ mercure:
     hubs:
         default:
             url: '%env(MERCURE_PUBLISH_URL)%'
-            jwt: '%env(MERCURE_JWT_SECRET)%'
+            jwt: '%env(MERCURE_JWT)%'


GregoireHebert:
ping @dunglas

dunglas:
This was done on purpose. The `_SECRET` suffix triggers some security features (like hiding the env var from logs) in some cloud providers, including SymfonyCloud. It's why it has this name, even if I agree that it can be confusing (the initial name was without the suffit, but I've been asked to change it).

Also, the JWT itself must be stored with care, it's equivalent to a password (especially if it doesn't expire).

GregoireHebert:
Alright then we should change the environment variable set in https://github.com/api-platform/api-platform/blob/master/api/.env#L30 by default.

Closing.

Pierstoval:
I also was confued a lot by this var name, I didn't understand that it was the secret at first and couldn't make Mercure work during the workshop because of this 😕 

I don't see why the fact that it's hidden on SfCloud or other provides is a problem: it's a "secret" after all, and should be kept secret 🙂

stof:
> I don't see why the fact that it's hidden on SfCloud or other provides is a problem: it's a "secret" after all, and should be kept secret slightly_smiling_face

that's not an issue. That's a feature. This is one of the reasons why this suffix was *added* previously.

teohhanhui:
This is really confusing, because this is a JWT token which is signed with a **secret** key. It's too easy to confuse the two.

teohhanhui:
IMO this should be named `MERCURE_JWT_TOKEN`. The CI providers should be smart enough to obfuscate anything that says "token" too, no? If not, perhaps they should?

dunglas:
I agree that it would be super cool if SymfonyCloud could hide variables containing the `_TOKEN` suffix too.

fabpot:
@dunglas That's already the case. when you add an env var that ends with `_TOKEN`, it's hidden (same for some other suffixes like `_PASS` or `_SECRET`).

teohhanhui:
So we can go ahead and fix the name of this env var, to avoid confusion.

PR #526, opened 2019-01-29
Making doctrine package config easier 

ad3n:
| Q             | A
| ------------- | ---
| License       | MIT

Making doctrine package config easier


fabpot:
It's done this way for a reason: Doctrine need the information at compile time. So, this cannot be configured via env vars.

PR #530, opened 2019-02-19
(WIP) Make tests faster by default

javiereguiluz:
| Q             | A
| ------------- | ---
| License       | MIT

As explained by @stof in [this comment](https://github.com/symfony/symfony/issues/30290#issuecomment-464846550) most of the times you can run tests in the (much faster) "debug = false" mode.

This PR is incomplete! We need to see how to implement the clearing of the test cache when needed. @stof do you suggest to use something similar to what you showed in this gist? https://gist.github.com/stof/fb45ca5940159b2741081e1cb2dae3b0

Thanks.

### DIFF
--- a/symfony/phpunit-bridge/3.3/phpunit.xml.dist
+++ b/symfony/phpunit-bridge/3.3/phpunit.xml.dist
@@ -10,6 +10,7 @@
     <php>
         <ini name="error_reporting" value="-1" />
         <env name="APP_ENV" value="test" />
+        <env name="APP_DEBUG" value="false" />
         <env name="SHELL_VERBOSITY" value="-1" />
     </php>
 


B-Galati:
One solution for the bootstrap file that would be simpler than a class.

```
require __DIR__ . '/../vendor/autoload.php';

$keepSymfonyCache = $_SERVER['KEEP_SYMFONY_TEST_CACHE'] ?? $_ENV['KEEP_SYMFONY_TEST_CACHE'] ?? false;

if (!$keepSymfonyCache) {
    $filesystem = new Filesystem();

    $filesystem->remove([__DIR__.'/../var/cache/test']);

    echo "Test cache cleared\n";
}
```

Far from perfect, it's just to give ideas.

mhujer:
Just tried flipping the switch on our project and it sped the test suite from `1.48 minutes` to `52.87 seconds`.

Awesome 👏 👍

fbourigault:
As I wrote in https://github.com/symfony/symfony/issues/30290, I'm using the following `bootstrap.php`:

```php
<?php

use Symfony\Component\Config\Resource\SelfCheckingResourceChecker;
use Symfony\Component\Config\ResourceCheckerConfigCache;
use Symfony\Component\Filesystem\Filesystem;

require __DIR__.'/../../vendor/autoload.php';

$cacheDir = __DIR__ . '/../../var/cache/test';
$configCache = new ResourceCheckerConfigCache("{$cacheDir}/appAppKernelTestContainer.php", [new SelfCheckingResourceChecker()]);
if (!$configCache->isFresh()) {
    $filesystem = new Filesystem();
    $filesystem->remove($cacheDir);
}
```
I use this since a week and it works like a charm. My container is fresh like before but the check is only done once and so I halved my Behat suite execution time.

It's a bit hardcoded, but with some changes in the Kernel, it would be possible to have this code dynamic and so get this a viable solution.

symfony/phpunit-bridge/3.3/phpunit.xml.dist
```diff
@@ -10,6 +10,7 @@
     <php>
         <ini name="error_reporting" value="-1" />
         <env name="APP_ENV" value="test" />
+        <env name="APP_DEBUG" value="false" />
```
alexander-schranz: We mostly have 4 environments in our application where the additional does match mostly the prod configuration, so I did change this line in the [bootstrap.php](https://github.com/symfony/recipes/blob/d9dc7a8ff5b46d20b1197d14bcf2551a448a8ae1/symfony/framework-bundle/4.2/config/bootstrap.php#L20) to avoid a `in_array` check to  `... && 'dev' === $_SERVER['APP_ENV']`.

B-Galati:
Note that the profiler will not be available by default in test environment if APP_DEBUG is false.

I made a PR to document that behavior symfony/symfony-docs#11429

fabpot:
Now that we moved everything back to `.env` files, shouldn't we add `APP_DEBUG=0` to a `.env.local.test` file instead?

teohhanhui:
We also need:

```yaml
framework:
    profiler: false
```

in https://github.com/symfony/recipes/blob/master/symfony/web-profiler-bundle/3.3/config/packages/test/web_profiler.yaml

nicolas-grekas:
Alternative proposal: https://github.com/symfony/symfony/pull/34399

javiereguiluz:
Let's close this ... because the new proposal looks much safer and numbers are truly promising! Thanks.

maks-rafalko:
Besides the speed, there is one more (I think major) benefit of using `APP_DEBUG=false`. Functional tests start responding with `Internal Server Error` rather than with real exception message.

This is a deal breaker in such projects like api-platform. I saw 2-3 projects, where people asserting that response contains some exception message `XXX` in **functional tests** when `APP_DEBUG=true`, while in fact the response message is `Internal Server Error`, and test start failing after using `APP_DEBUG=false`.

--

TL;DR Using `APP_DEBUG=false` with functional tests is right way from `5xx` errors points of view.

teohhanhui:
Anyway, there's nothing unsafe about setting `kernel.debug` to `false` in the test environment...

PR #537, opened 2019-02-25
Load routes for specific env first

lyrixx:
It's common to add a 'wildcard' route to an application that looks like
a CMS. Usually the route has the following path `/{slug}`.

But with the current routes loading order, the 'dev' routing is loaded
after regular routes. In previous SF versions, dev routing was loaded
first. This PR retores this behavior.

| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/framework-bundle/4.2/src/Kernel.php
+++ b/symfony/framework-bundle/4.2/src/Kernel.php
@@ -41,8 +41,8 @@ class Kernel extends BaseKernel
     {
         $confDir = $this->getProjectDir().'/config';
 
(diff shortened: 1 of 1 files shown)


lyrixx:
Should I do this path for 3.3 too?

stof:
> But with the current routes loading order, the 'dev' routing is loaded
> after regular routes. In previous SF versions, dev routing was loaded
> first. This PR retores this behavior.

To be exact, old versions were importing *only* the `routing_dev.yml` file, which was itself responsible for importing the normal file. This allowed loading dev routes both before and after the normal ones.

PR #538, opened 2019-02-26
Default favicon, preventing 404 requests on first startup

SerheyDolgushev:
| Q             | A
| ------------- | ---
| License       | MIT


fabpot:
We did not add a favicon on purpose (especially a Symfony one).

SerheyDolgushev:
Sorry, what is the purpose to not adding that? Maybe an empty one can be used?

Pierstoval:
I think a 404 is fine for the favicon: it reminds you to add your own, instead of relying on the Symfony one that would be visible everywhere.

SerheyDolgushev:
Sorry, so the main reason to not have any favicon is to avoid a bunch of Symfony projects with default favicon?
Maybe empty pixel favicon would be the best solution in this case?
1. It will fix 404 error.
2. There still will be an option to change favicon
3. If favicon is not changed, no default one will be shown.

stof:
But why is avoiding the 404 an advantage ? This 404 reminds you that you need to add a favicon in your project.

glennunipro:
I think it makes sense to ship with a default favicon. If I'm spinning up a bunch of sandbox projects in Symfony or Angular, it's nice to see in the browser which bunch of tabs are for what. I don't see a _downside_ to adding a default favicon.

nicolas-grekas:
It would be fun to have flex generate a random one per project :)

OskarStark:
Yeah like github creates avatars 😃

stof:
@glennunipro I don't see how putting the same favicon in all these sandbox projects would help you identify each tab, as they would get the same.

SerheyDolgushev:
So seems like random favicons is the winner here?

glennunipro:
@stof better than no favicon. Typically I've got a Symfony and Angular app on the go, distinguishing them at the moment is "which one has _no_ favicon?".

stof:
@glennunipro a transparent favicon would be indistinguishable from no favicon.

A symfony favicon is a bad idea: if you forget to change the favicon before deploying your app, there is a very good chance that you would be violating the Symfony trademark: https://symfony.com/trademark

stof:
And a random favicon would be something that cannot be done as a recipe. That would require building it in Flex (or as a separate composer plugin)

OskarStark:
We could use something like lorempixum where only a url is needed

nicolas-grekas:
That would create privacy concerns IMHO.

Pierstoval:
Another problem is that the base favicon must be an `.ico` file, so this needs conversion after download

fabpot:
Let's close as it does not seem to be easy to solve.

PR #540, opened 2019-03-01
Fix getenv() not working with .env.local.php

MisatoTremor:
| Q             | A
| ------------- | ---
| License       | MIT

When dumping the env using Flex 1.2 ``dump-env`` the getenv function does no longer work for these variables.
This is a quick fix just like it is done in ``DotEnv::populate``, but might need further review.


nicolas-grekas:
Sorry but this not have been merged: `putenv` is not thread safe and calling it breaks app on ZTS.

nicolas-grekas:
Reverted in  #571

PR #541, opened 2019-03-01
Documenting the WebpackEncoreBundle caching option

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

See symfony/webpack-encore-bundle#12

This is only available in 1.2, but I don't see a good reason to duplicate the recipe - it's an opt-in feature (so it will always be commented-out by default) and even if you, somehow, installed an older version today, you will now be aware of this feature.

Thanks!


Pierstoval:
I suggest to create a new recipe for 1.2 because  ̀composer sync-recipes --force` could be used to reset the recipe, but if the person running the command has only 1.0 or 1.1, they might uncomment this line and end up with errors...

Of course they should upgrade anyway, but IMO this is not the best option to introduce a comment in a version not compatible with this option...

weaverryan:
We'll see what other's think :). If someone runs `sync-recipes --force` and they now have this new comment, that's great! They get to see a new option, with documentation that says that it is available in 1.2 of the bundle.

PR #542, opened 2019-03-04
Fix creating of prod matching environments

alexander-schranz:
| Q             | A
| ------------- | ---
| License       | MIT

Based on the [documentation](https://symfony.com/doc/current/configuration/environments.html#creating-a-new-environment) there should be not more be needed to create a `prod` similiar environment:

```yaml
imports:
    - { resource: '../prod/' }
```

but thats because of `APP_DEBUG` not true. With this change the new environment will match 100% to the prod env.

### DIFF
--- a/symfony/phpunit-bridge/3.3/.env.test
+++ b/symfony/phpunit-bridge/3.3/.env.test
@@ -1,4 +1,5 @@
 # define your env variables for the test env here
 KERNEL_CLASS='App\Kernel'
 APP_SECRET='s$cretf0rt3st'
+APP_DEBUG='true'
 SYMFONY_DEPRECATIONS_HELPER=999999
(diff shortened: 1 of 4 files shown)


symfony/phpunit-bridge/3.3/.env.test
```diff
@@ -1,4 +1,5 @@
 # define your env variables for the test env here
 KERNEL_CLASS='App\Kernel'
 APP_SECRET='s$cretf0rt3st'
+APP_DEBUG='true'
```
alexander-schranz: Will maybe be not longer needed when https://github.com/symfony/recipes/pull/530 is merged

Nyholm:
I think the current implementation is a very good default. 
Users are of course free to change this in their apps.

PR #543, opened 2019-03-07
Check for Debug class existence

cvdwel:
It's being checked in bin/console, to keep things consistent let's check it for web requests as well.

| Q             | A
| ------------- | ---
| License       | MIT


nicolas-grekas:
Debug is a dependency of HttpKernel, which is itself a dependency of FrameworkBundle: this class is guaranteed to be always present. That's not the case for Console, that's why this difference exists.

cvdwel:
Thanks for the explanation!

PR #544, opened 2019-03-09
explode trustedHosts to array

danielchodusov:
This fixes problem when multiple trusted hosts are separated by comma in .env file and not being matched in Symfony\Component\HttpFoundation\Request::getHost method, causing SuspiciousOperationException exception.

Found when upgrading to new .env 

https://symfony.com/doc/current/configuration/dot-env-changes.html
> Update your public/index.php (index.php diff) file to load the new config/bootstrap.php file.
>  If you've customized this file, make sure to keep those changes (but use the rest of the changes).

| Q             | A
| ------------- | ---
| License       | MIT


nicolas-grekas:
Trusted hosts is a regular expression. There has been some confusion in the doc before. Exploding by commas will break the regexps. If you need multiple hosts, you have to write a single regexps that matches them all. Don't forget to anchor it and to escape dots.

PR #547, opened 2019-03-11
Rename config files in order to be consistent with the config nodes that they are concerned to

### LINKED ISSUE #396: [RFC] Use config node path as filename for recipes which are not root nodes
With this RFC, I want to propose that recipes containing configuration for sub nodes have a name according that node. By instance, configuration file for [symfony/translation](https://github.com/symfony/recipes/blob/master/symfony/translation/3.3/config/packages/translation.yaml#L3) recipe should be named `framework.translator.yaml`.
The main reason I'm proposing this, is based in the following assumption / idea / example about the current behavior:

1- Configuration files under "config/packages/" dir are mapped to packages (bundles), and you would expect the whole configuration for a single package be included in a single configuration file.
2- You already have declared configuration for subnodes at `framework.yaml`, including `framework.translator` node.
3- You install a recipe which ignore completely the already configured path and will install duplicated configuration for `framework.translator` node in `config/packages/translation.yaml`

For these reasons, I think it would be better to expose sub-node recipes with filenames that makes the node inheritance clear.

phansys:
|Q            |A     |
|---          |---   |
|Branch       |master|
|Bug fix?     |no    |
|New feature? |no    |
|BC breaks?   |maybe    |
|Deprecations?|no    |
|Tests pass?  |yes   |
|Fixed tickets|#396  |
|License      |MIT   |
|Doc PR       |n/a   |
                      

Closes #396.


fabpot:
This proposal makes file names longer for no real benefits IMHO. I'm 👎 for this change.

Nyholm:
I understand the idea. And Im happy you made this PR. But Im also leaning towards 👎 . 
Unless some other maintainer feels strongly about this Im going to close it.

phansys:
Thank you for your feedback!
@Nyholm, I think there are positive votes for this in https://github.com/symfony/recipes/issues/396#issuecomment-380651405, or at least, there is the intention to improve the current solution. So, I would like to keep the discussion open if possible.
One of the main benefits that I see in this proposal is the way it provides to avoid naming collisions between different packages providing similar features or covering similar responsibilities. By instance, if I have a package for cache handling, I need to take care about `cache.yaml`, which is probably in use by `framework.cache` configuration. The same applies for many other components.
Given that reasoning, I think a more robust, clean and "honest" way to define these package configurations is respecting their node path, relative to the configuration root.
I think the current approach is arbitrary, ambiguous and oblies the package provider to check if there is a previous existing recipe (regardless it is official or not) using a name for a file. This problem can be avoided in a transparent way following what I have proposed, given that name collisions between configuration nodes is already covered by the configuration itself.
IMHO, based in the explained gain, the longer names are an acceptable impact.
If the downside about the size of the filenames are taken as a no-go, I think the @OskarStark's proposal (https://github.com/symfony/recipes/issues/396#issuecomment-472733286) about merging children nodes in the main configuration file for each package is a good alternative.

PR #548, opened 2019-03-11
Add recipe for http-client

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Needs https://github.com/symfony/symfony/pull/30419

### DIFF
--- a/symfony/http-client/4.3/config/packages/http_client.yaml
+++ b/symfony/http-client/4.3/config/packages/http_client.yaml
@@ -1,0 +1,4 @@
+framework:
+    http_client:
+        default_options:
+            # see HttpClientInterface::OPTIONS_DEFAULTS for existing options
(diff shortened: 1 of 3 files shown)


symfony/http-client/4.3/config/packages/http_client.yaml
```diff
@@ -0,0 +1,4 @@
+framework:
```
OskarStark: In other recipes we only create config files if the bring additional value and don't only expose default variables and this looks only like the defaults 🤔

nicolas-grekas: the `http_client` service needs to be explicitly enabled, this does it
but maybe we should enable it conditionally in the fwb integration!
Is that what your mean? Doing a class_exists check there and remove the recipe?

Pierstoval: If the HttpClient package is not required by any of the skeleton's dependencies, it means it certainly can be activated automatically by the fwb if the package is installed, this would probably save a recipe for it 😄

OskarStark: Yes I thought about it. Does it makes sense?

weaverryan: I just added a comment on the PR that we should auto-activate it if the component is installed.

PR #549, opened 2019-03-12
Check for the dev environment instead of prod

trsteel88:
It feels dangerous that we're checking whether the environment !== 'prod'.

I think APP_DEBUG should only default to true if the environment is specifically "dev".

In a real example I have created a "staging" environment and it is showing debug information. I now need to specifically create a env var to stop this. Unless users are aware of this (or there is human error in creating this env var), the server could reveal too much information creating security issues.

| Q             | A
| ------------- | ---
| License       | MIT


weaverryan:
It makes sense, but we also advertise that `prod` is your production environment and you would expect different behavior in other environments. Plus, hopefully the behavior if you deploy and see the web debug toolbar will be pretty obvious.

So, while I think this is sound, it's not worth changing it (these types of changes are a pain to roll out consistently to the community).

-1 from me

But thanks!

OskarStark:
I am against this change too, as I have many more environments which are != prod and I have to maintain them all afterwards

fabpot:
Closing as I agree with the other commenters.

PR #546, opened 2019-03-11
Add nyholm/psr7 to the official recipes

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT

PR #550, opened 2019-03-12
Add PSR7 and PSR17 aliases

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT

This is a follow up on #546. I forgot to add some aliases.

### DIFF
--- a/nyholm/psr7/1.0/manifest.json
+++ b/nyholm/psr7/1.0/manifest.json
@@ -1,5 +1,5 @@
 {
-    "aliases": ["psr7", "psr-7","psr17", "psr-17"],
+    "aliases": ["psr7", "psr-7", "psr17", "psr-17"],
     "copy-from-recipe": {
         "config/": "%CONFIG_DIR%/"
     }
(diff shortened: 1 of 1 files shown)


nyholm/psr7/1.0/manifest.json
```diff
@@ -1,4 +1,5 @@
 {
+    "aliases": ["psr7", "psr-7","psr17", "psr-17"],
```
fabpot: space missing after the second comma

Nyholm: Thank you

Nyholm:
> I thought about this also, but wouldn't one expect to get the bridge when doing "composer require psr7"?

Actually no =)
I would not expect the bridge when I just want a PSR7 implementation.. 

Suggestion: I modify this to alias "PSR-17", then create a pack for alias "PSR-7". That pack would include the bridge and nyholm/psr7.

nicolas-grekas:
> That pack would include the bridge and nyholm/psr7

LGTM - we'd also need sensio-framework-extra-bundle in that pack so that everything works seamlessly, isn't it?

Nyholm:
Correct

Tobion:
Why should we alias this package for psr-7 for example? There are alot of other implementations that would qualify for this as well.

fabpot:
Aliases are choices made by Symfony. This is a perfect example and does not prevent anyone from installing something else.

Tobion:
To me, psr-7 should be added to https://github.com/symfony/psr-http-message-bridge or https://github.com/guzzle/psr7

PR #551, opened 2019-03-12
[cs-fixer] Configure properly the default finder

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

Provide a default Finder configuration matching the Flex directory structure. Then PHP-CS-Fixer can be called with `vendor/bin/php-cs-fixer fix` can be called directly.

### DIFF
--- a/friendsofphp/php-cs-fixer/2.2/.php_cs.dist
+++ b/friendsofphp/php-cs-fixer/2.2/.php_cs.dist
@@ -1,8 +1,14 @@
 <?php
 
+$finder = PhpCsFixer\Finder::create()
+    ->in(__DIR__)
+    ->exclude('var')
+;
+
 return PhpCsFixer\Config::create()
     ->setRules([
         '@Symfony' => true,
         'array_syntax' => ['syntax' => 'short'],
     ])
+    ->setFinder($finder)
 ;


friendsofphp/php-cs-fixer/2.2/.php_cs.dist
```diff
@@ -1,8 +1,14 @@
 <?php
 
+$finder = PhpCsFixer\Finder::create()
+    ->in(__DIR__)
```
Pierstoval: Why not 
```php
->in([
    __DIR__.'/src/',
    __DIR__.'/tests/',
])
```
Much more explicit, and by default excludes var and vendors

Nyholm: `->(__DIR__.'/src')`

Pierstoval: Haha, seems like we posted at the same time, I also suggested `tests/` :)

dunglas:
@Nyholm @Pierstoval There are PHP files in `bin/` and `config/` too.

Nyholm:
Yeah, But not really that I (the user) ever edit. 

Btw, Should we exclude `vendor`? Or is that automatic?

dunglas:
For vendor, it is automatic yes: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/bbcbe2eef9d1743969e998d956c1a4abd47fb6ea/src/Finder.php#L33

Other files are often edited unfortunately. For instance need to edit the trusted headers configuration for every of my new projects because I use k8s and not AWS (`web/index.php`). I also often tweak `config/bootstrap.php` and I'm sure I'm not the only one :)

Pierstoval:
> There are PHP files in `bin/`, `config/` and `public/` too.

True, but I'm not sure this is really relevant to add them since they are entrypoints and not really "business logic source code".

And for the `bin/` directory, if we want to apply cs fixing we must add the files manually in the finder since they don't have the `.php` extension, this applies mostly for console & phpunit, but could be applied to other scripts (which I sometimes add here too).

friendsofphp/php-cs-fixer/2.2/.php_cs.dist
```diff
@@ -1,8 +1,14 @@
...
+    ->exclude('var')
+;
+
 return PhpCsFixer\Config::create()
     ->setRules([
         '@Symfony' => true,
         'array_syntax' => ['syntax' => 'short'],
```
OskarStark: off topic here but @Symfony should soon include short array syntax rule IMO

friendsofphp/php-cs-fixer/2.2/.php_cs.dist
OskarStark: I like your proposal @Pierstoval 👌🏻

OskarStark:
I am against bin folder but public and config sounds good 👍🏻

friendsofphp/php-cs-fixer/2.2/.php_cs.dist
Pierstoval: Now it does by default: https://github.com/symfony/symfony/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+%22short+array%22+in%3Atitle+

dunglas:
@OskarStark as pointed ou by @Pierstoval, files in `bin/` are actually ignored because they don't end by `.php`.

friendsofphp/php-cs-fixer/2.2/.php_cs.dist
OskarStark: I mean PHP CS Fixer should include short array in the „group“ of fixers in „@ Symfony“

friendsofphp/php-cs-fixer/2.2/.php_cs.dist
lyrixx: I agree too

PR #552, opened 2019-03-12
[MonologBundle] Added (commented) configuration for log format

lyrixx:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/monolog-bundle/3.1/config/packages/dev/monolog.yaml
+++ b/symfony/monolog-bundle/3.1/config/packages/dev/monolog.yaml
@@ -17,3 +17,7 @@ monolog:
             type: console
             process_psr_3_messages: false
             channels: ["!event", "!doctrine", "!console"]
+            # # As of Symfony 4.3 you can configure the log output:
+            # console_formater_options:
+            #     format: "%%datetime%% %%start_tag%%%%level_name%%%%end_tag%% <comment>[%%channel%%]</> %%message%%%%context%%\n"
+            #     multiline: false


symfony/monolog-bundle/3.1/config/packages/dev/monolog.yaml
```diff
@@ -17,3 +17,7 @@ monolog:
             type: console
             process_psr_3_messages: false
             channels: ["!event", "!doctrine", "!console"]
+            # # As of Symfony 4.3 you can configure the log output:
```
Pierstoval: Is there a documentation link that explains all format options?

lyrixx: I can not find it on symfony.com.
But here is the list https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Monolog/Formatter/ConsoleFormatter.php#L50-L54

Would you mind create a PR? :heart:

Pierstoval: I don't really have time for this right now, and I guess your recipe would be merged before I could have time to work on it 🤣 , but I think it's an easy-pick so I proposed it in symfony/symfony-docs#11144 to see if someone wants an easy first contribution 😄

Nyholm:
I would much rather have a link to the documentation that showing comments like this in the recipes.

lyrixx:
@Nyholm Let's merge it?

PR #553, opened 2019-03-17
Add ".phpunit.result.cache" to .gitignore

gaiaz-iusipov:
| Q             | A
| ------------- | ---
| License       | MIT


mhujer:
@nicolas-grekas PHPUnit creates those cache files in different directories in the tree. When running a single test case through PhpStorm, it will create `.phpunit.result.cache` in the directory where the test case is located.

PR #555, opened 2019-03-23
Add project path to app Kernel instanciation

tdutrion:
| Q             | A
| ------------- | ---
| License       | MIT

Changed invocation of the Kernel according to the feature from symfony/symfony#30651.
This removes the need for a `composer.json` to define the project folder when using the framework bundle.

### DIFF
--- a/symfony/console/3.3/bin/console
+++ b/symfony/console/3.3/bin/console
@@ -33,6 +33,6 @@ if ($_SERVER['APP_DEBUG']) {
     }
 }
 
-$kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
+$kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG'], dirname(__DIR__));
 $application = new Application($kernel);
 $application->run($input);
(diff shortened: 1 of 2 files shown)


tdutrion:
> Missing similar change in bin/console

I totally forgot this entrypoint... (though I'm also removing the `bin/console` in production 😄)
Would that be an edit in `symfony/console/3.3/bin/console` or should I create a `4.2` folder with that modification?

nicolas-grekas:
To me: 3.3

symfony/console/3.3/bin/console
```diff
@@ -33,6 +33,6 @@ if ($_SERVER['APP_DEBUG']) {
     }
 }
 
-$kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
+$kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG'], dirname(__DIR__));
```
stof: shouldn't it be part of a 4.3+ recipe only instead ?

tdutrion: Same for the other entrypoint then (`public/index.php`) I guess.

> To me: 3.3

That's been done acting on @nicolas-grekas's recommendation above. I'll let him give his own rationale 👍

nicolas-grekas: Duplicating all the files in the recipe is not worth it.

fabpot:
reverted

PR #556, opened 2019-03-25
Remove http&http-client aliases from php-http/httplug-pack

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

symfony/http-client will be the new default for `http-client` (and I don't think we need one for `http`)

### DIFF
--- a/php-http/httplug-pack/1.0/manifest.json
+++ b/php-http/httplug-pack/1.0/manifest.json
@@ -1,3 +1,3 @@
 {
-    "aliases": ["http", "http-client", "httplug"]
+    "aliases": ["httplug"]
 }


weaverryan:
@nyholm maybe you should give the final +1 - just to make sure there aren’t any docs in http plug that are referencing the alias

dbu:
we did explain that as preferred installation instructions: http://docs.php-http.org/en/latest/integrations/symfony-bundle.html#using-symfony-flex

what happens to people that used the flex recipe in the past? will they lose httplug on the next update?

can we make the recipe available again, if necessary with a more specific name? like psr/http-client ?

dbu:
ah, should have looked at the diff. its still available but only with the specific name httplug. all fine then.

PR #557, opened 2019-03-27
[Encore] Make @babel/preset-env use polyfills by default

Lyrkan:
| Q             | A
| ------------- | ---
| License       | MIT

This PR is related to https://github.com/symfony/webpack-encore/pull/545 and must not be merged until a new version of Encore is released (0.26.x).

Quick summary: we will soon set the `useBuiltIns` option of `@babel/preset-env` to `false` instead of `entry` by default (which makes it not handle `core-js` polyfills anymore).

This is needed because the other values (`entry` and `usage`) require the users to at least add `core-js` to their projects and specify its version in `@babel/preset-env`'s settings.

With this update the recipe will automatically require `core-js@3` and configure `@babel/preset-env` so polyfills are automatically added when/where they are needed (`useBuiltIns: 'usage'` and `corejs: 3`).

/cc @weaverryan

### DIFF
--- a/symfony/webpack-encore-bundle/1.0/package.json
+++ b/symfony/webpack-encore-bundle/1.0/package.json
@@ -1,6 +1,7 @@
 {
     "devDependencies": {
-        "@symfony/webpack-encore": "^0.22.0",
+        "@symfony/webpack-encore": "^0.26.0",
+        "core-js": "^3.0.0",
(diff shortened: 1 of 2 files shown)


weaverryan:
I hijacked this to also talk about `Encore.enableIntegrityHashes()`, which was also part of the same release

PR #558, opened 2019-03-30
Add a recipe for Symfony mailer

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/mailer/4.3/post-install.txt
+++ b/symfony/mailer/4.3/post-install.txt
@@ -1,0 +1,18 @@
+<bg=blue;fg=white>              </>
+<bg=blue;fg=white> What's next? </>
+<bg=blue;fg=white>              </>
+
+  * You're ready to send emails.
+
+  * If you want to send emails via a supported email provider, install
+    the corresponding bridge.
+    For instance, <info>composer require mailgun-bridge</> for Mailgun.
+
+  * If you want to send emails asynchronously:
+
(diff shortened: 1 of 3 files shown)


fabpot:
The issue is that there is no way right now to have an alias that points to a different recipe depending of the version of Symfony used.

fabpot:
It's even worse than that. After merge, `mailer` will always reference `symfony/mailer`.

fabpot:
Ready!

symfony/mailer/4.3/post-install.txt
```diff
@@ -0,0 +1,18 @@
...
+    For instance, <info>composer require mailgun-bridge</> for Mailgun.
+
+  * If you want to send emails asynchronously:
+
+    1. Install the messenger component by running <info>composer require messenger</>;
+    2. Add <info>'Symfony\Component\Mailer\Messenger\SendEmailMessage': amqp</> to the
+       <info>config/packages/messenger.yaml</> file under <info>framework.messenger.routing</>
+       and replace <info>amqp</> with your transport name of choice.
+
+  * <fg=blue>Read</> the documentation at <comment>https://symfony.com/doc/master/mailer.html</>
```
sroze: @fabpot this is not yet available though.

fabpot: I know :)

PR #559, opened 2019-03-31
[fw-bundle] Add support for validator auto-mapping

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

Enables https://github.com/symfony/symfony/pull/27735 by default.


symfony/framework-bundle/4.3/config/bootstrap.php
```diff
@@ -0,0 +1,21 @@
+<?php
```
teohhanhui: This file seems to be a duplicate of https://github.com/symfony/recipes/blob/2eafd3c75425318e461b95c1a1ed0c0b3b29f273/symfony/framework-bundle/4.2/config/bootstrap.php

nicolas-grekas: we can use a symlink here instead (as we do already)

symfony/framework-bundle/4.3/config/packages/framework.yaml
```diff
@@ -0,0 +1,24 @@
...
+    #esi: true
+    #fragments: true
+    php_errors:
+        log: true
+
+    validation:
+        # Enables validator auto-mapping support.
+        # For instance, basic validation constraints will be inferred from Doctrine's metadata.
+        auto_mapping:
+            App\Entity\: []
```
weaverryan: I keep thinking that this might make more sense as an annotation o the class not here. Currently, I look at my class to find all my constraints. Now I will possibly have these “silent” constraints I can’t track down. Could we activate the auto-constraints... with a constraint? We could generate entities with this by default with make:entity.

I know, wrong place to discuss - I just didn’t make it in time to the original Pr.

ostrolucky: This is a config which applies to all entities in given namespace, not specific class. How would you do that with annotation? Where would you put such annotation?

dunglas: This config supports setting full namespaces, as here, but also a list of specific classes using the glob syntax.
There is no annotations because it’s a bundle config. Only XML and YAML are supported. It’s similar to the service discovery config.

dunglas: And in almost all cases, it’s safer to always enable autovalidation. If it triggers a false positive, it’s most likely because the user is doing something very wrong that should be fixed.

dunglas:
Recipe updated to use symlinks. Is it ok to use symlinks for whole directories too? (It's why the check is red right now).

I tested it with 4.3 beta and DoctrineBundle master, and it works as expected.

nicolas-grekas:
> Is it ok to use symlinks for whole directories too? (It's why the check is red right now).

Yes conceptually, no technically apparently :) @fabpot can we adapt the check?

fabpot:
@symfony-flex-server review

fabpot:
@symfony-flex-server review

fabpot:
@dunglas @nicolas-grekas Flex server fixed, recipe is green, BUT you should test it to be sure that it works correctly.

dunglas:
On my machine (Mac OS 10.14), symlinked files are **not** created during `composer create-project`, even for the current stable version.

Reproducer:

```
$ composer create-project symfony/skeleton testsf42
$ ls -a testsf42/config/packages/
dev            framework.yaml routing.yaml   test
```

`cache.yaml` hasn't been generated.

nicolas-grekas:
I confirm the issue: a project created a few days ago contains `config/packages/cache.yaml` while a new one created now doesn't.

fabpot:
Problem with symlinked files is fixed now.

fabpot:
And this PR seems to work fine. @dunglas Can you confirm?

symfony/framework-bundle/4.3/config/packages/framework.yaml
```diff
@@ -0,0 +1,23 @@
...
+        handler_id: ~
+        cookie_secure: auto
+        cookie_samesite: lax
+
+    #esi: true
+    #fragments: true
+    php_errors:
+        log: true
+
+    validation:
```
fabpot: This must be commented by default as validation is not available on `symfony/skeleton`.

fabpot: Or even better why not move this to the `symfony/validator` recipe instead?

dunglas:
Good idea. It's where this config belongs. Updated.

dunglas:
And I confirm this PR works as expected!

PR #561, opened 2019-04-01
Change PHPUnit env to server

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

/cc @nicolas-grekas


fabpot:
Needed as we are checking `$_SERVER` first in Symfony. So, using `$_ENV` here means that we can be in a situation where `$_SERVER` has been set first to `dev` and the PHPUnit sets `$_ENV` to `test` but it has no effect.

PR #564, opened 2019-04-03
Add project dir to framework bundle

tdutrion:
| Q             | A
| ------------- | ---
| License       | MIT

References #555 and symfony/symfony#30651

This defines the project directory without looking for the `composer.json`.

### DIFF
--- a/symfony/framework-bundle/4.2/src/Kernel.php
+++ b/symfony/framework-bundle/4.2/src/Kernel.php
@@ -25,6 +25,11 @@ class Kernel extends BaseKernel
         }
     }
 
+    public function getProjectDir(): string
+    {
+        return \dirname(__DIR__);
+    }
+
     protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader): void
     {
         $container->addResource(new FileResource($this->getProjectDir().'/config/bundles.php'));


symfony/framework-bundle/4.2/src/Kernel.php
```diff
@@ -25,6 +25,11 @@ public function registerBundles(): iterable
         }
     }
 
+    public function getProjectDir(): string
+    {
+        return dirname(__DIR__);
```
nicolas-grekas: `\dirname(__DIR__);` (dirname is part if the short list, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3048)

tdutrion: Updated.

symfony/framework-bundle/4.2/src/Kernel.php
```diff
@@ -25,6 +25,11 @@ public function registerBundles(): iterable
         }
     }
 
+    public function getProjectDir(): string
+    {
+        return \dirname(__DIR__);
```
fabpot: The blackslash can be removed.

PR #565, opened 2019-04-06
Move DoctrineCacheBundle from recipes-contrib to recipes

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT

I think this makes sense since DoctrineCacheBundle is part of the orm-pack.
FYI @xabbuh 

#eu-fossa

See contrib PR: https://github.com/symfony/recipes-contrib/pull/652

### DIFF
--- a/doctrine/doctrine-cache-bundle/1.3/config/packages/doctrine_cache.yaml
+++ b/doctrine/doctrine-cache-bundle/1.3/config/packages/doctrine_cache.yaml
@@ -1,0 +1,10 @@
+doctrine_cache:
+    aliases:
+        apc_cache: my_apc_cache
+
+    providers:
+        my_apc_cache:
+            type: apc
+            namespace: my_apc_cache_ns
+            aliases:
+                - apc_cache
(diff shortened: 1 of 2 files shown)


chalasr:
I believe that DoctrineCacheBundle should be deprecated soon in favor of symfony/cache doctrine adapters, so we might just want to revert the recipes-contrib PR instead. /cc @alcaeus

Nyholm:
Oh, I see. Then we should do something differently here. 

Im closing this PR and I'll come up with a different solution. Thank you.

Nyholm:
Okey. I changed my mind. 
DoctrineCacheBundle is not part of the ORM-pack but a dependency of the DoctrineBundle 1.x. 

DoctrineBundle 2.0 will probably be released later this year. That means that we still have 6 months + SF4.4 LTS that will get the question: "Do you want to allow this recipe from contrib?" each time someone install orm-pack.

Nyholm:
Change my mind again. 

Andreas says there is never a scenario where a Symfony developer needs to configure DoctrineCacheBundle. We have a better cache solution using symfony/cache.

xabbuh:
see symfony/recipes#652

PR #566, opened 2019-04-06
[Webpack encore] Enable cache only in prod

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT

We made a mistake in #541 

We should only enable cache when `%kernel.debug%` is NOT true. 

#eu-fossa


Nyholm:
Thank you for merging

PR #567, opened 2019-04-06
[Symfony][Validator] Added option disable_not_compromised_password

lyrixx:
| Q             | A
| ------------- | ---
| License       | MIT

EUFOSSA

### DIFF
--- a/symfony/validator/4.1/config/packages/test/validator.yaml
+++ b/symfony/validator/4.1/config/packages/test/validator.yaml
@@ -1,0 +1,4 @@
+framework:
+    validation:
+        # As of Symfony 4.3 you can disable the NotCompromisedPassword Validator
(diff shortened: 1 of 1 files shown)


symfony/validator/4.1/config/packages/test/validator.yaml
```diff
@@ -0,0 +1,4 @@
+framework:
+    validation:
+        # As of Symfony 4.3 you can disable the NotCompromisedPassword Validator
+        # disable_not_compromised_password: true
```
fabpot: Can we instead create a 4.3 recipe with this option uncommented by default?

lyrixx: As disused IRL, we will keep it like this because for most application, this is not an issue to make a real test in test env

Issue #532, opened 2019-02-22
Unable to use getenv when using .env.local.php


toddcarver:
I'm not sure if this is an issue, so I apologize in advance if I am mistaken.  We just upgraded to 4.2 and have implemented config/bootstrap.php as recommended here:

https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/4.2/config/bootstrap.php

We have created .env.local.php so the block below is evaulated:

```
if (is_array($env = @include dirname(__DIR__).'/.env.local.php')) {
    $_SERVER += $env;
    $_ENV += $env;
}
```

The problem is that all our calls using getenv are broken.  After researching I believe this is because the env vars are not loaded via Dotenv, which adds the env vars to $_ENV and by using putenv. Using getenv works if the following block runs:

```
} else {
    // load all the .env files
    (new Dotenv())->loadEnv(dirname(__DIR__).'/.env');
}
```

Any help you can offer with a recommendation to resolve this would be greatly appreciated.  Thank you.

nicolas-grekas:
Hum, yes, that's a limitation of using the compiled
.env file.
My recommendation is to not use `getenv()`: it's not thread safe so using it can break randomly on some machines, it's slower, and DI configuration is more flexible...

toddcarver:
Thanks, Nicolas. I will close the issue.

nicolas-grekas:
> can't we merge the values stored in $_SERVER with the values dumped in the .env.local.php?

that's exactly what should happen already, dumped values should always be defaults and should never override existing real vars. Please provide a reproduce if that's not what you experience.

benja-M-1:
I dove a little bit more with this php script

```php 
<?php

$key = 'foo';
$value = 'bar';
$_SERVER[$key] = $value;

if (is_array($env = include __DIR__.'/.env.local.php')) {
    $_SERVER += $env;
    $_ENV += $env;
}

var_dump([count($_ENV), count($_SERVER), getenv($key, true)]);
```
and the following value of `.env.local.php`

```php
<?php

return array (
  'APP_ENV' => 'baz',
);
```
Here is the result:
![image](https://user-images.githubusercontent.com/212269/55826796-119f5a00-5b09-11e9-84fb-329f252af89d.png)

It seems that `$_ENV` is empty at start and thus `getenv('foo)` , which I think  `EnvVarProcessor::getEnv` uses, returns `baz`. That's why I have my issue.

[...]

PR #571, opened 2019-04-10
Populate $_SERVER with $_ENV

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Reverts #540, fixes https://github.com/symfony/recipes/issues/532#issuecomment-481302280


MisatoTremor:
While you are right about thread safety this again introduces the problem that a system behaves different after calling ``composer dump-env prod``.

The better solution would be to remove putenv from Dotenv, too. But that's not how it currently is.

nicolas-grekas:
> The better solution would be to remove putenv from Dotenv, too. But that's not how it currently is.

I agree with that but indeed it would be quite a BC break.
Alternatively, you should ensure to always use `$_SERVER` and never `getenv()`. It's broken anyway.

MisatoTremor:
Yep, unfortunately it is in the Symfony doc prior to 4.2, so it might be out in the wild.
Should therefor mainly apply to the 3.3 version of the receipe ... whiiich I coincidentally overlooked in my PR :-\

PR #576, opened 2019-04-17
[symfony/security-bundle] set encoder to "auto" for all UserInterface by default

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Needs https://github.com/symfony/symfony/pull/31140

ping @weaverryan WDYT?

### DIFF
--- a/symfony/security-bundle/4.3/config/packages/security.yaml
+++ b/symfony/security-bundle/4.3/config/packages/security.yaml
@@ -1,0 +1,28 @@
+security:
+    # https://symfony.com/doc/current/security.html#where-do-users-come-from-user-providers
+    providers:
+        in_memory: { memory: ~ }
+    encoders:
+        Symfony\Component\Security\Core\User\UserInterface:
+            algorithm: auto
+    firewalls:
+        dev:
+            pattern: ^/(_(profiler|wdt)|css|images|js)/
+            security: false
+        main:
+            anonymous: true
+
+            # activate different ways to authenticate
+
+            # http_basic: true
+            # https://symfony.com/doc/current/security.html#a-configuring-how-your-users-will-authenticate
+
+            # form_login: true
+            # https://symfony.com/doc/current/security/form_login_setup.html
+
+    # Easy way to control access for large sections of your site
+    # Note: Only the *first* access control that matches will be used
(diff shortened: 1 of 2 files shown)


symfony/security-bundle/4.3/config/packages/security.yaml
```diff
@@ -0,0 +1,28 @@
+security:
+    # https://symfony.com/doc/current/security.html#where-do-users-come-from-user-providers
+    providers:
+        in_memory: { memory: ~ }
+    encoders:
+        Symfony\Component\Security\Core\User\UserInterface:
+            algorithm: auto
```
weaverryan:
```yml
# configure password encoding logic (if your app needs to encode passwords)
encoders:
    # Any User class will automatically use the best-available password hashing algorithm
    Symfony\Component\Security\Core\User\UserInterface:
        algorithm: auto
```

This requires a bit more explanation than I like. First, adding this section for everyone, even if it's pretty common not to need a hashing algo, isn't ideal. But, I think it's not *too* big of a deal.

The bigger thing is that the `Symfony\Component\Security\Core\User\UserInterface` config isn't immediately obvious... it's not terrible, but I need to explain to you that "Hey! All classes implement this interface, and because this is a class-based map that respects interfaces, all user classes will use this algorithm". Of course... the user might think "what do you mean about all user classES, should I have multiple"?

A few possible ideas:
1) Add a special `default` key under encoders

and/or

2) Let MakerBundle / the docs add this section for the user. I don't think causes a real security issue, as MakerBundle and the docs will recommend auto. 

and my unrelated point (3)... rename `encoders` to `password_encoders` ;)

stof: well, if we do a renaming in 5.0, we should actually move away from `encoder`. We are hashing password hopefully, not encoding them.

weaverryan: Probably better to leave it now, properly think about things for 4.4/5.0

nicolas-grekas:
Renaming encoders to something else is another story someone needs to pursue.

Thanks for the feedback: maker + doc will do the trick, no need for the recipe. I like it :)

PR #578, opened 2019-04-23
move environment variable resolving to simple-phpunit script

IonBazan:
| Q             | A
| ------------- | ---
| License       | MIT

This fixes #561 by moving all the environment variable resolving to `simple-phpunit` script from `symfony/phpunit-bridge` package. This allows users to specify PHPUnit version in `phpunit.xml.dist` file, without actually setting the env variable.


symfony/phpunit-bridge/3.3/phpunit.xml.dist
```diff
@@ -11,6 +11,8 @@
         <ini name="error_reporting" value="-1" />
         <server name="APP_ENV" value="test" force="true" />
         <server name="SHELL_VERBOSITY" value="-1" />
+        <server name="SYMFONY_PHPUNIT_VERSION" value="6.5" />
```
IonBazan: @fabpot @nicolas-grekas I am not really sure if we should use `server` or `env` node here.

nicolas-grekas:
Damn, I just submitted https://github.com/symfony/recipes/pull/579 :)
Looking at the attached patch, I think we should close this one.
See also https://github.com/symfony/symfony/pull/31210

IonBazan:
@nicolas-grekas You are right - 3.3 should not be affected by this change.

PR #580, opened 2019-04-25
Improve SymfonyFrameworkBundle Kernel.php

yyaremenko:
| Q             | A
| ------------- | ---
| License       | MIT

This commit makes Kernel.php look more professional for v4.2. If it's okay for you, I will update the Kernel.php for v3.3 in the same manner.

### DIFF
--- a/symfony/framework-bundle/4.2/src/Kernel.php
+++ b/symfony/framework-bundle/4.2/src/Kernel.php
@@ -13,11 +13,13 @@ class Kernel extends BaseKernel
 {
     use MicroKernelTrait;
 
-    private const CONFIG_EXTS = '.{php,xml,yaml,yml}';
+    private const CONFIGS_POSSIBLE_EXTENSIONS = '.{php,xml,yaml,yml}';
+
+    private const RESOURCE_TYPE_GLOBAL = 'glob';
 
     public function registerBundles(): iterable
     {
-        $contents = require $this->getProjectDir().'/config/bundles.php';
+        $contents = require $this->getBundlesFilePath();
         foreach ($contents as $class => $envs) {
             if ($envs[$this->environment] ?? $envs['all'] ?? false) {
                 yield new $class();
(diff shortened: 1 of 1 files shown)


B-Galati:
There is no such thing called professional code. Just choices to be made in a given context. 

IMHO this proposal just make things more complicated that they need to be for no good reason.

yyaremenko:
No, there **is** such thing called professional code, regardless of programming language. Like, using constants with explanatory names for repetitive strings, avoiding copy-and-paste (keeping to DRY principle), making one method doing one thing only etc. But hey, you are the contributor and, I guess, you have the better vision of the code.

Thanks for the reply and have a nice day.

yyaremenko:
(you are welcome to decline the PR)
thanks.

B-Galati:
Best practices must not be blindly applied. There is a difference between DRY and copy/paste.

Please read this great article: https://web-techno.net/dry-principle-explained/

Regarding your PR, I do not decide anything, I gave my opinion because I wanted to.

yyaremenko:
Thank you for the clarification, I am not quite aware of PR decision makers for this repository. Also, thank you for the article. Still I am afraid I don't quite understand how creation of unneeded abstract classes relates to the changes I offer, as my changes have nothing common with the problem described.

yyaremenko:
> this doesn't qualify as an improvement, it's just a different opinion.

1. renaming a constant from CONFIG_EXTS to a more descriptive name CONFIGS_POSSIBLE_EXTENSIONS 

is a different opinion, not a code improvement, no way

2. moving repetitively used string 'glob' to a constant with a descriptive name RESOURCE_TYPE_GLOBAL

the same purpose constant CONFIG_EXTS was fine though, but the constant I added is a 'different opinion', okay

3. instead of calling $this->getProjectDir().'/config/bundles.php' twice, a dedicated method returning this path is created getBundlesFilePath()

so, using a single entry point instead of copy-and-pasting is a 'different opinion', okay

4. instead of calling $this->getProjectDir().'/config' twice, a dedicated method returning the path is created

single entry point instead of copy-and-pasting is a 'different opinion', yeah, yeah

5. instead of copy-and-pasting code signature as follows
ob->doSomething(x, valueA, y, z);
ob->doSomething(x, valueB, y, z);
ob->doSomething(x, valueC, y, z);

a dedicated method was created, iterating over vallueA, valueB, valueC - this way, if a signature of
ob->doSomething(x, vallue, y, z) changes, you'll only have to change it in one place, not 3 or more...

but hey, that's a 'different opinion', let's just do copy-and-paste.

well, what can I say. Symfony positions itself as a high-quality framework, but what I see in code now and there tells a different story.  I am shocked I have to explain / advocate simple, obvious things - yet without any success...

as a senior developer, who does a lot code reviews daily, I would never ever allow the original code into production.

okay now, I'm fed enough and really tired advocating myself. keep on with your great code, folks!

PR #582, opened 2019-04-27
Update post-install.txt

stephanvierkant:
| Q             | A
| ------------- | ---
| License       | MIT

Since `.env` is supposed to be committed to the shared repository (not "gitignored"), you shouldn't change `.env` but only your local environment variables, as per https://symfony.com/doc/current/configuration/dot-env-changes.html

### DIFF
--- a/doctrine/doctrine-bundle/1.6/post-install.txt
+++ b/doctrine/doctrine-bundle/1.6/post-install.txt
@@ -2,7 +2,7 @@
 <bg=blue;fg=white> Database Configuration </>
 <bg=blue;fg=white>                        </>
 
-  * Modify your DATABASE_URL config in <fg=green>.env</>
+  * Modify your DATABASE_URL config in <fg=green>.env.local</>
 
   * Configure the <fg=green>driver</> (mysql) and
     <fg=green>server_version</> (5.7) in <fg=green>config/packages/doctrine.yaml</>


doctrine/doctrine-bundle/1.6/post-install.txt
```diff
@@ -2,7 +2,7 @@
 <bg=blue;fg=white> Database Configuration </>
 <bg=blue;fg=white>                        </>
 
-  * Modify your DATABASE_URL config in <fg=green>.env</>
+  * Modify your DATABASE_URL config in <fg=green>.env.local</>
```
weaverryan: Yes, but there may *not* be a `.env.local` file. Perhaps we should add a `.env.local` to the `symfony/flex` recipe?

Pierstoval: And add a comment in it to explain how to use it with custom and platform-specific changes 👍

dunglas:
I would keep this message as is (and not introduce a `.env.local` file), because:

* the referenced file already contains a detailed explanation of environments work: https://github.com/symfony/recipes/blob/master/symfony/flex/1.0/.env#L4-L7
* it's up to the user to update the appropriate file (for instance, Docker and VM users will edit `.env` directly, while users installing the project locally on a host machine will prefer using `.env.local`)

nicolas-grekas:
The file that needs to be changed is *really* `.env`, especially after https://github.com/symfony/recipes/pull/674
Thanks for proposing.

PR #583, opened 2019-04-28
Add Docker support for the Mercure recipe

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

Needs https://github.com/symfony/flex/pull/128

### DIFF
--- a/symfony/mercure-bundle/0.1/manifest.json
+++ b/symfony/mercure-bundle/0.1/manifest.json
@@ -10,8 +10,25 @@
         "env(MERCURE_JWT_SECRET)": ""
     },
     "env": {
-        "MERCURE_PUBLISH_URL": "https://demo.mercure.rocks/hub",
-        "MERCURE_JWT_SECRET": "#see https://mercure.rocks"
+        "#1": "See https://symfony.com/doc/current/mercure.html#configuration",
+        "MERCURE_PUBLISH_URL": "http://mercure/hub",
+        "MERCURE_JWT_SECRET": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJtZXJjdXJlIjp7InB1Ymxpc2giOltdfX0.Oo0yg7y4yMa1vr_bziltxuTCqb8JVHKxp-f_FwwOim0"
+    },
+    "docker-compose": {
+        "services": [
+            "mercure:",
+            "  # In production, you may want to use the managed version of Mercure, https://mercure.rocks",
+            "  image: dunglas/mercure",
+            "  environment:",
+            "    # You should definitely change all these values in production",
+            "    - JWT_KEY=!ChangeMe!",
+            "    - ALLOW_ANONYMOUS=1",
+            "    - CORS_ALLOWED_ORIGINS=*",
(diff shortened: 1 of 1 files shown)


symfony/mercure-bundle/0.1/manifest.json
```diff
@@ -10,8 +10,25 @@
...
+            "  image: dunglas/mercure",
+            "  environment:",
+            "    # You should definitely change all these values in production",
+            "    - JWT_KEY=!ChangeMe!",
+            "    - ALLOW_ANONYMOUS=1",
+            "    - CORS_ALLOWED_ORIGINS=*",
+            "    - PUBLISH_ALLOWED_ORIGINS=http://localhost:1337",
+            "    - DEMO=1",
+            "  ports:",
+            "    - \"1337:80\""
```
fabpot: What about letting Docker choose the local port?

Pierstoval: Random port forces the end user to change any env var related to the configuration urls in `.env.local`, especially if container is recreated with a new random port, because I'm not sure any Docker or Compose configuration could propagate the random value to any env var. I'm not sure it's the best solution for DX.

If anyone is working on multiple projects in parallel, my personal recommendation would be to `docker-compose down/stop` before leaving the previous project, and `docker-compose up -d` on the next project to work on, as a measure of safety for ports.

fabpot: See https://github.com/symfony/flex/pull/533#discussion_r309559954

dunglas: It’s not really practical for Mercure because because the client (the JS app) must know the port to use. And the JS app will run on the host.

fabpot: ok

symfony/mercure-bundle/0.1/manifest.json
```diff
@@ -10,8 +10,25 @@
         "env(MERCURE_JWT_SECRET)": ""
     },
     "env": {
-        "MERCURE_PUBLISH_URL": "https://demo.mercure.rocks/hub",
-        "MERCURE_JWT_SECRET": "#see https://mercure.rocks"
+        "#1": "See https://symfony.com/doc/current/mercure.html#configuration",
+        "MERCURE_PUBLISH_URL": "http://mercure/hub",
+        "MERCURE_JWT_SECRET": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJtZXJjdXJlIjp7InB1Ymxpc2giOltdfX0.Oo0yg7y4yMa1vr_bziltxuTCqb8JVHKxp-f_FwwOim0"
```
teohhanhui: I don't know why this was never mentioned, but apparently the secret used here is:

`!ChangeMe!`

dunglas: It's mentioned here: https://github.com/symfony/recipes/pull/583/files#diff-d19ccce8daab2b00dbfa4e827ebde212R24
But indeed it must be added as a comment in `.env`. Can you open a PR please?

PR #584, opened 2019-05-05
[Bridge/PhpUnit] Require .php file to fix problem with extra shebang

JustBlackBird:
This commit is a part of fix for https://github.com/symfony/symfony/issues/27035

The other part is https://github.com/symfony/symfony/pull/31364

| Q             | A
| ------------- | ---
| License       | MIT


symfony/phpunit-bridge/4.3/bin/phpunit
```diff
@@ -0,0 +1,19 @@
...
+    putenv('SYMFONY_PHPUNIT_VERSION=6.5');
+}
+if (false === getenv('SYMFONY_PHPUNIT_REMOVE')) {
+    putenv('SYMFONY_PHPUNIT_REMOVE=');
+}
+if (false === getenv('SYMFONY_PHPUNIT_DIR')) {
+    putenv('SYMFONY_PHPUNIT_DIR='.__DIR__.'/.phpunit');
+}
+
+require dirname(__DIR__).'/vendor/symfony/phpunit-bridge/bin/simple-phpunit.php';
```
nicolas-grekas: instead of creating a new version of the recipe, what about doing a file_exists check here in the 4.1 recipe?

JustBlackBird: I can do so  to but don't think it's needed.

There is no BC break in Bridge/PhpUnit 4.3 so the check is only decreases number of recipe versions but make the code less cleaner. I prefer cleaner code :)

nicolas-grekas: From a maintenance perspective it would make our life easier :)

JustBlackBird: Fair enough :)

Updated 4.1 recipe

nicolas-grekas: Looking at the resulting code, I decide to follow your initial proposal and add a recipe for 4.3.
I created symlinks to 4.1 files.

JustBlackBird: Will flex actually work fine with simlinks? :smile:

nicolas-grekas: It does already ;)

PR #585, opened 2019-05-07
[FrameworkBundle] Load packages configuration after services configuration

lyrixx:
Some people tends to declare some services in `/config/package/foobar.yaml`.
More over, some recipies does that [too](https://github.com/symfony/recipes/tree/master/sensiolabs/security-checker/4.0/config/packages)

But with the default configuration, services are overriden in
`services.yaml` (auto-discovery) and so the services does not work.

This patch made the packages directory prioritary overy the services
files

| Q             | A
| ------------- | ---
| License       | MIT

---

I found this "issue" by helping someone on slack.

### DIFF
--- a/symfony/framework-bundle/4.2/src/Kernel.php
+++ b/symfony/framework-bundle/4.2/src/Kernel.php
@@ -36,10 +36,10 @@ class Kernel extends BaseKernel
         $container->setParameter('container.dumper.inline_class_loader', true);
         $confDir = $this->getProjectDir().'/config';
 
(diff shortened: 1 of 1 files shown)


fabpot:
I think that the current configuration is the right one. `services.yaml` is your file, so it should override the "default" configuration coming from `packages/`

nicolas-grekas:
I agree with Fabien: packages configures something in vendor/ services.yaml is for src/
The linked recipe does follow this rule too.

lyrixx:
I agree with you too.

I just wanted your feedback :)

PR #587, opened 2019-05-10
Add recipes from third-party mailer transports

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

I've not added `gmail` as it should not be used in production (so let's avoid it here).

### DIFF
--- a/symfony/amazon-mailer/4.3/manifest.json
+++ b/symfony/amazon-mailer/4.3/manifest.json
@@ -1,0 +1,6 @@
+{
+    "env": {
+        "#1": "MAILER_DSN=smtp://$AWS_ACCESS_KEY:$AWS_SECRET_KEY@ses",
+        "#2": "MAILER_DSN=http://$AWS_ACCESS_KEY:$AWS_SECRET_KEY@ses"
+    }
+}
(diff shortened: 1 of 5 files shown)


symfony/amazon-mailer/4.3/manifest.json
```diff
@@ -0,0 +1,7 @@
+{
+    "env": {
+        "#1": "MAILER_DSN=smtp://$AWS_ACCESS_KEY:$AWS_SECRET_KEY@ses",
+        "#2": "MAILER_DSN=http://$AWS_ACCESS_KEY:$AWS_SECRET_KEY@ses",
```
OskarStark: Shouldn't we use `httpS` if possible?

fabpot: It **does** use httpS behind the scene, here `http` is not the "real" protocol, but merely a way to choose which interface to use.

OskarStark: Ok, I didn't know this 👍

fabpot:
I've removed the API URLs as we should not recommend using them.

fabpot:
merge now

bocharsky-bw:
> I've not added `gmail` as it should not be used in production (so let's avoid it here).

@fabpot recipes are not for configuring production only. If we don't have a recipe for this does not mean devs won't use it on production. It just looks like an incomplete/missing piece here.

What about to add a recipe for Gmail but with a comment explaining that devs should not use it on production and why? I think it would much better than do not have a recipe at all for this. WDYT?

PR #590, opened 2019-05-21
Update messenger

maxhelias:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/messenger/4.3/config/packages/messenger.yaml
+++ b/symfony/messenger/4.3/config/packages/messenger.yaml
@@ -1,0 +1,14 @@
+framework:
+    messenger:
+        # Uncomment this (and the failed transport below) to send failed messages to this transport for later handling.
+        # failure_transport: failed
+
+        transports:
+            # https://symfony.com/doc/current/messenger.html#transports
+            # async: '%env(MESSENGER_TRANSPORT_DSN)%'
+            # failed: 'doctrine://default?queue_name=failed'
+            # sync: 'sync://'
+
+        routing:
+            # Route your messages to the transports
+            # 'App\Message\YourMessage': [async, sync]
--- a/symfony/messenger/4.3/manifest.json
+++ b/symfony/messenger/4.3/manifest.json
@@ -1,0 +1,14 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "#1": "To use AMQP, also run ‘composer require amqp’",
+        "#2": "MESSENGER_TRANSPORT_DSN=amqp://guest:guest@localhost:5672/%2f/messages",
+        "#3": "MESSENGER_TRANSPORT_DSN=redis://localhost:6379/messages"
+    },
+    "aliases": ["messenger"],
+    "conflict": {
+        "symfony/framework-bundle": "<4.3"
+    }
+}
--- a/symfony/messenger/4.3/post-install.txt
+++ b/symfony/messenger/4.3/post-install.txt
@@ -1,0 +1,15 @@
+<bg=blue;fg=white>              </>
+<bg=blue;fg=white> What's next? </>
+<bg=blue;fg=white>              </>
+
+  * You're ready to use the Messenger component. You can define your own message buses
+    or start using the default one right now by injecting the <info>messenger.bus.default</info> service
+    or typehinting <info>Symfony\Component\Messenger\MessageBusInterface</info> in your code.
+
+  * To send messages to a transport and handle them asynchronously:
+
+    1. Uncomment the <comment>MESSENGER_TRANSPORT_DSN</> env var in .env
+       and <comment>framework.messenger.transports.async</> in config/packages/messenger.yaml;
+    2. Route your message classes to the async transport in config/packages/messenger.yaml.
+
+  * <fg=blue>Read</> the documentation at <comment>https://symfony.com/doc/master/messenger.html</>


symfony/messenger/4.3/config/packages/messenger.yaml
```diff
@@ -0,0 +1,13 @@
+framework:
+    messenger:
+        # Uncomment the following line and choose a transports for collect all messages failed and try it later
```
weaverryan: > transport to collect messages that fail so you can try them later

symfony/messenger/4.3/config/packages/messenger.yaml
```diff
@@ -0,0 +1,13 @@
...
+        # failure_transport: ...
+
+        transports:
+            # Uncomment the following lines to enable one of the transports named "amqp" or "sync"
```
weaverryan: Instead of the comment, I think we should just link to `https://symfony.com/doc/current/messenger.html#transports`

symfony/messenger/4.3/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "#1": "MESSENGER_TRANSPORT_DSN=amqp://guest:guest@localhost:5672/%2f/messages"
```
weaverryan: Now that we have multiple good transports (doctrine and redis), I think it makes sene to give one example for each (amqp, doctrine, redis) so the user can choose

symfony/messenger/4.3/config/packages/messenger.yaml
```diff
@@ -0,0 +1,13 @@
...
+            # amqp: '%env(MESSENGER_TRANSPORT_DSN)%'
```
weaverryan: What about calling this `async`? It's not related to amqp anymore. What I mean is: amqp is now just *one* option for your transport

symfony/messenger/4.3/post-install.txt
```diff
@@ -0,0 +1,19 @@
+<bg=blue;fg=white>              </>
+<bg=blue;fg=white> What's next? </>
+<bg=blue;fg=white>              </>
+
+  * You're ready to use the Messenger component. You can define your own message buses
+    or start using the default one right now by injecting the <info>messenger.bus.default</info> service
+    or typehinting <info>Symfony\Component\Messenger\MessageBusInterface</info> in your code.
+
+  * If you need to send messages to your broker, you can benefit from the built-in
+    AMQP transport by:
```
weaverryan: To send messages to a transport and handle them asynchronously:

symfony/messenger/4.3/post-install.txt
```diff
@@ -0,0 +1,19 @@
...
+
+    1. Installing the AMQP pack by running <info>composer require amqp</>;
+    2. Uncommenting the <comment>MESSENGER_TRANSPORT_DSN</> env var
+       and <comment>framework.messenger.transports.amqp</> config;
```
weaverryan: > Uncomment the <comment>MESSENGER_TRANSPORT_DSN</> env var in <comment>.env</comment>
       and <comment>framework.messenger.transports.amqp</> in <comment>config/packages/messenger.yaml</comment>

symfony/messenger/4.3/post-install.txt
```diff
@@ -0,0 +1,19 @@
...
+<bg=blue;fg=white>              </>
+
+  * You're ready to use the Messenger component. You can define your own message buses
+    or start using the default one right now by injecting the <info>messenger.bus.default</info> service
+    or typehinting <info>Symfony\Component\Messenger\MessageBusInterface</info> in your code.
+
+  * If you need to send messages to your broker, you can benefit from the built-in
+    AMQP transport by:
+
+    1. Installing the AMQP pack by running <info>composer require amqp</>;
```
weaverryan: Probably remove this - and add it as a "detail" in `.env` above the `amqp` example. We're making a lot of changes, so we'll need to play with the final result to see how it feels

symfony/messenger/4.3/config/packages/messenger.yaml
maxhelias: I like it, we will also have to update the documentation about it

symfony/messenger/4.3/post-install.txt
maxhelias: I wonder if we should not rename the pack too? Or depreciate it ?
Because it includes the serializer pack that is also useful for other transport...

symfony/messenger/4.3/config/packages/messenger.yaml
```diff
@@ -0,0 +1,13 @@
+framework:
+    messenger:
+        # Uncomment the following line and choose a transport to collect messages that fail so you can try them later
```
OskarStark: s/try/retry/ ?

symfony/messenger/4.3/config/packages/messenger.yaml
```diff
@@ -0,0 +1,13 @@
+framework:
+    messenger:
+        # Uncomment the following line and choose a transport to collect messages that fail so you can retry them later
+        # failure_transport: ...
```
fabpot: What about giving an example with Doctrine here?

maxhelias: There is an example in the `manifest.json`, but I can add a "failed" transport that will use Doctrine ?

fabpot: That's what I meant, having a failed transport using Doctrine (which I suppose will be the most common configuration).

weaverryan: Let’s try it. So we would add a transport, maybe called “failed” and show it using a doctrine dsn - maybe also with ?queue=failed (so that it doesn’t collide with a real transport of the user also uses doctrine for that). Double check that config is correct. Then, this line (still commented), would be failure_transport: failed

maxhelias: @weaverryan, I updated my PR just at the same time as your comment. What do you think of my suggestion with a new env var `MESSENGER__FAILED_TRANSPORT_DSN ` which contains the configuration ?

symfony/messenger/4.3/config/packages/messenger.yaml
```diff
@@ -0,0 +1,14 @@
+framework:
+    messenger:
+        # Uncomment the following line and choose a transport to collect messages that fail so you can retry them later
+        # failure_transport: failed
+
+        transports:
+            # https://symfony.com/doc/current/messenger.html#transports
+            # async: '%env(MESSENGER_TRANSPORT_DSN)%'
+            # failed: '%env(MESSENGER__FAILED_TRANSPORT_DSN)%'
```
weaverryan: Someone might disagree, but I don’t think we need a separate environment variable for this. Rather, I think we should literally use doctrine:///default?queue_name=default

We’ll document that you could also reuse the main env var here and pass other options, like if you wanted to reuse your Amqp config, but use a different exchange for failed messages.

symfony/messenger/4.3/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "#1": "For use AMQP install his pack by running composer require amqp",
+        "#2": "MESSENGER_TRANSPORT_DSN=amqp://guest:guest@localhost:5672/%2f/messages",
+        "#3": "MESSENGER_TRANSPORT_DSN=redis://localhost:6379/messages",
+        "#4": "MESSENGER__FAILED_TRANSPORT_DSN=doctrine://default?queue_name=failed"
```
weaverryan: If we remove this env var, this last example can go back to showing how to use doctrine for the main env var

symfony/messenger/4.3/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "#1": "For use AMQP install his pack by running composer require amqp",
```
weaverryan: > To use AMQP, also run ‘composer require amqp’

symfony/messenger/4.3/post-install.txt
```diff
@@ -0,0 +1,18 @@
+<bg=blue;fg=white>              </>
+<bg=blue;fg=white> What's next? </>
+<bg=blue;fg=white>              </>
+
+  * You're ready to use the Messenger component. You can define your own message buses
+    or start using the default one right now by injecting the <info>messenger.bus.default</info> service
+    or typehinting <info>Symfony\Component\Messenger\MessageBusInterface</info> in your code.
+
+  * If you need to send messages to your broker, you can benefit from the built-in
```
weaverryan: This line should be removed

symfony/messenger/4.3/post-install.txt
```diff
@@ -0,0 +1,18 @@
...
+    To send messages to a transport and handle them asynchronously:
+
+    1. Uncomment the <comment>MESSENGER_TRANSPORT_DSN</> env var in .env
+       and <comment>framework.messenger.transports.async</> in config/packages/messenger.yaml;
+    2. Routing your messages to the async sender.
```
weaverryan: Route your message classes to the async transport in config/packages/messenger.yaml

symfony/messenger/4.3/post-install.txt
```diff
@@ -0,0 +1,18 @@
...
+
+  * If you need to send and handle your messages, uncomment the sync transport and add it on your routing
```
weaverryan: I think we don’t need this. Messages are “sync” by default - you don’t need to do anything to make them sync. The sync transport is just handy if you want to be explicit, or maybe send a message to an async transport AND also handle it immediately.

symfony/messenger/4.3/config/packages/messenger.yaml
```diff
@@ -0,0 +1,14 @@
+framework:
+    messenger:
+        # Uncomment the following line and choose a transport to collect messages that fail so you can retry them later
```
weaverryan: We can now tweak this:

Uncomment this (and the failed transport below) to send failed messages to this transport for later handling.

symfony/messenger/4.3/config/packages/messenger.yaml
fabpot: I agree

PR #591, opened 2019-05-22
[WebpackEncoreBundle] Don't use kernel.debug for the cache parameter

pierredup:
| Q             | A
| ------------- | ---
| License       | MIT

When using the `%kernel.debug%` value for the `cache` parameter means the entrypoints will be cached when debug is true, where actually it should be the reverse. This currently causes issues in development mode, especially when using integrity hashes, since the integrity hash is cached with the entrypoints and it won't match the content of the assets if they changed. So I think this value should by default be set to `false`, where the developer can then decide if they want to enable it in production (an alternative option is to create another config file in the `prod` folder where this value defaults to `true`)

### DIFF
--- a/symfony/webpack-encore-bundle/1.0/config/packages/webpack_encore.yaml
+++ b/symfony/webpack-encore-bundle/1.0/config/packages/webpack_encore.yaml
@@ -10,4 +10,4 @@ webpack_encore:
     
     # Cache the entrypoints.json (rebuild Symfony's cache when entrypoints.json changes)
     # Available in version 1.2
-    #cache: '%kernel.debug%'
+    #cache: 'false'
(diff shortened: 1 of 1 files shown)


pierredup:
I've added a note about enabling cache for prod. If you're happy with the wording etc then I'll open the PR to fix the readme in the bundle

fabpot:
@weaverryan Why not creating a specific file for prod?

weaverryan:
> @weaverryan Why not creating a specific file for prod?

It's such a *minor* performance improvement... and requires users to run Encore before running cache:warmup (so maybe it would need to be commented out). In my mind, I'm not a huge fan of adding more files... but maybe I'm being over sensitive.

Let's merge this now, then maybe someone (@pierredup) could propose that file in another PR

PR #592, opened 2019-05-24
Tweaks to messenger 4.3 recipe

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

Minor tweaks while playing with the Messenger documentation today :). 

Cheers!

### DIFF
--- a/symfony/messenger/4.3/config/packages/messenger.yaml
+++ b/symfony/messenger/4.3/config/packages/messenger.yaml
@@ -11,4 +11,4 @@ framework:
 
         routing:
             # Route your messages to the transports
-            # 'App\Message\YourMessage': [async, sync]
+            # 'App\Message\YourMessage': async
--- a/symfony/messenger/4.3/manifest.json
+++ b/symfony/messenger/4.3/manifest.json
@@ -3,9 +3,10 @@
         "config/": "%CONFIG_DIR%/"
     },
     "env": {
-        "#1": "To use AMQP, also run ‘composer require amqp’",
+        "#1": "Choose one of the transports below",
         "#2": "MESSENGER_TRANSPORT_DSN=amqp://guest:guest@localhost:5672/%2f/messages",
-        "#3": "MESSENGER_TRANSPORT_DSN=redis://localhost:6379/messages"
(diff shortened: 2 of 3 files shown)


symfony/messenger/4.3/config/packages/messenger.yaml
```diff
@@ -11,4 +11,4 @@ framework:
 
         routing:
             # Route your messages to the transports
-            # 'App\Message\YourMessage': [async, sync]
+            # 'App\Message\YourMessage': async
```
weaverryan: This is a more realistic use-case

symfony/messenger/4.3/manifest.json
```diff
@@ -3,9 +3,10 @@
         "config/": "%CONFIG_DIR%/"
     },
     "env": {
-        "#1": "To use AMQP, also run ‘composer require amqp’",
```
weaverryan: I think the amqp-pack might not be necessary anymore. It installs messenger, serializer & makes sure the amqp extension is installed. In this case, we already know messenger is installed and the serializer is no longer needed. That leaves the pack as a way to give the user an error about the amqp extension if it's not installed.

symfony/messenger/4.3/post-install.txt
```diff
@@ -3,13 +3,13 @@
 <bg=blue;fg=white>              </>
 
   * You're ready to use the Messenger component. You can define your own message buses
-    or start using the default one right now by injecting the <info>messenger.bus.default</info> service
-    or typehinting <info>Symfony\Component\Messenger\MessageBusInterface</info> in your code.
+    or start using the default one right now by injecting the <info>message_bus</info> service
```
weaverryan: `message_bus` is the alias of the default bus

symfony/messenger/4.3/post-install.txt
```diff
@@ -3,13 +3,13 @@
...
+    or type-hinting <info>Symfony\Component\Messenger\MessageBusInterface</info> in your code.
 
   * To send messages to a transport and handle them asynchronously:
```
Tobion: This seems to be missing an important step to start the worker.  Otherwise they don't get handled as the description says.

PR #594, opened 2019-05-27
Use API scheme instead of SMTP by default for SendGrid

bocharsky-bw:
| Q             | A
| ------------- | ---
| License       | MIT

SendGrid recommends using API instead of SMTP:

Screen Shot 2019-05-27 at 18 52 28

Probably it makes sense to tweak this recipe?

### DIFF
--- a/symfony/sendgrid-mailer/4.3/manifest.json
+++ b/symfony/sendgrid-mailer/4.3/manifest.json
@@ -1,5 +1,5 @@
 {
     "env": {
-        "#1": "MAILER_DSN=smtp://$SENDGRID_KEY@sendgrid"
+        "#1": "MAILER_DSN=api://$SENDGRID_KEY@sendgrid"
     }
 }


fabpot:
We've (I have really) decided to only use SMTP or HTTP for all providers as this is how we can be. sure that they all provide the same rendering of emails. In our case, using the API, makes the mailer less powerful as we cannot control anything anymore.

bocharsky-bw:
@fabpot thank you for the clarification!

Do you recommend to always use SMTP/HTTP instead of API in projects even if email platforms recommend it? Is it like best practice for the mailer to use SMTP/HTTP?

fabpot:
Yes, I would recommend to use HTTP when available then SMTP. The reason is that using HTTP or SMTP gives you total control over the creation of the email contents and headers. With the API, you are limited to what is exposed by the provider, and not all emails will work. For instance, when we will have support for encryption and signing (4.4), this will never work with APIs, only HTTP and SMTP.

bocharsky-bw:
@fabpot Thank you a lot for more information on it!

stof:
I think the provider recommends using the API because they are easier to use than having to build the whole email yourselves. But the whole point of the Mailer component is that the component builds the email, not you, so it already handles that complexity.

bocharsky-bw:
Thank you, Christophe! Yeah, fair enough

PR #596, opened 2019-05-30
Changing mailer alias to Mailer component

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

Time to celebrate 4.3's arrival!


fabpot:
Should be reverted as the mailer component integration is (I think) not ready for prime time (unfortunately).

fabpot:
reverted

PR #597, opened 2019-05-30
Specifying mailer servers when possible and tweaking dynamic parts

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

1) Instead of just `@ses` I've tried to add the real host. I didn't know all of them, so I left those alone for now. I'd love for someone to test the missing `http` ones and update them accordingly.

2) I changed the syntax for the parts that the user is supposed to update - I thought using the `$` might make a developer think this was some sort of variable. I also added `YOUR-` at the beginning as (hopefully) a queue that they should be *their* information there.

Cheers!

### DIFF
--- a/symfony/sendgrid-mailer/4.3/manifest.json
+++ b/symfony/sendgrid-mailer/4.3/manifest.json
@@ -1,5 +1,5 @@
 {
     "env": {
-        "#1": "MAILER_DSN=smtp://$SENDGRID_KEY@sendgrid"
+        "#1": "MAILER_DSN=smtp://apikey:{YOUR-SENDGRID-API-KEY}@smtp.sendgrid.net"
     }
 }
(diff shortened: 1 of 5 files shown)


nicolas-grekas:
$VAR are real references, not just for-the-doc placeholders. Not sure this is better.

weaverryan:
> $VAR are real references,

You mean they are actually variables of some sort? Or am I misunderstanding?

symfony/sendgrid-mailer/4.3/manifest.json
```diff
@@ -1,5 +1,5 @@
 {
     "env": {
-        "#1": "MAILER_DSN=smtp://$SENDGRID_KEY@sendgrid"
+        "#1": "MAILER_DSN=smtp://apikey:{YOUR-SENDGRID-API-KEY}@smtp.sendgrid.net"
```
weaverryan: Actually, I think I'm misunderstanding this too - the `apikey` username looks like it should be provided by the transport class: https://github.com/symfony/sendgrid-mailer/blob/master/Smtp/SendgridTransport.php#L29

weaverryan:
Closing - I may propose some tweaks, but a lot of this was me misunderstanding things... as usual :p

PR #598, opened 2019-05-30
RFC: explicitly adding mailer env vars

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

It's cool how all the mailers embed other environment variables inside of them. But actually, I got confused by this - I didn't think about how I could define `SENDGRID_KEY` as its own env var, and then leave `MAILER_DSN` alone. Why not also add the `SENDGRID_KEY` as an env var? If we agree, I'll do the same for all the mailer configs.

### DIFF
--- a/symfony/amazon-mailer/4.3/manifest.json
+++ b/symfony/amazon-mailer/4.3/manifest.json
@@ -1,6 +1,8 @@
 {
     "env": {
-        "#1": "MAILER_DSN=smtp://$AWS_ACCESS_KEY:$AWS_SECRET_KEY@ses",
-        "#2": "MAILER_DSN=http://$AWS_ACCESS_KEY:$AWS_SECRET_KEY@ses"
+        "#1": "AWS_ACCESS_KEY=",
+        "#2": "AWS_SECRET_KEY=",
+        "#3": "MAILER_DSN=smtp://$AWS_ACCESS_KEY:$AWS_SECRET_KEY@ses",
+        "#4": "MAILER_DSN=http://$AWS_ACCESS_KEY:$AWS_SECRET_KEY@ses"
     }
 }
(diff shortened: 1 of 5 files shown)


fabpot:
Works for me.

symfony/amazon-mailer/4.3/manifest.json
```diff
@@ -1,6 +1,8 @@
 {
     "env": {
-        "#1": "MAILER_DSN=smtp://$AWS_ACCESS_KEY:$AWS_SECRET_KEY@ses",
-        "#2": "MAILER_DSN=http://$AWS_ACCESS_KEY:$AWS_SECRET_KEY@ses"
+        "#1": "AWS_ACCESS_KEY=",
+        "#2": "AWS_SECRET_KEY=",
+        "#3": "MAILER_DSN=smtp://$AWS_ACCESS_KEY:$AWS_SECRET_KEY@ses",
+        "#4": "MAILER_DSN=http://$AWS_ACCESS_KEY:$AWS_SECRET_KEY@ses"
```
nicolas-grekas: do we have any reco for smtp vs http?
looks like provide a choice is calling for trouble :)

bocharsky-bw: It was discussed a bit here: #594 - though I didn't see any docs that would recommend using HTTP if available, otherwise SMTP. Probably add a comment above these env vars explaining it?

fabpot: I would recommend using SMTP for now. That way, you will have the "same" experience across the board (from a local SMTP server on your dev machine to production via a provider).

fabpot:
@weaverryan Can you remove the `http` env vars and only keep the `smtp` ones?

PR #604, opened 2019-06-02
MonologBundle: use excluded_http_codes

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

This uses `excluded_http_codes` instead of `excluded_404s`, which is more flexible (and `excluded_404s` is deprecated in the next, unreleased version of MonologBundle).

### DIFF
--- a/symfony/monolog-bundle/3.3/config/packages/prod/monolog.yaml
+++ b/symfony/monolog-bundle/3.3/config/packages/prod/monolog.yaml
@@ -1,0 +1,25 @@
+monolog:
+    handlers:
+        main:
+            type: fingers_crossed
+            action_level: error
+            handler: nested
+            excluded_404s:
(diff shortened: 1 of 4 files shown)


symfony/monolog-bundle/3.3/config/packages/prod/monolog.yaml
```diff
@@ -0,0 +1,23 @@
+monolog:
+    handlers:
+        main:
+            type: fingers_crossed
+            action_level: error
+            handler: nested
+            excluded_http_codes: [404, 405]
```
weaverryan: This line is the only change:

```diff
-            excluded_404s:
-                # regex: exclude all 404 errors from the logs
-                - ^/
+            excluded_http_codes: [404, 405]
```

PR #605, opened 2019-06-04
Remove the symfony/symfony recipe

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

The message is not displayed anymore.

image


Pierstoval:
How is it supposed to work?

For now, compatibility with flex & `symfony/symfony` is granted 😕 

image

PR #610, opened 2019-06-13
Added forking step to instructions

vatri:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | https://github.com/symfony/recipes-contrib/issues/678

Added forking step to instructions as discussed in the above issue.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,7 +10,7 @@ the Symfony Core Team. For contributed recipes, see the `contrib repository`_.
 Creating Recipes
 ----------------
 
-1. Clone a repository
+1. Fork the repository
 ~~~~~~~~~~~~~~~~~~
 In order to create a recipe, you will need to fork symfony/recipes-contrib repository. Then follow below instructions.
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -10,6 +10,13 @@ the Symfony Core Team. For contributed recipes, see the `contrib repository`_.
 Creating Recipes
 ----------------
 
+1. Clone a repository
```
bocharsky-bw: Hm, technically it's not required to _clone_ the repo, you can use GitHub GUI for this

vatri: @bocharsky-bw This is error , I wanted to write "fork" instead "clone". I have fixed it now.

bocharsky-bw:
Fairly speaking, I agree with @maxhelias - it seems too obvious for me. IIRC everywhere in Symfony components, we just say "send Pull Requests in the main Symfony repository" without explaining how to do this, because for us it's done upstream by GitHub. The same here I think

vatri:
@maxhelias You might be right.

I was reading some tutorial so I was confused of how and where recipes are being created.

fabpot:
Let's close for the reasons explained by some commenters

PR #611, opened 2019-06-14
ensure $_ENV is contains real env vars from $_SERVER , or defaults from .env.local.php

bendavies:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | https://github.com/symfony/flex/issues/516b

Fixes the issue where by `.env.local.php` was not overridable by real env vars.


tristanbes:
@nicolas-grekas // @javiereguiluz 
This is a very important patch. I think the community should be notified of this patch so we can take action to apply it. 

It took down our production and was hard to debug since in the documentation it's clearly saying that real env var takes over the var contained on those files.

mareksuscak:
The same thing happened to us @tristanbes and I totally agree this needs to get more attention — possibly a blog post?

PR #616, opened 2019-06-25
Add built-in support for symfony/requirements-checker

ro0NL:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | symfony/symfony-docs#...    (https://symfony.com/doc/current/reference/requirements.html)

This is an alternative approach for https://github.com/symfony/requirements-checker/pull/35 at the recipe level.

:+1:  It works for everybody
:+1:  It enforces requirements checking
:+1:  It's boilerplate, ditch it if not needed
:man_shrugging: It's boilerplate
:man_shrugging:  we could remove the template upstream to reduce maintenance, plain text is nice actually IMHO
:+1:  CLI/WEB support out-of-the-box, no upstream changes needed other than polishing maybe
:+1: Out-of-the-box workflow by leveraging `composer require/remove req-check`

image

image

### DIFF
--- a/symfony/framework-bundle/4.2/config/bootstrap.php
+++ b/symfony/framework-bundle/4.2/config/bootstrap.php
@@ -7,7 +7,7 @@ require dirname(__DIR__).'/vendor/autoload.php';
 
 // Verify Symfony requirements
 if (class_exists(SymfonyRequirements::class)) {
-    $argv = [];
+    $argv = ['-v'];
     echo (isset($_SERVER['HTTP_HOST']) ? '<pre>' : '')."REMOVE THIS MESSAGE: composer remove symfony/requirements-checker\n\n";
     require dirname(__DIR__).'/vendor/symfony/requirements-checker/bin/requirements-checker.php';
     die(1);
--- a/symfony/framework-bundle/4.2/config/bootstrap.php
+++ b/symfony/framework-bundle/4.2/config/bootstrap.php
@@ -1,16 +1,14 @@
 <?php
 
 use Symfony\Component\Dotenv\Dotenv;
-use Symfony\Component\HttpKernel\Kernel;
 use Symfony\Requirements\SymfonyRequirements;
 
 require dirname(__DIR__).'/vendor/autoload.php';
 
 // Verify Symfony requirements
 if (class_exists(SymfonyRequirements::class)) {
-    $requirements = new SymfonyRequirements(dirname(__DIR__), Kernel::VERSION);
     $argv = [];
-    echo "<pre>REMOVE THIS MESSAGE: composer remove symfony/requirements-checker\n\n";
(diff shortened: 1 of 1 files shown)


weaverryan:
Is the requirements checker a dependency that users get by default with the skeletons? I think not... just checking. If not, what’s the workflow? The user would need to still know to install this, right?

It’s an interesting idea - I just don’t know the full vision for the experience :). And while I like this check (and it can prevent wtf’s), I don’t want to “interrupt” the experience of starting a new project.

bocharsky-bw:
The `config/bootstrap.php` file is included in a few more spots, not only in `bin/console` but also in `public/index.php`, etc. I don't think it's a good idea

ro0NL:
It's a bit of a trade-off path i realize.

> what’s the workflow? The user would need to still know to install this, right?

Yes. It should be documented like `composer require req-check` (assuming flex), check bin/console + URL, fix requirements, `composer remove ...`

So the extra bonus is we enforce a temporary dependency and enforce to remove it after (as shown in screenshot).

> not only in bin/console but also in public/index.php

Those are the targeted front controllers yes, so we can check both CLI and WEB.

It's a bit of a trade-off path i realize.

ro0NL:
effectively the workflow is the same as currently; you need the package. But it's a tool, we don't want it to be part of the project deps in e.g. VCS.

Then you need to know about extra infrastructure like `vendor/bin/requirements-checker` and `public/check.php` or rely on SF Flex. 

`public/check.php`might need e.g. custom NGINX config as well.

ro0NL:
we can also leverage .env to opt-in/out

as mentioned in https://github.com/symfony/requirements-checker/pull/35, i tend to believe it should be possible to run an environment check on prod (the current localhost check blocks it), before e.g. live DNS is enabled.

weaverryan:
I like the workflow quite a lot. I wish we could do it without adding extra, permanent code to index.php... but, indeed, I’m not sure if that’s possible. For the CLI case, I don’t think knowing about vendor/bin/requirements-checker is a big deal, as that can be documented with a post-install on that package. We could even add a check.php in that recipe. That does have the disadvantage of possibly needing special Nginx config. I’m just trying to brainstorm possible ways to do this without permanent changes to bootstrap.php... though I’m not necessarily opposed to this (it’s just not ideal).

fabpot:
We don't want to add such a check in the hot path.

ro0NL:
I see.. what about moving it after the .env loading, and trigger based on `SOMETHING=1`. If within this scope the class does not exists we throw.

fabpot:
The main point I think is that I don't want to leak code from the requirement checker in the main code. It should be entirely decoupled like it is now.

ro0NL:
Ok, fair. In this case i will create a custom workflow for us :)

ro0NL:
So.. i figured we can rely on the same php.ini in cli+web.

For now a simple make task does the job pretty well :)

```bash
requirement-check:
	${composer} require symfony/requirements-checker ${composer_args} --no-scripts -q
	${app} vendor/bin/requirements-checker
	${composer} remove symfony/requirements-checker -q
```

PR #619, opened 2019-06-28
Suggest installing symfony/debug

yceruto:
| Q             | A
| ------------- | ---
| License       | MIT

See https://github.com/symfony/symfony/pull/32270

### DIFF
--- a/symfony/framework-bundle/4.2/public/index.php
+++ b/symfony/framework-bundle/4.2/public/index.php
@@ -9,7 +9,11 @@ require dirname(__DIR__).'/config/bootstrap.php';
 if ($_SERVER['APP_DEBUG']) {
     umask(0000);
 
-    Debug::enable();
+    if (class_exists(Debug::class)) {
+        Debug::enable();
+    } else {
+        throw new RuntimeException('You need to add "symfony/debug" as a development Composer dependency.');
+    }
 }
 
 if ($trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? $_ENV['TRUSTED_PROXIES'] ?? false) {
--- a/symfony/console/3.3/bin/console
+++ b/symfony/console/3.3/bin/console
@@ -34,6 +34,8 @@ if ($_SERVER['APP_DEBUG']) {
 
     if (class_exists(Debug::class)) {
         Debug::enable();
+    } else {
+        throw new RuntimeException('You need to add "symfony/debug" as a development Composer dependency.');
     }
 }
 


OskarStark:
But this isn’t a suggestion, it’s a hard dependency, isn’t it? If it exists it’s enabled, if not you cannot go further. Or am I wrong?

symfony/framework-bundle/4.2/public/index.php
```diff
@@ -9,7 +9,11 @@
 if ($_SERVER['APP_DEBUG']) {
     umask(0000);
 
-    Debug::enable();
+    if (class_exists(Debug::class)) {
+        Debug::enable();
```
weaverryan: I understand wrapping this in an if and only running when it exists, but I don’t understand throwing an exception if it doesn’t exist.

nicolas-grekas:
On another recipe/PR, I suggested moving the Debug class to the new component. Applies here too.

yceruto:
Yes, that would solve the problem. I'll make the move, but that mean the Debug component is almost empty, does make make sense move all and deprecating it completely?

fancyweb:
> On another recipe/PR, I suggested moving the Debug class to the new component. Applies here too.

The `Debug` class calls the `DebugClassLoader`. So the `ErrorCatcher` would need to require `symfony/debug`... IMHO, the `Debug` class needs to stay in the `Debug` component or this component can be renamed to the `Debug class loader` component 😕 

Or, since the `DebugClassLoader` usage in the `Debug` class is only `DebugClassLoader::enable()`, we could : 
- Deprecate the `Debug` component `Debug` class in 4.4 to remove it in 5.0.
- Create a new class `ErrorCatcher` with the `enable` method in the `ErrorCatcher` component (copy pasted from the current `Debug` class).
- We update the recipes to call `DebugClassLoader::enable(); ErrorCatcher::enable();` in the if.
- To solve the existing project update, we can include a file in the `FrameworkBundle` that would substitute the `Debug` class if it does not exist, do the calls but throw a deprecation to ensure users switch to the new behaviors for 5.0.

The problem with splitting is that if we have new debug behaviors, it would be better to have a single method like `Debug::enable();` to be able to enable it by default and without user effort.
So we can also just add the `symfony/debug` dependency to the recipe so new projects are ready and for the existing ones, do my number 4 point of the above list. So at least `symfony/debug` is not a hard dependency of the `FrameworkBundle` as it feels wrong.

yceruto:
Closing in favor of https://github.com/symfony/symfony/pull/32303

PR #621, opened 2019-07-02
[easy-log-handler] Fix service configuration (As it throws ServiceNotFoundException)

youassi:
| Q             | A
| ------------- | ---
| License       | MIT


maxhelias:
If I remember it's an old service name, what is your use case ?
I think it's a bad idea to change the name of the service. Maybe the solution would be to set an alias

youassi:
@maxhelias  Ok thanks for the reply, I just install the bundle, I did copy an old config with that service name, and I got a ServiceNotFoundException.

PR #626, opened 2019-07-10
[doctrine-bundle] Add DATABASE_DRIVER in .env

mazarini:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | doctrine/doctrine-bundle#626    

Allow to share one config and use differents databases with .env.local.
(add DATABASE_SERVER_VERSION too)

### DIFF
--- a/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml
@@ -1,8 +1,7 @@
 doctrine:
     dbal:
-        # configure these for your database server
-        driver: 'pdo_mysql'
-        server_version: '5.7'
+        driver: '%env(resolve:DATABASE_DRIVER)%'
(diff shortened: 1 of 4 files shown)


mazarini:
Test OK for "composer require orm" instead of "composer require symfony/doctrine-bundle".

BoShurik:
See https://github.com/symfony/recipes/pull/462

mazarini:
An other option is to add a fallback in doctrine.yaml for DATABASE_SERVER_VERSION and DATABASE_DRIVER. So we can put it as comment in .env and set it only if needed.
Nothing change if it's not use.

weaverryan:
This issue is tough. There are two problems with your approach:

1) `DATABASE_SERVER_VERSION` should not be something that is allowed to change (as an env var). Doing this might mean that one developer (with one version) generates migrations using newer SQL features than our supported on production's true server version.

2) The `default_table_options` are still in the YAML file, and are specific to MySQL.

Cheers!

PR #627, opened 2019-07-12
[PhpUnitBridge] git ignore result cache

ro0NL:
| Q             | A
| ------------- | ---
| License       | MIT

Same as #553

### DIFF
--- a/symfony/phpunit-bridge/4.1/manifest.json
+++ b/symfony/phpunit-bridge/4.1/manifest.json
@@ -8,6 +8,7 @@
     },
     "gitignore": [
         ".phpunit",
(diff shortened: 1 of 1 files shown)


ro0NL:
@fabpot another approach is XML config, maybe by default even:

```xml
<phpunit
    cacheResult="true"
    cacheResultFile="var/phpunit.cache"
    executionOrder="defects"
    stopOnDefect="true"
```

ro0NL:
also for the bridge i think we should write to `bin/.phpunit/<version.cache` somehow :thinking: 

@nicolas-grekas thoughts?

PR #630, opened 2019-07-20
Use the new ErrorHandler\Debug class if possible

yceruto:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -  

As part of the Debug component deprecation.

### DIFF
--- a/symfony/console/3.3/bin/console
+++ b/symfony/console/3.3/bin/console
@@ -4,7 +4,8 @@
 use App\Kernel;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Component\Console\Input\ArgvInput;
-use Symfony\Component\Debug\Debug;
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
I have a different proposal: to not break existing project for no real reason, what about having a Debug class in the Debug namespace in the ErrorHandler component? That would help create a smooth transition.

Tobion:
If the debug class is only for debugging as the name suggests, why not keep it in the Debug component together with other classes for debugging like DebugClassLoader?

I still don't get the role of the Debug class. It has this param doc
>@param bool $displayErrors Whether to display errors (for development) or just log them (for production)

So it's also meant for production but enables the DebugClassLoader? I don't think you ever want to do that.

yceruto:
Closing in favor of https://github.com/symfony/symfony/pull/32665

PR #634, opened 2019-07-31
New 4.4 recipe to use ErrorHandler\Debug class by default

yceruto:
| Q             | A
| ------------- | ---
| License       | MIT

https://github.com/symfony/symfony/pull/32665#issuecomment-516921646

New projects based on 4.4 must to use `Symfony\Component\ErrorHandler\Debug` class in their front controller


symfony/console/4.4/bin/console
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env php
+<?php
+
+use App\Kernel;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\ErrorHandler\Debug;
```
yceruto: only this line was changed here

symfony/framework-bundle/4.4/public/index.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+use App\Kernel;
+use Symfony\Component\ErrorHandler\Debug;
```
yceruto: only this line was changed here

Tobion:
Other reciepes use symbolic links between versions for files that didn't change.

PR #636, opened 2019-08-04
fix phpunit documentation link

garak:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | none

PHPUnit moved its documentation to readthedocs.io since a while

### DIFF
--- a/phpunit/phpunit/4.7/phpunit.xml.dist
+++ b/phpunit/phpunit/4.7/phpunit.xml.dist
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<!-- https://phpunit.de/manual/current/en/appendixes.configuration.html -->
+<!-- https://phpunit.readthedocs.io/en/latest/configuration.html -->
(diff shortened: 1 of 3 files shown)


phpunit/phpunit/4.7/phpunit.xml.dist
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<!-- https://phpunit.de/manual/current/en/appendixes.configuration.html -->
+<!-- https://phpunit.readthedocs.io/en/latest/configuration.html -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.5/phpunit.xsd"
```
Tobion: this should also be changed to https as http just redirects to https

garak: I'm not sure about it. It's reported as is on every configuration in phpunit repository

bocharsky-bw: Hm, probably should be fixed upstream as well?

PR #637, opened 2019-08-08
bumping to php 7.5 to stay on supported version

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | none should be needed   

6.5 if EOL since February. Is the philosophy that we stay on the oldest-supported... or stay on the newest (and I should bump to 8.2?)


OskarStark:
😆 ah you mean phpUNIT 7.5 I thought you were living on the edge and use the super new shiny (not available yet) PHP 7.5 🤣

PR #644, opened 2019-09-03
Fix Mailer DSNs for 4.4

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/google-mailer/4.4/manifest.json
+++ b/symfony/google-mailer/4.4/manifest.json
@@ -1,8 +1,6 @@
 {
     "env": {
         "#1": "Gmail SHOULD NOT be used on production, use it in development only.",
-        "#2": "GMAIL_USERNAME=",
-        "#3": "GMAIL_PASSWORD=",
-        "#4": "MAILER_DSN=gmail://$GMAIL_USERNAME:$GMAIL_PASSWORD@default"
+        "#2": "MAILER_DSN=gmail://USERNAME:PASSWORD@default"
     }
 }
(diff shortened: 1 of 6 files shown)


fabpot:
This pull request updates the DSNs and also removes the intermediary env vars for following reasons:

* Having env vars embedded into other ones does not work well with Dotenv;
* Having less env vars is "better"
* Having the user/pass/key/id embedded into the DSN makes it easier to understand that they need to be URL encoded.

symfony/google-mailer/4.4/manifest.json
```diff
@@ -0,0 +1,6 @@
+{
+    "env": {
+        "#1": "Gmail SHOULD NOT be used on production, use it in development only.",
```
Koc: Should comment to be prefixed with `#`? Also why Gmail is not recommended for prod?

stof: it is prefixed (in the name of the env var, which is what goes at the beginning of the line)

fabpot: Gmail is not meant to be used to send emails from an application. Accounts are personal. So, it works for testing on a dev machine, but not for prod.

symfony/sendgrid-mailer/4.4/manifest.json
```diff
@@ -0,0 +1,5 @@
+{
+    "env": {
+        "#1": "MAILER_DSN=sendgrid://KEY@default"
```
bocharsky-bw: Shouldn't it be `MAILER_DSN=sendgrid+smtp://KEY@default` to be consistent with 4.3 recipe?

fabpot: `sendgrid` is the equivalent of `sendgrid+smtp`. The default is now automatically the "best" transport possible.

bocharsky-bw: I got it, thank you for the clarification!

PR #645, opened 2019-09-05
Add errors preview routes to the framework config

yceruto:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | TODO

According to https://github.com/symfony/symfony/pull/33271 and https://github.com/symfony/symfony/pull/33476


symfony/twig-bundle/5.0/config/packages/twig.yaml
```diff
@@ -0,0 +1,4 @@
+twig:
+    default_path: '%kernel.project_dir%/templates'
+    debug: '%kernel.debug%'
+    strict_variables: '%kernel.debug%'
```
yceruto: removed `exception_controller` option

symfony/framework-bundle/4.4/config/routes/dev/framework.yaml
```diff
@@ -0,0 +1,3 @@
+_errors:
+    resource: '@FrameworkBundle/Resources/config/routing/errors.xml'
+    prefix: /__error
```
yceruto: added double underscore to avoid route conflicts with TwigBundle.

yceruto:
Also removed `config/routes/dev/twig.yaml` from 5.0 recipe

yceruto:
I don't get the failure spotted by Symfony Recipes Server:
> There are some validation errors
> Package symfony/framework-bundle (4.4)
symfony/framework-bundle/4.4/config
File must be under a directory defined in the "copy-from-recipe" section of "manifest.json"

But `symfony/framework-bundle/4.4/config` is already a directory defined in "manifest.json" :/ any clue?

fabpot:
@yceruto You must copy the `manifest.json` file, having a symlink does not work here.

yceruto:
@fabpot still failing :/

yceruto:
It's weird, because the `manifest.json` file in TwigBundle 5.0 recipe is a symlink, and it's ok

yceruto:
well, just to check, I've copy all files without symlink into a new 4.4 recipe, and it fails anyway, now 2 errors `config/` and `src/` dirs... :/

yceruto:
See also https://github.com/symfony/recipes/pull/634 which passed without errors, and there `manifest.json` is a symlink.

symfony/twig-bundle/5.0/config/packages/twig.yaml
```diff
@@ -0,0 +1,4 @@
+twig:
+    default_path: '%kernel.project_dir%/templates'
```
stof: I think this could be removed too, as it is already the default path in 5.0 (this option being here to allow opting into the new structure in a 4.x project)

yceruto:
@fabpot I think it's a server validation bug, I did some tests and concluded the following: when I delete a symlink directory and add the real one instead, the validation fails. It is only related to the changeset, because if I add a new directory or a new symlink directory, it will pass.

symfony/twig-bundle/5.0/config/packages/twig.yaml
yceruto: That will apply to `debug` and `strict_variables` options.

yceruto: I also think they are not here due to of the new structure 4.0 only, but because they are also relevant options for the user who might want to change.

fabpot: We tend to not have default values when people should not change them often. I think that this is the case here. So, I would be in favor of removing it.

fabpot: So, it means that we can remove the file in 5.0?

fabpot:
@yceruto I won't have time to investigate the issue in the coming days, so, whenever this PR is ready, I will force the merge.

symfony/twig-bundle/5.0/config/packages/twig.yaml
yceruto: Technically yes, we can.

stof: AFAIU, `default_path` was a migration-only setting that we will deprecate again in 5.1. That's not something users should change in 5.0 (they should however change it in 4.x to opt-in for the new structure).

yceruto:
the `packages` directory was excluded from the TwigBundle 5.0 recipe, this is ready on my side.

symfony/twig-bundle/5.0/config/packages/twig.yaml
yceruto: @stof I don't think so, the `default_path` value was since the begining the same `%kernel.project_dir/templates%`, and it was intended to be configurable.

PR #646, opened 2019-09-09
Remove PANTHER env variable

rvanlaak:
Q | A
-- | --
License | MIT

... as https://github.com/symfony/test-pack/pull/9 removed Panther from `symfony/test-pack` last June.

cc @nicolas-grekas


dunglas:
👎 doing this will break all Panther installs.

dunglas:
Panther can still be installed using `composer req panther`. It has just been removed of the bundled dependencies for `website-skeleton` because PHP WebDriver (used by Panther) requires `ext-curl`. Currently Flex doesn't provide a way to update `.env.test` in a dedicated recipe. So there there is no other way for now.

Maybe than introducing a new key such as `env-test` in the manifest could help, but it will require more work (patching Flex).

rvanlaak:
> 👎 doing this will break all Panther installs.

I was under the assumption that the recipe solely let composer copy the file with initial values, where thereafter userland is responsible again for it's value.

Implementing Panther is in one of our follow-up tickets, so it was weird to find the env value while there was no vendor code yet.

dunglas:
Actually the best fix would be to allow recipes to add env var per environment `.env` files.

PR #649, opened 2019-09-23
[security-bundle] enable anonymous: lazy by default

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

Needs https://github.com/symfony/symfony/pull/33676

### DIFF
--- a/symfony/security-bundle/4.4/config/packages/security.yaml
+++ b/symfony/security-bundle/4.4/config/packages/security.yaml
@@ -1,0 +1,22 @@
+security:
+    # https://symfony.com/doc/current/security.html#where-do-users-come-from-user-providers
+    providers:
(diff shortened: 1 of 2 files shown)


symfony/security-bundle/4.4/config/packages/security.yaml
```diff
@@ -0,0 +1,22 @@
+security:
+    # https://symfony.com/doc/current/security.html#where-do-users-come-from-user-providers
+    providers:
+        in_memory: { memory: null }
+    firewalls:
+        dev:
+            pattern: ^/(_(profiler|wdt)|css|images|js)/
+            security: false
+        main:
+            anonymous: lazy
```
nicolas-grekas: that's the only changed line from 3.3

PR #651, opened 2019-09-27
Trust proxies on private and local IPs by default

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

From https://en.wikipedia.org/wiki/Reserved_IP_addresses

I did not list `192.168.0.0/16` because it's used by home routers and these don't fit in the defaults to me. Makes sense?

Fixes https://github.com/symfony/symfony/issues/33578

### DIFF
--- a/symfony/framework-bundle/4.2/manifest.json
+++ b/symfony/framework-bundle/4.2/manifest.json
@@ -14,7 +14,7 @@
     "env": {
         "APP_ENV": "dev",
         "APP_SECRET": "%generate(secret)%",
-        "#TRUSTED_PROXIES": "127.0.0.1,127.0.0.2",
+        "TRUSTED_PROXIES": "10.0.0.0/8,172.16.0.0/12,127.0.0.0/16",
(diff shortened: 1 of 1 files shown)


B-Galati:
IMHO 192.168.0.0/16 should be part of the default whitelist. It's an internal network like the others. And I am sure some home router are using 10.0.0.0  or 172.16.0.0 or whatever.

nicolas-grekas:
I've never seen anyone putting a reverse proxy on `192.168.*`.If nobody does, we don't have to.

B-Galati:
> I've never seen anyone put a reverse proxy on `192.168.*`. If nobody does, we don't have to support it by default.

How can you be so sure nobody does it ?

nicolas-grekas:
> How can you be so sure nobody does it ?

That's reverse thinking: I'm not sure of anything. But I prefer closing and opening later on request. Then we'll know. And if nobody reports anything, you'll have your answer.

nicolas-grekas:
I'm now pretty sure this must be opt-in.

PR #652, opened 2019-09-30
Make use of the local phpunit.xsd

julienfalque:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Using the local `phpunit.xsd` file guarantees we use the exact version that is installed. It also improves DX with PhpStorm as it becomes able to validate `phpunit.xml` without manually registering XSD file's URI.

### DIFF
--- a/symfony/phpunit-bridge/3.3/phpunit.xml.dist
+++ b/symfony/phpunit-bridge/3.3/phpunit.xml.dist
@@ -2,7 +2,7 @@
 
 <!-- https://phpunit.readthedocs.io/en/latest/configuration.html -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.5/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="bin/.phpunit/phpunit.xsd"
          backupGlobals="false"
          colors="true"
          bootstrap="config/bootstrap.php"
--- a/symfony/phpunit-bridge/4.1/phpunit.xml.dist
+++ b/symfony/phpunit-bridge/4.1/phpunit.xml.dist
@@ -2,7 +2,7 @@
 
 <!-- https://phpunit.readthedocs.io/en/latest/configuration.html -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.5/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="bin/.phpunit/phpunit.xsd"
          backupGlobals="false"
          colors="true"
          bootstrap="config/bootstrap.php"
--- a/phpunit/phpunit/4.7/phpunit.xml.dist
+++ b/phpunit/phpunit/4.7/phpunit.xml.dist
@@ -2,7 +2,7 @@
 
 <!-- https://phpunit.readthedocs.io/en/latest/configuration.html -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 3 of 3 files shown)


symfony/phpunit-bridge/3.3/phpunit.xml.dist
```diff
@@ -2,7 +2,7 @@
 
 <!-- https://phpunit.readthedocs.io/en/latest/configuration.html -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.5/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
```
Pierstoval:
```suggestion
         xsi:noNamespaceSchemaLocation="bin/.phpunit/phpunit-6.5/phpunit.xsd"
```

symfony/phpunit-bridge/4.1/phpunit.xml.dist
```diff
@@ -2,7 +2,7 @@
 
 <!-- https://phpunit.readthedocs.io/en/latest/configuration.html -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.5/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
```
Pierstoval:
```suggestion
         xsi:noNamespaceSchemaLocation="bin/.phpunit/phpunit-6.5/phpunit.xsd"
```

symfony/phpunit-bridge/3.3/phpunit.xml.dist
nicolas-grekas: Putting the version in the path means the reference will be broken when upgrading.
We could make simple-phpunit copy the xsd in the parent directory when it runs.
PR welcome on symfony/symfony

Pierstoval: This would be quite a dirty hack IMO, I'm actually not in favor of my proposal at all and I prefer to use the final URL anyway.

Many devs are using PHPStorm and it's really straightforward to fetch the XSD file:
image

nicolas-grekas: Can you elaborate why you write "dirty hack" here? That would need some justifying.

Pierstoval: Also, erasing copy/pasting the XSD file would mean we'll have to save it somewhere, so if it's removed but phpunit isn't, the XSD wouldn't be taken in account anymore by PHPStorm.

Fetching from the URL is the easiest solution IMO

julienfalque: Using the HTTP URI requires to check whether the referenced version is correct and is a manual step that this PR means to avoid.

I'll try to work on a PR on symfony/symfony.

Pierstoval: In such case, the bridge possibly could fetch PHPUnit's version and only use the `xml` extension to update the version link right inside the project's xml file and check it at runtime before running the tests, so updating phpunit would automatically update `phpunit.xml.dist`

julienfalque: See symfony/symfony#33776.

julienfalque: I updated the path to `bin/.phpunit/phpunit.xsd`.

Tobion:
This is so good. Thanks for the idea.

PR #654, opened 2019-10-08
Trust proxies on private and local IPs (commented by default)

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Closes https://github.com/symfony/symfony/issues/33578


nicolas-grekas:
@stof OK for you?

stof:
should we add the ipv6 private range too ?

nicolas-grekas:
> should we add the ipv6 private range too ?

I know nothing about these, never used. In a later PR if required?

PR #661, opened 2019-10-24
Add a recipe for Notifier

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/notifier/5.0/config/packages/notifier.yaml
+++ b/symfony/notifier/5.0/config/packages/notifier.yaml
@@ -1,0 +1,16 @@
+framework:
+    notifier:
+        #chatter_transports:
+        #    slack: '%env(SLACK_DSN)%'
+        #    telegram: '%env(TELEGRAM_DSN)%'
+        #texter_transports:
+        #    twilio: '%env(TWILIO_DSN)%'
+        #    nexmo: '%env(NEXMO_DSN)%'
+        channel_policy:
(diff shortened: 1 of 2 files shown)


symfony/notifier/5.0/config/packages/notifier.yaml
```diff
@@ -0,0 +1,16 @@
+framework:
+    notifier:
+        #chatter_transports:
+        #    slack: '%env(SLACK_DSN)%'
+        #    telegram: '%env(TELEGRAM_DSN)%'
+        #texter_transports:
+        #    twilio: '%env(TWILIO_DSN)%'
+        #    nexmo: '%env(NEXMO_DSN)%'
+        channel_policy:
+            # use chat/slack, chat/telegram, sms/twilio or sms/nexmo
```
weaverryan: I don’t know the component yet, but it’s not clear where these available options come from and why email is also an option. The chatter and texter above don’t match the chat/sms here. This might not be the right place for the feedback - just passing my novice questions :)

fabpot: We have severals channels: sms, chatter, email. Each channel might have specific transports, so you have chat/slack and chat/telegram.

yceruto: That'd mean we can use: `urgent: ['email', 'sms/twilio']` where `sms` is the channel and `twilio` the transport, right?

fabpot:
@yceruto That's correct

PR #663, opened 2019-10-27
Add strict mode to webpack encore bundle

PabloKowalczyk:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | no

This PR adds new options to WebpackEncoreBundle and disables `strict-mode` in test environment.
EDIT: Ping @weaverryan


symfony/webpack-encore-bundle/1.6/config/packages/webpack_encore.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+    # Throw an exception if the entrypoints.json file is missing or an entry is missing from the data
+    # strict_mode: false
+
+    # if you have multiple builds:
+    # builds:
+    # pass "frontend" as the 3rg arg to the Twig functions
+    # {{ encore_entry_script_tags('entry1', null, 'frontend') }}
+
+    # frontend: '%kernel.project_dir%/public/frontend/build'
```
weaverryan: This should be indented under `builds`, right? The formatting in this section with the comments confuses me (might just be some missing indentation)

symfony/webpack-encore-bundle/1.6/config/packages/webpack_encore.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+    # Cache the entrypoints.json (rebuild Symfony's cache when entrypoints.json changes)
+    # Available in version 1.2
```
weaverryan: Don't need this comment anymore, since the recipe is for 1.6

symfony/webpack-encore-bundle/1.6/config/packages/test/webpack_encore.yaml
```diff
@@ -0,0 +1,2 @@
+webpack_encore:
+    strict_mode: false
```
weaverryan: I'm not convinced by this. I mean, I'm really mixed - I'm not against it. I've never personally hit this issue - we always build our Encore assets in test (because we do a lot of browser testing), and I would appreciate a clear exception if we forgot vs just the page being very broken. The fact that the `strict_mode` option exists might be enough? It really depends on what "most" users will do.

PabloKowalczyk: Now `strict_mode` is commented for tests, it will allow users to fast change it if they need it. WDYT?

symfony/webpack-encore-bundle/1.6/config/packages/test/webpack_encore.yaml
```diff
@@ -0,0 +1,2 @@
+#webpack_encore:
+#    strict_mode: false
```
nicolas-grekas: should be removed instead

PabloKowalczyk: Idea behind leaving this file is to make people aware that there is `strict_mode` option and they don't need to download 1TB of npm modules and compile Webpack just to run functional tests. WDYT?

PR #664, opened 2019-10-31
[Validator] Enable auto mapping by default

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | todo

Now that https://github.com/symfony/symfony/pull/32107 has been merged, auto mapping can be on again.

### DIFF
--- a/symfony/validator/4.4/config/packages/test/validator.yaml
+++ b/symfony/validator/4.4/config/packages/test/validator.yaml
@@ -1,0 +1,1 @@
+../../../../4.3/config/packages/test/validator.yaml
--- a/symfony/validator/4.4/config/packages/test/validator.yaml
+++ b/symfony/validator/4.4/config/packages/test/validator.yaml
@@ -1,1 +1,3 @@
-../../../../4.3/config/packages/test/validator.yaml
+framework:
+    validation:
+        not_compromised_password: false
--- a/symfony/validator/4.4/config/packages/validator.yaml
+++ b/symfony/validator/4.4/config/packages/validator.yaml
@@ -1,0 +1,8 @@
+framework:
+    validation:
+        email_validation_mode: html5
+
+        # Enables validator auto-mapping support.
+        # For instance, basic validation constraints will be inferred from Doctrine's metadata.
+        auto_mapping:
+            App\Entity\: []
--- a/symfony/validator/4.4/manifest.json
+++ b/symfony/validator/4.4/manifest.json
@@ -1,1 +1,6 @@
-../4.3/manifest.json
+{
+    "copy-from-recipe": {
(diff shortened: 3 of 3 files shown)


weaverryan:
I think we should add a docs PR to go along with this - it’s a really important change.

nicolas-grekas (dismissed):
(@dunglas please create a doc issue at least)

dunglas:
Actually, it's already documented https://github.com/symfony/symfony-docs/pull/11132

weaverryan:
I like this feature! But it does have side effects. For example, in `make:user`, we will need to make sure that we add the "opt-out" annotation above the `password` property... which we can only actually do if the validator component is installed at the moment the command is run :/.

If a user doesn't have the "opt-out" annotation above `password`, when they create a registration form, it will always fail with the generic "this value cannot be blank" validation error - which will appear at the *top* of the form, not attached to any field. That's because `password` won't be part of the form - the user will have some sort of a `plainPassword` which is *actually* what's on the form.

That's the big WTF that we need to navigate around if we're going to enable this feature by default.

dunglas:
Does it make sense to create a user system without validation? I mean you already need to check that the username is not taken, that the password isn’t empty, etc... isn’t it?

Cannot the command throw if the Validator component isn’t installed?

weaverryan:
> Cannot the command throw if the Validator component isn’t installed?

I was thinking about this as well. Probably we need to allow a `User` to be created without requiring validator (there are likely at least some cases where users are created via some system that doesn't need validation). But we could (at least) *definitely* add the opt-out annotation when they run `make:registration-form`. Basically, I think we can do a pretty good job in Maker with messaging and being smart.

Btw, there is another docs issue about this being automatically enabled, that still needs work.

weaverryan:
Btw, I'd like to merge this as soon as we can so that it "feels" like 4.4 and 5.0 consistently have this feature enabled. But, I think we need the MakerBundle tweak and docs first. I'm working as quickly as I can on those :).

weaverryan:
MakerBundle PR for `@DisableAutoMapping`: https://github.com/symfony/maker-bundle/pull/514

But the feature currently doesn't work :/ - see symfony/symfony#34672

And (sorry!) I'm fighting back a little bit again on this feature - symfony/symfony#34860.

nicolas-grekas:

dunglas:
@nicolas-grekas done. The failure looks unrelated, or am I missing something?

javiereguiluz:
I'm closing this proposal as "won't merge" because after discussing about this in the Core Team, it was decided that it's better to not enable `auto_mapping` by default in Symfony apps. Thanks.

PR #669, opened 2019-11-07
Replaced silent fail for loading php env file

fightbulc:
Replaced with actual test if file exists since the suppressing of an error just causes issues.

For instance if you run Xdebug.halt_level = E_WARNING

Xdebug notices that an error has happened but since we silenty fail for the include we never get an error output. With the change in place we do see an error if we want to see it.

| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/framework-bundle/4.2/config/bootstrap.php
+++ b/symfony/framework-bundle/4.2/config/bootstrap.php
@@ -6,7 +6,8 @@ require dirname(__DIR__).'/vendor/autoload.php';
 
 // Load cached env vars if the .env.local.php file exists
 // Run "composer dump-env prod" to create it (requires symfony/flex >=1.2)
-if (is_array($env = @include dirname(__DIR__).'/.env.local.php')) {
+$phpEnvFile = dirname(__DIR__).'/.env.local.php');
+if (file_exists($phpEnvFile) && is_array($env = include($phpEnvFile)) {
     foreach ($env as $k => $v) {
         $_ENV[$k] = $_ENV[$k] ?? (isset($_SERVER[$k]) && 0 !== strpos($k, 'HTTP_') ? $_SERVER[$k] : $v);
     }


symfony/framework-bundle/4.2/config/bootstrap.php
```diff
@@ -6,7 +6,8 @@
 
 // Load cached env vars if the .env.local.php file exists
 // Run "composer dump-env prod" to create it (requires symfony/flex >=1.2)
-if (is_array($env = @include dirname(__DIR__).'/.env.local.php')) {
+$phpEnvFile = dirname(__DIR__).'/.env.local.php');
+if (file_exists($phpEnvFile) && is_array($env = include($phpEnvFile)) {
```
bocharsky-bw: Hm, what if the file exists but could not be read due to strict permissions? Is it possible? Probably we still need `@include` here

nicolas-grekas: `file_exists` does i/o while `@include` does not. If supporting weird xdebug features goes against perf, we won't to it. This should be investiguated.

fightbulc: @bocharsky-bw if the permissions are skewed than file_exists probably can't read it as well, since the permission won't allow it, I guess. @nicolas-grekas I think the case is not about some fancy xdebug feature. The questions comes down to if surpressing errors is in the interest of the user. If a user wishes to see `E_WARNINGS` that should be respected, shouldn't it?

nicolas-grekas: The author expressing failures should be ignored is what should have the highest precedence.

nicolas-grekas:
:-1: on my side because of the I/O check this would add on every single execution.

fabpot:
Let's close then.

PR #672, opened 2019-11-08
[Mercure] Fix enable_profiler node position

chalasr:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

### DIFF
--- a/symfony/mercure-bundle/0.2/config/packages/mercure.yaml
+++ b/symfony/mercure-bundle/0.2/config/packages/mercure.yaml
@@ -1,4 +1,5 @@
 mercure:
+    enable_profiler: '%kernel.debug%'
     hubs:
         default:
             url: '%env(MERCURE_PUBLISH_URL)%'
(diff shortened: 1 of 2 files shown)


dunglas:
Actually `enable_profiler` should be a `0.3` folder, because it's a new flag, and it breaks the stable version.

symfony/mercure-bundle/0.3/manifest.json
```diff
@@ -0,0 +1,31 @@
+{
```
dunglas: This can be a symlink to the file in `0.1`.

chalasr: fixed

stof:
The validation error is because the master branch has alias 1.0.x-dev, not 0.3.x-dev (which looks wrong to me btw, if you plan multiple 0.x releases before reaching 1.0, with BC breaks between them)

stof:
Also, why 0.3 ? There is no 0.2 release yet

dunglas:
You’re right, the next bundle version is 0.2 (I mixed with the library version). And you’re also right about the branch alias. Both the bundle and the lib must be updated.

dunglas:
This PR can now be merged (the CI will be green now).

chalasr:
Green

PR #678, opened 2019-11-15
Dont look recursively into per-env folders

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Double-star discovery is a slow path, yet no recipes use any higher level - and I don't see why one would need this.
Let's gain some perf with this change.

### DIFF
--- a/symfony/framework-bundle/4.2/src/Kernel.php
+++ b/symfony/framework-bundle/4.2/src/Kernel.php
@@ -37,7 +37,7 @@ class Kernel extends BaseKernel
         $confDir = $this->getProjectDir().'/config';
 
         $loader->load($confDir.'/{packages}/*'.self::CONFIG_EXTS, 'glob');
-        $loader->load($confDir.'/{packages}/'.$this->environment.'/**/*'.self::CONFIG_EXTS, 'glob');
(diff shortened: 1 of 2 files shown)


stof:
I think the recipe validator should then add a rule ensuring that no new recipe starts using deeper files (without importing them from another file matched by the glob). What do you think ?

symfony/framework-bundle/4.2/src/Kernel.php
```diff
@@ -37,7 +37,7 @@ protected function configureContainer(ContainerBuilder $container, LoaderInterfa
         $confDir = $this->getProjectDir().'/config';
 
         $loader->load($confDir.'/{packages}/*'.self::CONFIG_EXTS, 'glob');
-        $loader->load($confDir.'/{packages}/'.$this->environment.'/**/*'.self::CONFIG_EXTS, 'glob');
+        $loader->load($confDir.'/{packages}/'.$this->environment.'/*'.self::CONFIG_EXTS, 'glob');
```
yceruto: shouldn't we do the same for 3.3 recipe?

nicolas-grekas: updated for 3.3 also

nicolas-grekas:
> the recipe validator should then add a rule ensuring that no new recipe starts using deeper files

/cc @fabpot ;)

PR #682, opened 2019-11-19
Add recipe for DoctrineBundle 1.12 and newer

alcaeus:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | None

This PR adds a new recipe for DoctrineBundle 1.12 and newer, including 2.0. Changes from 1.16:
* The `underscore` naming strategy is replaced with `underscore_number_aware`. This is introduced in ORM 2.7, where the original naming strategy is deprecated. On ORM 2.6, this will behave the same way as `underscore`
* ~The entire production configuration is removed. `auto_generate_proxy_classes` is `false` by default, and the cache configuration is no longer necessary.~ Brainfart, read below
* The `prod` configuration has been replaced with `dev` which explicitly enables autogenerating proxy classes. By default, they are not generated automatically.

### Changes to caching

With DoctrineBundle 1.11, we introduced support for configuring Symfony cache pools as caches for ORM. In 1.12, the default cache type was changed to use the default `system` and `app` pools, unless otherwise configured. With this change it is no longer necessary to provide a cache configuration. Users can still override the default configuration by defining a separate cache pool or service and using the `pool` and `service` types to configure these as appropriate.


doctrine/doctrine-bundle/1.12/config/packages/doctrine.yaml
```diff
@@ -0,0 +1,23 @@
...
+    orm:
+        naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
+        auto_mapping: true
+        mappings:
+            App:
+                is_bundle: false
+                type: annotation
+                dir: '%kernel.project_dir%/src/Entity'
+                prefix: 'App\Entity'
+                alias: App
```
Pierstoval: Weren't aliases supposed to be removed? Or I might have forgotten about the discussion 🤔 😕

alcaeus: They might be removed in ORM 3.0, but they are still part of the config, so I wouldn't change that part right now.

doctrine/doctrine-bundle/1.12/config/packages/dev/doctrine.yaml
```diff
@@ -0,0 +1,3 @@
+doctrine:
+    orm:
+        auto_generate_proxy_classes: true
```
weaverryan: I think we need to revert back to the `prod` config that sets this to false. Because with this config, the in the `test` environment, the proxy classes would not be automatically generated

alcaeus: Good catch! I've reverted to the previous behaviour, which still makes the prod config significantly shorter!

doctrine/doctrine-bundle/1.12/config/packages/prod/doctrine.yaml
```diff
@@ -0,0 +1,3 @@
+doctrine:
+    orm:
+        auto_generate_proxy_classes: false
```
xabbuh: What about using `%kernel.debug%` as the value in `doctrine/doctrine-bundle/1.12/config/packages/doctrine.yaml` instead?

dmaicher:
Previously without any config we used an in-memory `array` cache. 

Now unfortunately in dev environment persistent caching for meta data is enabled by default :cry:
Means whenever I change something on my entity (like annotations) I need to purge the cache. That's quite annoying.

And for the test environment now I have a persistent result cache (with a default FileSystemAdapter in my case). See https://github.com/doctrine/DoctrineBundle/issues/1066

Should the cache config really be the same for all environments? This means we would have to change the cache pools (app and system) based on the environment?

weaverryan:
> Should the cache config really be the same for all environments? This means we would have to change the cache pools (app and system) based on the environment then I guess

No, this IS something that needs to fixed. I think the cache pools should only be used (by default) in the `prod` environment. I know @alcaeus is aware of the issue - I'm not sure yet what the status us.

PR #683, opened 2019-11-19
Update app.js

noniagriconomie:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | none

I think it is better, isn't it?
Was just testing this bundle with a bare demo project


stof:
Using ES imports is indeed better (some webpack optimizations apply only for ES imports).

For jQuery, it won't change much though, given that jQuery does not ship a module build. But it might have an impact for other packages.

noniagriconomie:
@stof thx for this explanation
i was also looking at the symfony cast where this code was written :)
(https://symfonycasts.com/screencast/webpack-encore)

PR #687, opened 2019-11-21
Simplify Twig config for 5.0 recipe

yceruto:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Follow up https://github.com/symfony/recipes/pull/645#discussion_r321765847:
* the `debug` config does not change often as well as `strict_variables` (but we cannot remove it from 4.4 as its default is being deprecated)
 * the `exception_controller` option is deprecated and does not work in 5.0

About `default_path` I think it's still relevant as it reveals the project dir structure (same as translation recipe).

wdyt?


yceruto:
the symfony-flex-server[bot] failure is a false positive (same https://github.com/symfony/recipes/pull/645#issuecomment-528873582).

PR #691, opened 2019-11-24
[DoctrineBundle] Add charset option to postgres example

chr-hertel:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | 

We should add charset on example DSN otherwise the charset option of `doctrine.yaml` leads to an error

Bildschirmfoto 2019-11-24 um 23 31 36


javiereguiluz:
Could we fix this differently?

If someone opens `config/packages/doctrine.yaml`, they'll see this:

```yaml
doctrine:
    dbal:
        # Only needed for MySQL (ignored otherwise)
        charset: utf8mb4
        # ...
```

Why is the `charset` option only for MySQL? Why should I configure the charset differently depending on the database? This is very confusing to me.

chr-hertel:
@javiereguiluz i'm not deep enough into DoctrineBundle to really explain how the charset is set in a connection, but TBH i think `ignored otherwise` is not completely true

PR #693, opened 2019-11-25
Remove charset config for doctrine-bundle v2

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Requires https://github.com/doctrine/DoctrineBundle/pull/1070 before merging.


doctrine/doctrine-bundle/2.0/config/packages/doctrine.yaml
```diff
@@ -0,0 +1,19 @@
+doctrine:
+    dbal:
+        url: '%env(resolve:DATABASE_URL)%'
+
+        # IMPORTANT: You MUST configure your db driver and server version,
+        # either here or in the DATABASE_URL env var (see .env file)
+        #driver: 'mysql'
+        #server_version: '5.7'
```
fabpot: What about removing everything here? Or put another way: why would one want to configure the driver and version here instead of in the DATABASE_URL env var?

javiereguiluz: Also, the DSN are like: `mysql://...`, `postgresql://...`, `sqlite://...`  So, why do I have to define the `driver` explicitly? It's just the first word of the DSN, right?

stof: you might want to configure the version here in case your PaaS does not expose a serverVersion parameter in the query string (which is a Doctrine DBAL convention). This happens on Heroku for instance.

nicolas-grekas: @dzuelke WDYT about this? Could Heroku include the serverVersion in the DATABASE_URL?

nicolas-grekas: I removed the "driver" line but kept "server_version" for now.

dzuelke: We can't, since we don't offer MySQL, @nicolas-grekas ;) That's probably a question for the folks at ClearDB and JawsDB. I generally think that the server version is more of a "static" environment thing, since your code is probably also written against the specific capabilities of a version series, so it shouldn't differ between environments, unlike the DSN.

nicolas-grekas: That's not only for MySQL - the same is needed for PG :)
But I get your point. Actually, DoctrineBundle/DBAL could throw an exception when both setting are found but don't match (DSN+config). That would be even more useful as if would help spot mistakes.

/cc @alcaeus in case you agree and want to create an issue there :)

alcaeus: See https://github.com/doctrine/dbal/issues/3757.

alcaeus:
@nicolas-grekas does this need a rebase after the changes in #695?

nicolas-grekas:
No rebase needed,this is ready to go.

PR #698, opened 2019-12-02
Disable logging for test env

ro0NL:
| Q             | A
| ------------- | ---
| License       | MIT

I'd argue by default you need the full application logs when running (or writing) tests.

### DIFF
--- a/symfony/monolog-bundle/3.3/config/packages/test/monolog.yaml
+++ b/symfony/monolog-bundle/3.3/config/packages/test/monolog.yaml
@@ -5,8 +5,8 @@ monolog:
             action_level: error
             handler: nested
             excluded_http_codes: [404, 405]
+            channels: ["!event"]
         nested:
             type: stream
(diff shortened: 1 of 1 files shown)


xabbuh:
Doesn't it make sense to at least have the fingers crossed handler then? Otherwise your functional tests fail without given you any clue about what went wrong.

stof:
Logs are indeed quite useful to debug a failure.

ro0NL:
good point, updated.

this reduces the no. of log lines from +-120 by default per testcase, to +- 20 only when failed. Nice :)

last question, is adding `400` to `excluded_http_codes` OK? we test quite a bunch of bad request cases

ro0NL:
hm, ill leave `excluded_http_codes` for now. Most noise seems to come from API platform, as it logs exceptions before applying `exception_to_status` ... so it seems.

need to investigate this part a bit further, other this PR is ready :)

PR #702, opened 2019-12-06
Use configureBabelPresetEnv() in webpack.config.js

ihmels:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

Use `configureBabelPresetEnv()` introduced in symfony/webpack-encore#642 instead of `configureBabel()`.

### DIFF
--- a/symfony/webpack-encore-bundle/1.0/package.json
+++ b/symfony/webpack-encore-bundle/1.0/package.json
@@ -1,6 +1,6 @@
 {
     "devDependencies": {
-        "@symfony/webpack-encore": "^0.28.0",
+        "@symfony/webpack-encore": "^0.28.2",
         "core-js": "^3.0.0",
         "regenerator-runtime": "^0.13.2",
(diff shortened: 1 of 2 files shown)


weaverryan:
@Lyrkan Do you have any strong preference?

Lyrkan:
I'm fine with both ways, as long as we don't also set other non-preset-env Babel options (since it would then need two different calls).

The version of `@symfony/webpack-encore` included in the recipe should however probably be bumped to `^0.28.2` if we go with that one. Even though it doesn't really matter (people using the recipe should get the latest allowed version) the current range also allows `0.28.0` and `0.28.1` in which `Encore.configureBabelPresetEnv()` was not available/didn't work.

ihmels:
> The version of `@symfony/webpack-encore` included in the recipe should however probably be bumped to `^0.28.2` if we go with that one.

Oh, I missed that. I've updated the `package.json`.

PR #703, opened 2019-12-06
Add receipts for twig/*-extra to allow these packages to 'just work'

mcfedr:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/twig/cssinliner-extra/3.0/config/packages/twig_cssinliner_extra.yaml
+++ b/twig/cssinliner-extra/3.0/config/packages/twig_cssinliner_extra.yaml
@@ -1,0 +1,7 @@
+services:
+    _defaults:
+        public: false
+        autowire: true
+        autoconfigure: true
+
+    Twig\Extra\CssInliner\CssInlinerExtension: {}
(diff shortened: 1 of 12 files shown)


Pierstoval:
If you use the TwigExtraBundle, that's already the case, it already does the magic for you: https://github.com/twigphp/twig-extra-bundle/blob/2.x/src/Extensions.php

stof:
These recipes will cause issues when installing twig-extra-bundle, because extensions will then be registered twice.

fabpot:
As already mentioned, the Twig Extra bundle should be used instead.

mcfedr:
ah, wow, i didnt find that anywhere, i guess some links needed in the twig docs. 

Is it possible to have the receipe just do nothing when that bundle is installed? might be a nice way of avoiding a bundle that is basically a receipe, and also lettting me easily just use one of these.

Pierstoval:
> Is it possible to have the receipe just do nothing when that bundle is installed? might be a nice way of avoiding a bundle that is basically a receipe, and also lettting me easily just use one of these.

Well, that's what the bundle actually does 😉  

The bundle does not require the twig extra packages. If the package is installed, the bundle registers it in the container.

PR #704, opened 2019-12-06
enable query profiling for debug

ro0NL:
| Q             | A
| ------------- | ---
| License       | MIT

IMHO this is rather useful to look into when debugging. Assumed disabling is more easy, then enabling ... yet the proposed default should be best of both.

Hinted in https://github.com/symfony/symfony/issues/18235#issuecomment-562600074

### DIFF
--- a/doctrine/doctrine-bundle/1.12/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/1.12/config/packages/doctrine.yaml
@@ -1,11 +1,13 @@
 doctrine:
     dbal:
         url: '%env(resolve:DATABASE_URL)%'
-        profiling_collect_backtrace: '%kernel.debug%'
 
         # IMPORTANT: You MUST configure your server version,
         # either here or in the DATABASE_URL env var (see .env file)
         #server_version: '5.7'
+
+        # backtrace queries in profiler (significantly increases memory usage per request)
+        #profiling_collect_backtrace: '%kernel.debug%'
 
         # only needed for MySQL
         charset: utf8mb4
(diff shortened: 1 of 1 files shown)


stof:
The reason why DoctrineBundle disables it by default is that it increases memory usage a lot.

And if you have this turned on in your worker process, you will face a big problem (well, even collecting queries is already an issue).

doctrine/doctrine-bundle/1.12/config/packages/doctrine.yaml
```diff
@@ -6,6 +6,9 @@ doctrine:
         # either here or in the DATABASE_URL env var (see .env file)
         #server_version: '5.7'
 
+        # backtrace queries in profiler (significantly increases memory usage per request)
```
nicolas-grekas: I would remove the word "significantly", it's not factual.

doctrine/doctrine-bundle/1.12/config/packages/doctrine.yaml
```diff
@@ -6,6 +6,9 @@ doctrine:
         # either here or in the DATABASE_URL env var (see .env file)
         #server_version: '5.7'
 
+        # backtrace queries in profiler (increases memory usage per request)
+        #profiling_collect_backtrace: '%kernel.debug%'
```
nicolas-grekas: another note: I'd suggest putting this after the charset options, as server_version+charset are related and should stay as a group to me

ro0NL: i grouped by commented nodes :sweat_smile: better like this :+1:

PR #705, opened 2019-12-07
Simplify the default Kernel

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Requires https://github.com/symfony/symfony/pull/34873

Removes also the multi file-extensions support. Nobody needs that by default AFAIK. Ppl that do can easily add it back.

### DIFF
--- a/symfony/framework-bundle/5.1/src/Kernel.php
+++ b/symfony/framework-bundle/5.1/src/Kernel.php
@@ -3,9 +3,7 @@
 namespace App;
 
 use Symfony\Bundle\FrameworkBundle\Kernel\MicroKernelTrait;
-use Symfony\Component\Config\Loader\LoaderInterface;
-use Symfony\Component\Config\Resource\FileResource;
-use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 use Symfony\Component\HttpKernel\Kernel as BaseKernel;
 use Symfony\Component\Routing\Loader\Configurator\RoutingConfigurator;
 
@@ -13,42 +11,18 @@ class Kernel extends BaseKernel
 {
     use MicroKernelTrait;
 
-    private const CONFIG_EXTS = '.{php,xml,yaml,yml}';
-
-    public function registerBundles(): iterable
+    protected function configureContainer(ContainerConfigurator $container): void
     {
-        $contents = require $this->getProjectDir().'/config/bundles.php';
-        foreach ($contents as $class => $envs) {
-            if ($envs[$this->environment] ?? $envs['all'] ?? false) {
-                yield new $class();
-            }
-        }
+        $container->import('../config/{packages}/*.yaml');
+        $container->import('../config/{packages}/'.$this->environment.'/*.yaml');
+        $container->import('../config/{services}.yaml');
+        $container->import('../config/{services}_'.$this->environment.'.yaml');
     }
 
-    public function getProjectDir(): string
+    protected function configureRoutes(RoutingConfigurator $routes): void
     {
-        return \dirname(__DIR__);
-    }
-
-    protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader): void
-    {
-        $container->addResource(new FileResource($this->getProjectDir().'/config/bundles.php'));
-        $container->setParameter('container.dumper.inline_class_loader', \PHP_VERSION_ID < 70400);
-        $container->setParameter('container.dumper.inline_factories', true);
-        $confDir = $this->getProjectDir().'/config';
-
-        $loader->load($confDir.'/{packages}/*'.self::CONFIG_EXTS, 'glob');
-        $loader->load($confDir.'/{packages}/'.$this->environment.'/*'.self::CONFIG_EXTS, 'glob');
-        $loader->load($confDir.'/{services}'.self::CONFIG_EXTS, 'glob');
-        $loader->load($confDir.'/{services}_'.$this->environment.self::CONFIG_EXTS, 'glob');
-    }
-
-    protected function configureRouting(RoutingConfigurator $routes): void
-    {
-        $confDir = $this->getProjectDir().'/config';
-
-        $routes->import($confDir.'/{routes}/'.$this->environment.'/*'.self::CONFIG_EXTS, 'glob');
-        $routes->import($confDir.'/{routes}/*'.self::CONFIG_EXTS, 'glob');
-        $routes->import($confDir.'/{routes}'.self::CONFIG_EXTS, 'glob');
+        $routes->import('../config/{routes}/'.$this->environment.'/*.yaml');
+        $routes->import('../config/{routes}/*.yaml');
+        $routes->import('../config/{routes}.yaml');
     }
 }


symfony/framework-bundle/5.1/src/Kernel.php
```diff
@@ -3,52 +3,31 @@
...
+        $container->import('../config/{packages}/'.$this->environment.'/*.yaml');
+        $container->import('../config/{services}.yaml');
+        $container->import('../config/{services}_'.$this->environment.'.yaml');
     }
 
-    protected function configureRouting(RoutingConfigurator $routes): void
+    protected function configureRoutes(RoutingConfigurator $routes): void
     {
-        $confDir = $this->getProjectDir().'/config';
-
```
Taluu: why do you stop relying on the `getProjectDir()` method to have a realpath, and rely on a simpler `../` instead ?

nicolas-grekas: Because it is simpler :) and because it is possible now: paths are resolved relative to the current file by the `ContainerConfigurator`. That was not the case with a `ContainerBuilder`.

symfony/framework-bundle/5.1/src/Kernel.php
```diff
@@ -3,52 +3,26 @@
...
-        $contents = require $this->getProjectDir().'/config/bundles.php';
-        foreach ($contents as $class => $envs) {
-            if ($envs[$this->environment] ?? $envs['all'] ?? false) {
-                yield new $class();
-            }
-        }
+        $container->import('../config/{packages}/*.yaml');
+        $container->import('../config/{packages}/'.$this->environment.'/*.yaml');
+        $container->import('../config/{services}.yaml');
+        $container->import('../config/{services}_'.$this->environment.'.yaml');
```
bastnic: I worked on that part yesterday on a big project.

Why do you need the accolade? By using it, you force the use of Symfony/Finder over the Spl implementation and the `uasort(SortbaleIterator)` is wayyyyyyy much slower that the simpler `uasort(strnatcmp)`. 

If you remove the multiple extensions, you do not need Symfony/Finder at all, and a nice perf improvement. 

The only experct is that it doesn't handle the file/direction absence, which is not supposed to happen..

bastnic: (perf in dev mode only)

nicolas-grekas: For the last two lines, the brackets make the file optional. Since there is only one file to check for them, they don't have any perf impact.
For the first two lines, the brackets are delimiters to define the prefix before the glob pattern starts. Extracting this prefix (`../config/`) is needed to track resources correctly.
Lastly, you mention that brackets trigger a slow path, but that's the case only if your OS doesn't define `GLOG_BRACE`:
https://github.com/symfony/symfony/blob/8ffdef2958d4190994d7540013925b9580e03493/src/Symfony/Component/Config/Resource/GlobResource.php#L103
To my knowledge, only Solaris doesn't define it. If you're on Solaris, I'm sorry for you :)

bastnic: the condition before https://github.com/symfony/symfony/blob/8ffdef2958d4190994d7540013925b9580e03493/src/Symfony/Component/Config/Resource/GlobResource.php#L102 that line, bypass totally this path if there is a bracket in the pattern. So we fallback on the Finder.

nicolas-grekas: `defined('GLOB_BRACE')` is going to be true most of the time so we *do* enter this "if".
Aren't you looking for double stars instead? https://github.com/symfony/recipes/pull/678

Taluu: IIRC, it's skipped if there's a brace AND `GLOB_BRACE` isn't defined, which is the case on solaris according to @nicolas-grekas

symfony/framework-bundle/5.1/src/Kernel.php
```diff
@@ -3,52 +3,26 @@
...
     {
-        $contents = require $this->getProjectDir().'/config/bundles.php';
-        foreach ($contents as $class => $envs) {
-            if ($envs[$this->environment] ?? $envs['all'] ?? false) {
-                yield new $class();
-            }
-        }
+        $container->import('../config/{packages}/*.yaml');
+        $container->import('../config/{packages}/'.$this->environment.'/*.yaml');
+        $container->import('../config/{services}.yaml');
```
BoShurik: What about adding additional import:
```php
$container->import('../config/{services}/**/*.yaml');
```
I've always ended with this as `services.yaml` is growing significantly

symfony/framework-bundle/5.1/src/Kernel.php
```diff
@@ -3,52 +3,26 @@
...
-    public function getProjectDir(): string
+    protected function configureRoutes(RoutingConfigurator $routes): void
     {
-        return \dirname(__DIR__);
-    }
-
-    protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader): void
-    {
-        $container->addResource(new FileResource($this->getProjectDir().'/config/bundles.php'));
-        $container->setParameter('container.dumper.inline_class_loader', \PHP_VERSION_ID < 70400);
```
bendavies: is this done elsewhere then?

nicolas-grekas: in https://github.com/symfony/symfony/pull/34872

symfony/framework-bundle/5.1/src/Kernel.php
nicolas-grekas: I don't think the defaults should provide that much flexibility.
Flexibility is provided by this very file: patch it when you need to.

symfony/framework-bundle/5.1/src/Kernel.php
bastnic: Wow, I'm on an alpine linux, and `GLOB_BRACE` is not defined. Hance the perf decrease.

symfony/framework-bundle/5.1/src/Kernel.php
bastnic: @bocharsky-bw you need to know that this gives a big perofrmance penalty by forcing the use of Symfony/Finder.

symfony/framework-bundle/5.1/src/Kernel.php
bastnic: To be clear, I do have `GLOB_NOSORT` but not `GLOB_BRACE` on alpine. By removing the bracket, I go in the first if, then 0 is passed to the glob method. And everything works just fine. With a difference of ~100ms per request.

bendavies: confirmed no `GLOB_BRACE` on our alpine based prod images. great spot!

nicolas-grekas: OK, so I need to add musl libc to my list of platforms not supporting `GLOB_BRACE`. Too bad.
Maybe we could use `$container->import('../config/packages/*.yaml', 'glob', true);`.
Please give it a try and submit a PR if that works.
An alternative might be to use an implementation that doesn't require the Finder - ie we could have some logic that splits a single glob-with-braces into several of them. PR welcome as usual :)

Not a blocker for this very PR.

bastnic: Absolutely, I opened the discussion in a new ticket https://github.com/symfony/recipes/issues/712

Tobion:
> Removes also the multi file-extensions support. Nobody needs that by default AFAIK. Ppl that do can easily add it back.

Can we be sure that no recipe uses something else than yaml? Even contrib ones? All those would be ignored suddenly.

nicolas-grekas:
IIRC the CI not enforces yaml. IIRC also, I had a look at the repos back then.

symfony/framework-bundle/5.1/src/Kernel.php
```diff
@@ -3,52 +3,26 @@
...
-        $container->setParameter('container.dumper.inline_factories', true);
```
Tobion, line 37: For reference why this got removed: https://github.com/symfony/symfony/pull/36193

jrjohnson: Thank you for this note @Tobion! Super helpful.

PR #707, opened 2019-12-11
Remove ;

fabpot:


symfony/webpack-encore-bundle/1.0/webpack.config.js
```diff
@@ -51,7 +51,7 @@ Encore
     .configureBabelPresetEnv((config) => {
         config.useBuiltIns = 'usage';
         config.corejs = 3;
-    });
+    })
```
fabpot: If we keep `;`, un-commeting anything below breaks the file.

PR #710, opened 2019-12-12
Making symfony/console 4.2 use symfony/framework 4.2's config/bootstrap.php

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | Not needed

This one is a bit tricky :). Currently, if you install all components on (for example) Symfony 4.4, both the `symfony/console` and `symfony/framework-bundle` will add the `config/bootstrap.php` file - but they will add *different* contents: the `symfony/console` recipe ultimately points to `symfony/framework-bundle/3.3/config/bootstrap.php` while the recipe for `symfony/framework-bundle` will use its 4.4 recipe.

This corrects that by adding a `symfony/console` 4.2 recipe that uses `symfony/framework-bundle` 4.2's `config/bootstrap.php` file.

Important points:

*  A) This assumes that the components are upgraded at the same time - e.g. if I have `symfony/console` 4.2 I must also have `symfony/dot-env` 4.2 installed. I think we need to make this assumption for any hope at sane versioning. And, this assumption is already
made (for example) in the framework-bundle 4.2 recipe, whose `config/bootstrap.php`
file references a DotEnv 4.2 method

* B) Any time we create a new "recipe version" for a package where we update a file that is shared by another recipe, we probably *also* need to create a new version of that other recipe, which points to the newly-updated file. For example, when `symfony/framework-bundle/4.2` was added and `config/bootstrap.php` was updated, we should have also created a `symfony/console/4.2` recipe that pointed to the new `config/bootstrap.php` file (that is what this PR does).

Thanks!


weaverryan:
I just tested this - by updating my `symfony/console` recipe only (via the new `composer sync-recipes symfony/console --force -v` which will soon be renamed) from 4.0 to 4.4 and DID get the new (and functional) `config/bootstrap.php` file.

PR #711, opened 2019-12-13
Creating phpunit-bridge 4.2 recipe to point to 4.2 version of bootstrap.php

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | Not needed

Same as #710: if you ran `symfony recipe:install symfony/phpunit-bridge --force -v` on Symfony 4.4, you would actually get the `config/bootstrap.php` file from `symfony/framework-bundle/3.3`.

We need to be VERY careful whenever any modification is made to a file that is shared by other recipes. All those other recipes would need a new version to point to the new file.


weaverryan:
I just tested this deployment by using `symfony recipe:install symfony/phpunit-bridge --force -v` on a Symfony 4.4 project: it correctly used the `config/bootstrap.php` from `symfony/framework-bundle/4.2` (which is the latest `config/bootstrap.php` file).

PR #713, opened 2019-12-18
Enable inline_class_loader in debug mode

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

We forgot about it, but setting inline_class_loader to true allows bypassing `DebugClassLoader` in debug mode, which provides a significant DX improvement, even on PHP 7.4+

### DIFF
--- a/symfony/framework-bundle/4.2/src/Kernel.php
+++ b/symfony/framework-bundle/4.2/src/Kernel.php
@@ -33,7 +33,7 @@ class Kernel extends BaseKernel
     protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader): void
     {
         $container->addResource(new FileResource($this->getProjectDir().'/config/bundles.php'));
-        $container->setParameter('container.dumper.inline_class_loader', \PHP_VERSION_ID < 70400);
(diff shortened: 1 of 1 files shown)


stof:
> We forgot about it, but setting inline_class_loader to true allows bypassing `DebugClassLoader` in debug mode, which provides a significant DX improvement

wouldn't this also make us loose all the DX warnings of the DebugClassLoader ?

nicolas-grekas:
> wouldn't this also make us loose all the DX warnings of the DebugClassLoader ?

the warnings are still triggered when the container is compiled, that's what we figured out at the time when we introduced this behavior.

weaverryan:
> which provides a significant DX improvement, even on PHP 7.4+

What is the difference before & after? I'm not very familiar with the specifics of `inline_class_loader`. This is a DX improvement? Not a performance one?

nicolas-grekas:
Check https://twitter.com/nicolasgrekas/status/929032213815005184
and https://github.com/symfony/symfony/pull/26276/files

stof:
@weaverryan that's a performance feature. The DX improvement is that your development setup is faster :smile:

PR #716, opened 2020-01-02
Update console require file

noniagriconomie:
| Q             | A
| ------------- | ---
| License       | MIT

Hi,
cf https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/4.4/public/index.php#L7 (the other php symfony entry point)

should we do the same here? (or maybe I miss something)

Thank you,

### DIFF
--- a/symfony/console/4.4/bin/console
+++ b/symfony/console/4.4/bin/console
@@ -12,7 +12,7 @@ if (false === in_array(\PHP_SAPI, ['cli', 'phpdbg', 'embed'], true)) {
 
 set_time_limit(0);
 
-require dirname(__DIR__).'/vendor/autoload.php';
+require dirname(__DIR__).'/config/bootstrap.php';
 
 if (!class_exists(Application::class)) {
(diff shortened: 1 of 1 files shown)


jvasseur:
The bootstrap file reads values in `$_SERVER`, this means it needs to be included after we read the `--end` and `--no-debug` options (that modify the content of `$_SERVER`.

noniagriconomie:
@jvasseur thank you for the reply,
but why this difference with the other file `index.php` which does not used `$_SERVER`?

jvasseur:
The difference comes from those lines :
https://github.com/symfony/recipes/blob/5f8acedd41d584611b2422aa3d4d66f97c70c3f2/symfony/console/4.4/bin/console#L22-L28

They needs to be executed between loading the autoloader and loading the bootstrap file. Since they don't exists in `index.php` we can directly load the bootstrap file.

noniagriconomie:
oki understood, lets close this then :)

Issue #574, opened 2019-04-11
Confusing default bootstrapping for "test" env


phansys:
Given the changes made at #491, using [`config/bootstrap.php`](https://github.com/symfony/recipes/blob/c466e7a6cfa12b61a8e73537df518b4008a8f1c7/symfony/framework-bundle/3.3/config/bootstrap.php#L9) as bootstrap of phpunit is not possible if `.env.local.php` exists without variables that closely belongs to the "test" env ("KERNEL_CLASS", "SYMFONY_DEPRECATIONS_HELPER", etc); since the file from where these variables should be loaded (`.env.test`) isn't required at all:

https://github.com/symfony/recipes/blob/c466e7a6cfa12b61a8e73537df518b4008a8f1c7/symfony/phpunit-bridge/3.3/.env.test#L2-L4

Since these variables are not common for the local environment, the dumped array at `.env.local.php` doesn't contains them, making the test suite unusable without changes:

https://github.com/symfony/recipes/blob/c466e7a6cfa12b61a8e73537df518b4008a8f1c7/symfony/framework-bundle/3.3/config/bootstrap.php#L9-L13

nicolas-grekas:
Yes, that's correct. We need to enhance documentation around this as it's quite new.
As you spotted, .env.local.php freezes .envelope files. As such and on purpose, it's only for prod-like needs. Don't use it in test nor dev.

Issue #529, opened 2019-02-12
[FrameworkBundle] forced use of "fake" environment


akomm:
Why are this the only options:
```php
// Load cached env vars if the .env.local.php file exists
// Run "composer dump-env prod" to create it (requires symfony/flex >=1.2)
if (is_array($env = @include dirname(__DIR__).'/.env.local.php')) {
    $_SERVER += $env;
    $_ENV += $env;
} elseif (!class_exists(Dotenv::class)) {
    throw new RuntimeException('Please run "composer require symfony/dotenv" to load the ".env" files configuring the application.');
} else {
    // load all the .env files
    (new Dotenv())->loadEnv(dirname(__DIR__).'/.env');
}
```

Why not for example this?

```php
if (!($_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? null)) {
    if (is_array($env = @include dirname(__DIR__).'/.env.local.php')) {
        $_SERVER += $env;
        $_ENV += $env;
    } elseif (!class_exists(Dotenv::class)) {
        throw new RuntimeException('Please run "composer require symfony/dotenv" to load the ".env" files configuring the application.');
    } else {
        // load all the .env files
        (new Dotenv())->loadEnv(dirname(__DIR__).'/.env');
    }
}
```

Is there a reason not allowing "native" environment to be used I missed? I know, this is all not optimized for docker, because docker is not the only way of doing things, but:

Docker is not the problem! Docker only makes the design flaw visible. The same applies, when you run without docker on a "user per app" basis and the user has environment set to run the app and there might be some task executed that has to run with the same user and share the environment. The environment is here isolated to the user the app runs on. No collisions at all. Now if you do not have this setup, the above would still work with `DotEnv` or dumped .env.local.php.

nicolas-grekas:
Why do you think native env is not supported? It's perfectly supported actually. `Dotenv` is used to load default env vars from `.env*` files sure, but that doesn't prevent using native env at all.
What you propose creates all sort of DX issues people are not comfortable dealing with.

If you want to completely ignore .env files, you can create a `.env.local.php` file that contains `<?php return [];`. But it should be opt-in so that things work as expected by default.

akomm:
@nicolas-grekas 

[...]

PR #719, opened 2020-01-06
Allow missing APP_ENV in .env.local.php

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Reported by @bendavies

### DIFF
--- a/symfony/framework-bundle/3.3/config/bootstrap.php
+++ b/symfony/framework-bundle/3.3/config/bootstrap.php
@@ -6,7 +6,7 @@ require dirname(__DIR__).'/vendor/autoload.php';
 
 // Load cached env vars if the .env.local.php file exists
 // Run "composer dump-env prod" to create it (requires symfony/flex >=1.2)
-if (is_array($env = @include dirname(__DIR__).'/.env.local.php') && ($_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? $env['APP_ENV']) === $env['APP_ENV']) {
+if (is_array($env = @include dirname(__DIR__).'/.env.local.php') && ($_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? $env['APP_ENV'] ?? null) === ($env['APP_ENV'] ?? null))) {
     foreach ($env as $k => $v) {
         $_ENV[$k] = $_ENV[$k] ?? (isset($_SERVER[$k]) && 0 !== strpos($k, 'HTTP_') ? $_SERVER[$k] : $v);
     }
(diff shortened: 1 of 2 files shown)


bendavies:
Thanks!

bendavies:
hmm no, this is no good actually.

I'm not really sure what https://github.com/symfony/recipes/pull/647 was trying to solve, but it https://github.com/symfony/recipes/pull/647 + this PR break this workflow:
```
## DotEnv not installed
rm .env
echo '<?php return [];' > .env.local.php
APP_ENV=prod bin/console
```

This errors because this comparison fails:
```
($_SERVER['APP_ENV'] ?? $_ENV['APP_ENV'] ?? $env['APP_ENV'] ?? null) === ($env['APP_ENV'] ?? null)
// ->
prod === null = false
```

As above, I don't know how to fix this because I honestly can't work out what https://github.com/symfony/recipes/pull/647 was trying to achieve.
@nicolas-grekas can you advise?

Thanks.

guilliamxavier:
<sup>(FTR: `syntax error, unexpected ')'` fixed in #720 and #723)</sup>

@bendavies: As I understand it, the purpose of #647 was to have e.g. `APP_ENV=test bin/console` always load `.env.test` even if `.env.local.php` exists [and contains `'APP_ENV' => 'prod'`].

But maybe it should not have been merged, because #501 clearly said "When `.env.local.php` is found, other .env files are ignored" and [the docs](https://symfony.com/doc/4.4/configuration.html#configuring-environment-variables-in-production) say "After running this command [`composer dump-env prod`], Symfony will load the `.env.local.php` file to get the environment variables and will not spend time parsing the `.env` files".

@nicolas-grekas what do you think, should #647 be reverted?

Besides, Ben's workflow highlights a question: if you only use real environment variables and don't use any `.env` file nor the `Dotenv` component [in prod], you are still required to either create an "empty" (`<?php return [];`) `.env.local.php` or install `symfony/dotenv`, even though you actually don't need them?

nicolas-grekas:
@guilliamxavier see https://github.com/symfony/recipes/pull/719

> if you only use real environment variables and don't use any .env file nor the Dotenv component [in prod], you are still required to either create an empty .env.local.php

yes, that's desired as that's the best tradeoff for everyone.

guilliamxavier:
@nicolas-grekas: 719 was this PR, I guess your meant #727. Indeed, I had missed it (thanks!). But then I also found #574, to which you answered:

> As you spotted, .env.local.php freezes .env files. As such and on purpose, it's only for prod-like needs. Don't use it in test nor dev.

so isn't it paradoxical that you accepted #647 (~ six months later)? The docs aren't crystal clear either, so what is the actually intended behavior?

(As for the second question, thanks for the confirmation. I also found #529.)

nicolas-grekas:
> I guess your meant #727

correct, updated

> isn't it paradoxical that you accepted #647

PR #721, opened 2020-01-10
Use PHP to define the services and routes of the app

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

This PR is here to demonstrate the new capabilities of Symfony 5.1.
Basically, we can now develop a full Symfony app in a single file, like "microframeworks" 
No `services.yaml`, no `routes.yaml` here.

### DIFF
--- a/symfony/framework-bundle/5.1/config/routes.php
+++ b/symfony/framework-bundle/5.1/config/routes.php
@@ -1,0 +1,12 @@
+<?php
+
+use App\Kernel;
+use Symfony\Bundle\FrameworkBundle\Routing\Loader\Configurator\RoutingConfigurator;
+
+return static function (RoutingConfigurator $routes, Kernel $kernel): void {
+
+//    $routes->add('index', '/')
+//        ->controller([App\Controller\DefaultController::class, 'index'])
+//    ;
+
+};
--- a/symfony/framework-bundle/5.1/config/services.php
+++ b/symfony/framework-bundle/5.1/config/services.php
@@ -1,0 +1,59 @@
+<?php
+
+use App\Kernel;
+use Symfony\Component\DependencyInjection\Loader\Configurator as di;
+
+// This file is the entry point to configure your own services.
+// Files in the packages/ subdirectory configure your dependencies.
+
+return static function (di\ContainerConfigurator $container, Kernel $kernel): void {
+
+    // Parameters are configuration that don't need to change depending on the machine where the app is deployed.
+    // see https://symfony.com/doc/current/best_practices/configuration.html#application-related-configuration
+    $container->parameters()
+    //    ->set(...)
+    ;
+
+    $services = $container->services()
+        ->defaults()
+            ->autowire()
+            ->autoconfigure()
+    ;
+
+    // Makes classes in src/ available to be used as services
+    $src = dirname(__DIR__).'/src';
+    $services
+        ->load('App\\', $src)
+        ->exclude([
+            $src.'/DependencyInjection',
+            $src.'/Entity',
+            $src.'/Migrations',
+            $src.'/Tests',
+            $src.'/Kernel.php',
+        ])
+    ;
+
+    // Controllers are imported separately to make sure services can be injected
+    // as action arguments even if you don't extend any base controller class
+    $services
+        ->load('App\\Controller\\', $src.'/Controller')
+        ->tag('controller.service_arguments')
+    ;
+
+    // Add more service definitions when explicit configuration is needed.
+    // Please note that last definitions *replace* previous ones when using $services->set().
+    // It is possible to alter a previously declared definition by using $services->get() instead.
+    $services
+    //    ->set(App\MyService::class)
+    //        ->args([di\ref(App\AnotherService::class)])
+    ;
+
+    if ('test' === $kernel->getEnvironment()) {
+        // When a test case needs access to a service, getting it via
+        // a public alias with the "test." prefix is recommended.
+        $services->public()
+        //    ->alias('test.App\MyService', App\MyService::class)
+        ;
+    }
+
+};
--- a/symfony/framework-bundle/5.1/src/Kernel.php
+++ b/symfony/framework-bundle/5.1/src/Kernel.php
@@ -15,14 +15,26 @@ class Kernel extends BaseKernel
     {
         $container->import('../config/{packages}/*.yaml');
         $container->import('../config/{packages}/'.$this->environment.'/*.yaml');
-        $container->import('../config/{services}.yaml');
-        $container->import('../config/{services}_'.$this->environment.'.yaml');
+
+        if (file_exists(\dirname(__DIR__).'/config/services.yaml')) {
+            $container->import('../config/{services}.yaml');
+            $container->import('../config/{services}_'.$this->environment.'.yaml');
+        } else {
+            $path = \dirname(__DIR__).'/config/services.php';
+            (require $path)($container->withPath($path), $this);
+        }
     }
 
     protected function configureRoutes(RoutingConfigurator $routes): void
     {
         $routes->import('../config/{routes}/'.$this->environment.'/*.yaml');
         $routes->import('../config/{routes}/*.yaml');
-        $routes->import('../config/{routes}.yaml');
+
+        if (file_exists(\dirname(__DIR__).'/config/routes.yaml')) {
+            $routes->import('../config/{routes}.yaml');
+        } else {
+            $path = \dirname(__DIR__).'/config/routes.php';
+            (require $path)($routes->withPath($path), $this);
+        }
     }
 }
--- a/symfony/framework-bundle/5.1/config/services.yaml
+++ b/symfony/framework-bundle/5.1/config/services.yaml
@@ -1,1 +1,0 @@
-../../4.2/config/services.yaml
--- a/symfony/routing/5.1/config/packages
+++ b/symfony/routing/5.1/config/packages
@@ -1,0 +1,1 @@
+../../4.2/config/packages/
--- a/symfony/routing/5.1/manifest.json
+++ b/symfony/routing/5.1/manifest.json
@@ -1,0 +1,6 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "aliases": ["router"]
+}


danabrey:
There is a lot of documentation and tutorial content that will become confusing when this change is made. SymfonyCasts doesn't generally refer to PHP service/route configuration even being a possibility, every example is based on the `.yaml` files. I realise these can be updated but hopefully that transition can be made quickly enough to not put off too many newcomers :)

ricknox:
We need to be sure to update the documentation to avoid confusion.

symfony/framework-bundle/5.1/src/Kernel.php
```diff
@@ -15,14 +16,41 @@ protected function configureContainer(ContainerConfigurator $container): void
...
+
+        // controllers are imported separately to make sure services can be injected
+        // as action arguments even if you don't extend any base controller class
+        $services
+            ->load('App\\Controller\\', './Controller')
+            ->tag('controller.service_arguments');
+
+        // add more service definitions when explicit configuration is needed
+        // please note that last definitions always *replace* previous ones
+        //$services->set(...);
```
yceruto: I like this approach a lot, but I'd prefer a separate `config/services.php` file.

symfony/framework-bundle/5.1/src/Kernel.php
```diff
@@ -15,14 +16,41 @@ protected function configureContainer(ContainerConfigurator $container): void
...
     }
 
     protected function configureRoutes(RoutingConfigurator $routes): void
     {
         $routes->import('../config/{routes}/'.$this->environment.'/*.yaml');
         $routes->import('../config/{routes}/*.yaml');
-        $routes->import('../config/{routes}.yaml');
+
+        $routes->add('hello', '/hello')
```
ricknox: Should we add this to the recipe? If no routes are defined you are being greeted by the new welcome page, right?

nicolas-grekas:
The validation error from the bot is a bug /cc @fabpot :)

symfony/framework-bundle/5.1/src/Kernel.php
nicolas-grekas: That'd be way more confusing. Here, we have full context, see the type hint on the method. This makes autocompletion just work. In a separate file, you loose this context.

symfony/framework-bundle/5.1/src/Kernel.php
nicolas-grekas: PR is draft for a reason ;)

yceruto: Well, the welcome page is shown in the root path `/` only, so there's no conflict, although I'm not sure why we might need this route here. Most of the time we don't worry about the kernel file, I wouldn't want to remove this line every time I create a new project or forget where it came from.

symfony/framework-bundle/5.1/src/Kernel.php
yceruto: what about a PHPdoc `/** @var ContainerConfigurator $container */` ?

nicolas-grekas: Make it a PR to compare :)

yceruto: It would be a pleasure, after you mark it ready :)

nicolas-grekas: I can't, the CI has a bug :)
But consider it ready, use my fork.

symfony/framework-bundle/5.1/src/Kernel.php
```diff
@@ -15,14 +16,41 @@ protected function configureContainer(ContainerConfigurator $container): void
     {
         $container->import('../config/{packages}/*.yaml');
         $container->import('../config/{packages}/'.$this->environment.'/*.yaml');
-        $container->import('../config/{services}.yaml');
```
ogizanagi: I'm not convinced removing the base `services` and `routes` files is a great step forward. I think the less we have to tweak the Kernel in userland, the best it is. Especially I don't find this much friendly for newcomers.

weaverryan:
> There is a lot of documentation and tutorial content that will become confusing when this change is made. SymfonyCasts doesn't generally refer to PHP service/route configuration even being a possibility, every example is based on the .yaml files. I realise these can be updated but hopefully that transition can be made quickly enough to not put off too many newcomers :)

Yea, this is the BIG cost - not to be taken lightly. I think it's at least good to explore though. 

In practice, if we *were* to make a big recipe change like this, it makes more sense on a major version (even if the major version doesn't contain the feature that's required to power the recipe). Then you can say "Oh, in Symfony 6, it's done this new way". For SymfonyCasts, if we made this change for 5.1, then any Symfony 5 screencasts we make between now and May (next 5ish months) will have a big outdated piece. I don't think it's worth it - especially because `routes.yaml` and `services.yaml` are not huge YAML pain-points anymore (as you don't do a lot of coding in them, typically thank to autowiring & annotations).

javiereguiluz:
Thanks for proposing this feature!

I like it a lot ... but I have a minor comment: I agree with Yonel and Maxime ... I like this approach, but I'd prefer to keep the config/services.php and config/routes.php files.

I think it makes things easier to understand, especially for newcomers, and also solves the problem that would make the Kernel class grow too much (up to thousands of lines) when you have lots of routes or services.

symfony/framework-bundle/5.1/src/Kernel.php
nicolas-grekas: The Kernel holds nothing but configuration. The Kernel *is* configuration. Look at it.
I get this is surprising, like anything new and unexpected.
Yet the Kernel is the perfect place to do configuration - that's its only purpose actually.

Like I mentioned to @yceruto, the Kernel is actually much better than a random file in a folder, because *it has context*: the `use`, the type hint of the method, the MicroKernelTrait.

All these are the heart of a better configuration system than work with autosuggestions.

nicolas-grekas:
The Kernel holds nothing but configuration. The Kernel is configuration. Look at it.
I get this is surprising, like anything new and unexpected.
Yet the Kernel is the perfect place to do configuration - that's its only purpose actually.

Like I mentioned to @yceruto, the Kernel is actually much better than a random file in a folder, because it has context: the use, the type hint of the method, the MicroKernelTrait.

All these are the heart of a better configuration system than work with autosuggestions.

Pierstoval:
I also disagree. If one ends up having a big Kernel, it means that on each HTTP request, this "big kernel" is loaded for no reason (loaded, not executed).

Contrary to `services.{yaml,php}` which is loaded only at compile-time.

Also, Kernel is documented and explained as the "app encapsulator". App config is mostly in `config/`.

An alternative to provide such direct example and demonstration would be a special skeleton or generator to create a single-file app with only the Kernel.

symfony/framework-bundle/5.1/src/Kernel.php
nicolas-grekas: Newcomers will be able to discover the framework using autocompletion. I call that more friendly to them :) Also using PHP can be seen as more friendly.
The DI configuration for an app is very thin nowadays, there is no need to build a full architecture to organize it - at least to at the beginning of a project.
(note that I'm specifically talking about the app's config. vendors do need dedicated files, yaml is the best for them)

nicolas-grekas:
I cracked it, the DI+Routes config are now split from the Kernel.

Tobion:
Considering most people are comfortable using YAML for services and routes, I don't see the need to change the default to PHP. If people want to use this, they can easily do so in their app as this PR shows. But I don't think it's a good idea to make this the new default at this point.

mnapoli:
This is awesome!

Programming PHP applications with PHP! No, I am not being ironic ^^ For PHP developers not that familiar with Symfony, this is conceptually simpler to use PHP than learning a Symfony-specific DSL written in YAML. A PHP API is familiar and discoverable easily (with any IDE) without having to keep the docs on a second screen.

It makes sense that experienced Symfony developers might not see the point in this (after all, what is the problem to fix here?), but I think it may go a long way for casual Symfony developers or junior developers.

(by "casual" I mean developers, me included, that know the framework but not enough to remember all the details)

❤️  this PR!

kuraobi:
I think there is more to the debate than just "why do we need YAML anyway?" . I have 2 main concerns with this approach:
- configuring services through PHP opens the door to... anything, including dirty hacks and weird magic code that developers unfamiliar with Symfony (or lazy, or who believe living only once is an excuse to get away with everything) will inevitably come up with. I don't want anyone to have to work on projects were even the DI and config part of the code is crappy magic code, been there already with Laravel. YAML is not perfect but, just like Twig, it provides an adequately constrained subset of features that allow you to do what you need while forcing you to do it in a proper way. Junior devs don't struggle with YAML, they struggle with DI and there is no shortcut through that, they have to learn it.
- looks like we are going to have another pseudo-fluent API where you're supposed to indent your method calls to try to keep track of what you're calling them onto. The tree builder is already the most painful part of Symfony to use/read/review/test, I don't think we should replicate the idea anywhere.

Pierstoval:
I also agree with this feature. It's time to move forward to "proper" conventions!

Yaml has never been a good convention IMO. It's only positive value is the fact that it can be easily read. But indentation, complete lack of documentation and validation, etc. (parsing time? :p ) make it a bad practice at first sight. There's also the fact that newcomers (I've seen that too much since I'm a Symfony teacher) are keen to believe that their Yaml config looks similar when defining extensions configuration, services definitions, parameters, routing, fixtures (using Faker and some fixtures-related libs that use Yaml), ORM mapping, and many other things that we can define in Yaml these days.

Many developers argue that XML is the best choice because of the XSD file that make it documented, auto-completed in ANY code editor (since XML has been here for 22 years and XSD for 16 years).

I'm on the side of the ones that like compromises, and the best compromise to me is to use the native API in PHP. This way, native auto-completion is possible, dynamic loading is possible too (like in Compiler Passes, in the end...), environment-specific config becomes possible in one single file (something impossible in Yaml), etc.

Laravel, in comparison, is one of the most popular frameworks, and there's no sign of Yaml in its default configuration. All of it is plain PHP arrays, which is similar than Yaml in its readability, and some parts are using some OOP like routes. It's not perfect (at all, I said a lot about this in a tweeter feed some time ago), but still, it's PHP, and people are used to it.

The benefits of using PHP are numerous, compared to Yaml which only has one.

symfony/framework-bundle/5.1/config/routes.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+use App\Kernel;
+use Symfony\Component\Routing\Loader\Configurator\RoutingConfigurator;
+
+return static function (RoutingConfigurator $routes, Kernel $kernel): void {
+
+//    $routes->add('index', '/')
+//        ->controller('App\Controller\DefaultController::index');
```
lyrixx: Could we do something to be able to do:
```php
->controller(DefaultController::class, 'index')
```
or
```php
->controller([DefaultController::class, 'index'])
```

symfony/framework-bundle/5.1/config/services.php
```diff
@@ -0,0 +1,50 @@
...
+    $services = $container->services()
+        ->defaults()
+            ->autowire()
+            ->autoconfigure()
+    ;
+
+    // Makes classes in src/ available to be used as services
+    $services
+        ->load('App\\', '../src/*')
+        ->exclude('../src/{DependencyInjection,Entity,Migrations,Tests,Kernel.php}');
```
lyrixx: What is the best:

:+1::
```php
        ->exclude('../src/{DependencyInjection,Entity,Migrations,Tests,Kernel.php}');
```

:smile::
```php
        ->exclude(['../src/DependencyInjection', '../src/Entity', ...,]');
```

:tada::
```php
        ->exclude('../src/DependencyInjection')
        ->exclude('../src/Entity')
        ->exclude('../src/Migrations')
        ->exclude('../src/Tests')
        ->exclude('../src/Kernel.php')
```
---

I really like the last one (:tada:) because:

* it demonstrates it's possible ([ref](https://github.com/symfony/symfony/issues/30865))
* it scales really well
* the diff is not messed up when editing one directory
* it's simpler to find / edit something

symfony/framework-bundle/5.1/config/services.php
```diff
@@ -0,0 +1,50 @@
...
+
+    $services = $container->services()
+        ->defaults()
+            ->autowire()
+            ->autoconfigure()
+    ;
+
+    // Makes classes in src/ available to be used as services
+    $services
+        ->load('App\\', '../src/*')
```
lyrixx: Since we use plain PHP, should it be better to use `__DIR__` ?

symfony/framework-bundle/5.1/config/services.php
```diff
@@ -0,0 +1,50 @@
...
+    $services
+        ->load('App\\Controller\\', '../src/Controller')
+        ->tag('controller.service_arguments');
+
+    // Add more service definitions when explicit configuration is needed.
+    // Please note that last definitions *replace* previous ones when using $services->set().
+    // It is possible to alter a previously declared definition by using $services->get() instead.
+    $services
+    //    ->set('my_service_id')
+    //        ->args([di\ref('another_service')])
```
lyrixx: Why `ref`? I prefer the German Notation (:joy:): `di\reference`

symfony/framework-bundle/5.1/src/Kernel.php
```diff
@@ -17,12 +17,20 @@ protected function configureContainer(ContainerConfigurator $container): void
         $container->import('../config/{packages}/'.$this->environment.'/*.yaml');
         $container->import('../config/{services}.yaml');
         $container->import('../config/{services}_'.$this->environment.'.yaml');
+
+        if (file_exists($path = dirname(__DIR__).'/config/services.php')) {
+            (require $path)($container->withPath($path), $this);
```
lyrixx: IMHO the API will be better with something like
```
$container->loadFromPath($path, $this);
```

symfony/framework-bundle/5.1/src/Kernel.php
```diff
@@ -17,12 +17,20 @@ protected function configureContainer(ContainerConfigurator $container): void
         $container->import('../config/{packages}/'.$this->environment.'/*.yaml');
         $container->import('../config/{services}.yaml');
         $container->import('../config/{services}_'.$this->environment.'.yaml');
+
+        if (file_exists($path = dirname(__DIR__).'/config/services.php')) {
```
lyrixx: Since the recipes contains the file, it's always here.

symfony/framework-bundle/5.1/src/Kernel.php
nicolas-grekas: I'm not sure this makes sense: there is already $container->import() for this purpose. We don't need and API, and we don't *want* to maintain any.
Passing the kernel as the second argument is totally adhoc wiring.

symfony/framework-bundle/5.1/config/services.php
nicolas-grekas: that's already the name in the PHP-DSL, we're not gonna change this in this PR

symfony/framework-bundle/5.1/config/routes.php
nicolas-grekas: 1st proposal doesn't work, 2nd would, works for me.

symfony/framework-bundle/5.1/config/services.php
nicolas-grekas: We should pick the alternative that actually works, I don't remember which one does :)

symfony/framework-bundle/5.1/config/services.php
nicolas-grekas: Both work, it's a matter of preference, that's the very same in the end.

lyrixx: Yes, but it's more clear IMHO

symfony/framework-bundle/5.1/src/Kernel.php
nicolas-grekas: What about existing projects that will adopt the new recipes?

lyrixx: The new file will come together

nicolas-grekas: > `$container->import('../config/{services}.yaml');`

this line is using `{...}` *because* we want it to load `services.yaml` conditionally. I think we want the same there: not forcing ppl to update their kernel if they don't want to use `services.php`.

> The new file will come together

Oh, and this new file will break existing projects, which will have a wtf moment because of it breaking their config. How do we prevent that?

nicolas-grekas: updated

symfony/framework-bundle/5.1/config/services.php
nicolas-grekas: updated

symfony/framework-bundle/5.1/config/services.php
nicolas-grekas: Multiple-excludes don't work: the next one replaces the previous one.
But an array work. updated.

symfony/framework-bundle/5.1/config/routes.php
nicolas-grekas: updated to using `[...::class, 'index']` without `use` statement, because this is a commented example.

nicolas-grekas:
There is no need to dismiss yaml while supporting this PR. Yaml is still going to be the recommended format for configuring any vendor packages because there is just no better alternative right now.

This PR introduces PHP **for the services of the *app* itself**. The benefit of using this format is tighter integration with the IDE, allowing to discover how to config these services using the suggestions provided by the IDE.

This should lower the barrier to entry in this domain.

symfony/framework-bundle/5.1/config/services.php
```diff
@@ -0,0 +1,57 @@
...
+    // as action arguments even if you don't extend any base controller class
+    $services
+        ->load('App\\Controller\\', $src.'/Controller')
+        ->tag('controller.service_arguments');
+
+    // Add more service definitions when explicit configuration is needed.
+    // Please note that last definitions *replace* previous ones when using $services->set().
+    // It is possible to alter a previously declared definition by using $services->get() instead.
+    $services
+    //    ->set('my_service_id')
```
Tobion: isn't that service missing a class?

symfony/framework-bundle/5.1/config/services.php
```diff
@@ -0,0 +1,57 @@
...
+    //        ->args([di\ref('another_service')])
+    ;
+
+    if ('test' === $kernel->getEnvironment()) {
+        // When a test case needs access to a service, getting the it
```
Tobion: typeo: getting the it

fabpot:
@symfony-flex-server review

fabpot:
@symfony-flex-server review

symfony/framework-bundle/5.1/config/services.php
Pierstoval: Could be replaced with something like `App\Mailer\ContactMailer::class` for some boilerplate example with a bit of meaning in terms of business logic, WDYT?

symfony/framework-bundle/5.1/config/services.php
```diff
@@ -0,0 +1,57 @@
...
+        ->defaults()
+            ->autowire()
+            ->autoconfigure()
+    ;
+
+    // Makes classes in src/ available to be used as services
+    $src = dirname(__DIR__).'/src';
+    $services
+        ->load('App\\', $src)
+        ->exclude([
```
sstok: Isn't it more readable to use a glob here?

symfony/framework-bundle/5.1/config/services.php
nicolas-grekas: updated

symfony/framework-bundle/5.1/config/services.php
nicolas-grekas: fixed thanks

symfony/framework-bundle/5.1/config/routes.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+use App\Kernel;
+use Symfony\Component\Routing\Loader\Configurator\RoutingConfigurator;
```
HeahDude: Needs to use the new class :)

nicolas-grekas: updated

symfony/framework-bundle/5.1/config/services.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+use App\Kernel;
+use Symfony\Component\DependencyInjection\Loader\Configurator as di;
+
+// This file is the entry point to configure your own services.
+// Files in the packages/ subdirectory configure your dependencies.
+
+return function (di\ContainerConfigurator $container, Kernel $kernel): void {
```
andersonamuller: perhaps make it static as with the routes

nicolas-grekas: agreed, updated.

symfony/framework-bundle/5.1/config/services.php
```diff
@@ -0,0 +1,57 @@
...
+    //        ->args([di\ref(App\AnotherService::class)])
+    ;
+
+    if ('test' === $kernel->getEnvironment()) {
+        // When a test case needs access to a service, getting it via
+        // a public alias with the "test." prefix is recommended.
+        $services->public()
+        //    ->alias('test.App\MyService', App\MyService::class)
+        ;
+    }
```
Pierstoval: Won't this be a readability issue when projects grow and test overrides become more numerous? Shouldn't we stick to `services_test.php` as usual?

symfony/framework-bundle/5.1/config/routes.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+use App\Kernel;
+use Symfony\Bundle\FrameworkBundle\Routing\Loader\Configurator\RoutingConfigurator;
+
+return static function (RoutingConfigurator $routes, Kernel $kernel): void {
+
+//    $routes->add('index', '/')
+//        ->controller([App\Controller\DefaultController::class, 'index']);
```
HeahDude:
```suggestion
//        ->controller([App\Controller\DefaultController::class, 'index'])
//.   ;
```

symfony/framework-bundle/5.1/config/services.php
```diff
@@ -0,0 +1,57 @@
...
+    $src = dirname(__DIR__).'/src';
+    $services
+        ->load('App\\', $src)
+        ->exclude([
+            $src.'/DependencyInjection',
+            $src.'/Entity',
+            $src.'/Migrations',
+            $src.'/Tests',
+            $src.'/Kernel.php',
+        ]);
```
HeahDude:
```suggestion
        ])
    ;
```

symfony/framework-bundle/5.1/config/services.php
```diff
@@ -0,0 +1,57 @@
...
+
+    // Controllers are imported separately to make sure services can be injected
+    // as action arguments even if you don't extend any base controller class
+    $services
+        ->load('App\\Controller\\', $src.'/Controller')
+        ->tag('controller.service_arguments');
```
HeahDude:
```suggestion
        ->tag('controller.service_arguments')
    ;
```

symfony/framework-bundle/5.1/src/Kernel.php
```diff
@@ -15,14 +15,26 @@ protected function configureContainer(ContainerConfigurator $container): void
     {
         $container->import('../config/{packages}/*.yaml');
         $container->import('../config/{packages}/'.$this->environment.'/*.yaml');
-        $container->import('../config/{services}.yaml');
-        $container->import('../config/{services}_'.$this->environment.'.yaml');
+
+        if (file_exists(\dirname(__DIR__).'/config/services.yaml')) {
```
Pierstoval: I don't really understand this change.

Why not load it with a glob by default for the yaml file so it's loaded only if it exists, and load the php file if it exists?

Same comment for the `routes.php` vs `routes.yaml` file by the way.

symfony/framework-bundle/5.1/config/services.php
nicolas-grekas: that's why I preferred having everything in the Kernel, as that allowed for using methods to split the code :)

But anyway, this can be added later on, on projects that actually need it.

symfony/framework-bundle/5.1/src/Kernel.php
nicolas-grekas: We must one XOR another. Loading both would create big WTF moments.

Pierstoval: Seems legit, I'm just unsure of how devs are going to figure out this new feature if they're really used to the yaml file 🤔

Seems similar to me than when the `.env` files were reorganized in the past

nicolas-grekas:
oups :)

weaverryan:
For the record, while I really like this, in -1 for doing it now. A big change like this should be done on a major release - it’s confusing for important files to disappear from 5.0 to 5.1. It also has huge effects on docs and screencasts. That’s manageable if we do it on a major release - but not in a minor release.

symfony/framework-bundle/5.1/config/services.php
weaverryan: This section isn't needed anymore, as we removed the `services_test.yaml` file (there is that test container where everything is public)

PR #726, opened 2020-01-24
[DoctrineBundle] Enable collect-backtrace for dev env

ottaviano:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | 

Enable collect backtrace of each query for Dev env

### DIFF
--- a/doctrine/doctrine-bundle/1.12/config/packages/dev/doctrine.yaml
+++ b/doctrine/doctrine-bundle/1.12/config/packages/dev/doctrine.yaml
@@ -1,0 +1,4 @@
+doctrine:
+    dbal:
+        # backtrace queries in profiler (increases memory usage per request)
+        #profiling_collect_backtrace: '%kernel.debug%'
--- a/doctrine/doctrine-bundle/2.0/config/packages/dev/doctrine.yaml
+++ b/doctrine/doctrine-bundle/2.0/config/packages/dev/doctrine.yaml
@@ -1,0 +1,4 @@
+doctrine:
+    dbal:
+        # backtrace queries in profiler (increases memory usage per request)
+        #profiling_collect_backtrace: '%kernel.debug%'
--- a/doctrine/doctrine-bundle/1.12/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/1.12/config/packages/doctrine.yaml
@@ -11,8 +11,6 @@ doctrine:
         default_table_options:
             collate: utf8mb4_unicode_ci
 
-        # backtrace queries in profiler (increases memory usage per request)
-        #profiling_collect_backtrace: '%kernel.debug%'
     orm:
         auto_generate_proxy_classes: true
         naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware


Pierstoval:
This has a really big impact on performances and memory usage, I wouldn't recommend to add it in dev at all.

A compromise would be to comment it and add an explanation, but I'm not sure of the added value.

nicolas-grekas:
This has been added to the main doctrine.yaml, commented.

stof:
The performance impact is precisely the reason why it is disabled by default in DoctrineBundle. So I agree about keeping it commented.

ottaviano:
yes, no problem for keeping it commented, the goal of this PR is to show this option. 
And I think the right place for it is in `dev/doctrine.yaml`.
This PR adds it for 3 versions: 1.12, 1.6 and 2.0 (not only for 1.12)

nicolas-grekas:
then please send another PR that move the comment (but doesn't add it to prod/test config files)

ottaviano:
Why another PR? This PR already adds it **only** for dev env and for 3 versions. I can update it for moving the comment, no?

doctrine/doctrine-bundle/1.6/config/packages/dev/doctrine.yaml
```diff
@@ -0,0 +1,3 @@
+doctrine:
+    dbal:
+        profiling_collect_backtrace: true
```
stof, line 3: This option only exists in DoctrineBundle 1.11+, so it should not be in this version of the recipe.

ottaviano: absolutely right, thank you 👍

ottaviano:
updated, status: need review

lyrixx:
1. This Pr does not work

> Unrecognized option "profiling_collect_backtrace" under "doctrine.dbal". Available options are "connections", "default_connection", "types".

It should be under the `connections` keys

```yaml
doctrine:
    dbal:
        default_connection: default
        connections:
            default:
                profiling_collect_backtrace: true
```

2. About the performance impact, I tested it why a real application, in easy admin. I did not notice any **big** performance impact as state by @Pierstoval and @stof. Could you share your numbers?

Mine: 

#### Without the collector enabled

image
image
image
image

#### With the collector enabled

image
image
image
image

Pierstoval:
@lyrixx When I tested this, the cost was at least of twice the time, but maybe some internal improvements have been in Doctrine in the meantime, I'll retry it with two of my projects to see if there's anything new in perfs 👍

lyrixx:
If you look at the code, the diff is about `debug_backtrace`. 

So here a simple bench script that simulate 100K SQL requests, at a depth of 50 nested calls (to bench the CPU)

```php

function empty_function_call()
{
}

function a(int $i = 50) {
    if ($i === 0) {
        $s = microtime(true);
        for ($i=0; $i < 100000; $i++) {
            empty_function_call(DEBUG_BACKTRACE_IGNORE_ARGS);
        }
        printf("void %.2fs\n", microtime(true) - $s);
        printf("void %.2fMb\n", memory_get_peak_usage() / 1024 / 1024);


        $s = microtime(true);
        for ($i=0; $i < 100000; $i++) {
            debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
        }
        printf("backtrace %.2fs\n", microtime(true) - $s);
        printf("backtrace %.2fMb\n", memory_get_peak_usage() / 1024 / 1024);

        return;
    }
    a(--$i);
}

a();
```

Results:

```
void 0.00s
void 0.61Mb
backtrace 0.20s
backtrace 0.61Mb
```

So in term of CPU it's almost the same. I recall we talk of 100K req here! 
And in term of memory it's the same. BUT, here I don't store the trace, so this should be taken in account.
A trace weight highly varies (depth, path length, etc) but it's about 25Kb

If I ran again the previous script with 1K SQL request (which is huge!!!) and I store the trace, I got the following results

```
void 0.00s
void 0.61Mb
backtrace 0.01s
backtrace 21.23Mb
```

So, IMHO, the impact is not so much.

TL;DR: **With 1K SQL requests:  CPU +10ms ; Memory +20Mb**(the values depends on each setup)

ottaviano:
@lyrixx thank you for doing this bench!
@stof, @nicolas-grekas, @Pierstoval should we reopen the discussion about leaving this param commented?

For your first point, I have two projects with exactly: `doctrine: dbal: profiling_collect_backtrace: true` in `dev/doctrine.yaml` and it works fine (doctrine/doctrine-bundle: `1.11.1` and `2.2.0`) 🤔

Pierstoval:
Wow, I don't know what changed, but I checked on a project where an operation has to run more than 900 selects and runs 972 inserts (yep, a bit complex project), the whole operation takes approx 2 seconds of query time, and there's no difference.

And with such amount of queries, I expected the web profiler's Doctrine page to take longer to load, but it's kinda the same time for both.

Great news, LGTM then 👍

Pierstoval (approved):

doctrine/doctrine-bundle/2.0/config/packages/dev/doctrine.yaml
```diff
@@ -0,0 +1,4 @@
+doctrine:
+    dbal:
+        # backtrace queries in profiler (increases memory usage per request)
+        #profiling_collect_backtrace: '%kernel.debug%'
```
OskarStark, line 4: This can be uncommented, as it is activated/deactivated using `'%kernel-debug%'` parameter.

```suggestion
        # backtrace queries in profiler (increases memory usage per request)
        profiling_collect_backtrace: '%kernel.debug%'
```

doctrine/doctrine-bundle/1.12/config/packages/dev/doctrine.yaml
```diff
@@ -0,0 +1,4 @@
+doctrine:
+    dbal:
+        # backtrace queries in profiler (increases memory usage per request)
+        #profiling_collect_backtrace: '%kernel.debug%'
```
OskarStark, line 4: This can be uncommented, as it is activated/deactivated using `%kernel-debug%` parameter.

```suggestion
        # backtrace queries in profiler (increases memory usage per request)
        profiling_collect_backtrace: '%kernel.debug%'
```

nicolas-grekas:

nicolas-grekas:
Sorry, can you please rebase once again and target only the recipe for the latest version of the bundle (which uses `when@` now)

PR #728, opened 2020-01-29
Fix TRUSTED_HOSTS regexp example

guilliamxavier:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Fixes https://github.com/symfony/recipes/pull/494#discussion_r372441028

Actually the direct translation for the old `explode(',', "localhost,example.com")` into regexp would be `['^localhost$', '^example\.com$']` (like in [the docs](https://symfony.com/doc/3.4/reference/configuration/framework.html#trusted-hosts)), but this env var is expected as a single string (by [public/index.php](https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/public/index.php)), hence `'^localhost$|^example\.com$'` (JSON-encoded in the file).

Equivalent alternative: `'^(?:localhost|example\.com)$'` (i.e. `"'^(?:localhost|example\\.com)$'"`), but not sure if better

### DIFF
--- a/symfony/framework-bundle/3.3/manifest.json
+++ b/symfony/framework-bundle/3.3/manifest.json
@@ -15,7 +15,7 @@
         "APP_ENV": "dev",
         "APP_SECRET": "%generate(secret)%",
         "#TRUSTED_PROXIES": "127.0.0.1,127.0.0.2",
-        "#TRUSTED_HOSTS": "'^localhost|example\\.com$'"
+        "#TRUSTED_HOSTS": "'^localhost$|^example\\.com$'"
     },
     "gitignore": [
         "/.env.local",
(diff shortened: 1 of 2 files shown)


symfony/framework-bundle/3.3/manifest.json
```diff
@@ -15,7 +15,7 @@
         "APP_ENV": "dev",
         "APP_SECRET": "%generate(secret)%",
         "#TRUSTED_PROXIES": "127.0.0.1,127.0.0.2",
-        "#TRUSTED_HOSTS": "'^localhost|example\\.com$'"
+        "#TRUSTED_HOSTS": "'^localhost$|^example\\.com$'"
```
nicolas-grekas: better use brackets instead?

guilliamxavier: as explained in the PR description, I'm not sure which is better (w.r.t. the docs and how users will edit it with real values), so if you ask, I answer "please decide" :)

nicolas-grekas: I prefer brackets, it's more common

guilliamxavier: :+1: updated (with a simple group, without `?:`, like in the docs and because "more common" too)

Issue #718, opened 2020-01-03
move doctrine migrations away from src


ro0NL:
it was briefly discussed here https://github.com/symfony/recipes/pull/172#issuecomment-328376549, but seems to be ignored eventually

i like to propose to move `src/Migrations` to `migrations/`. I dont understand why we are moving mountains to keep it in src (extra config etc.)  :/ is there any technical blocker to not move it away from src?

semantically it does not belong in src IMHO.

(https://www.tomasvotruba.cz/blog/2019/12/23/5-things-i-improve-when-i-get-to-new-repository/#3-move-code-to-src)

ogizanagi:
I'm already doing this for all of my projects. So 👍 from me.

ro0NL:
alternatively to introduce a more generic/reusable directory in our landscape: `./scripts/database-migrations/`? But im not sure it's valid UNIX style let's say.

We should carefully think about root directories though, at some point it gets out of control.

ogizanagi:
The more generic/reusable proposition [was made a long-time ago](https://github.com/symfony/recipes/pull/253#issuecomment-345467929), but without enough traction/consensus to happen: a `res` directory which would contains any resources / non "runtime" source code of the application, i.e:

- migrations
- assets
- fixtures
- templates
- docs
- translations
- ...

ro0NL:
:heart_eyes: can we do that? please? :grin:

yceruto:
I like `migrations/` at the root too :+1:

ro0NL:
@javiereguiluz ive reread the issue real quick, fair enough :) the ship has sailed. I still think it's more concise in general, but :+1: for many root dirs tied to SF, in a SF specific project dir.

dmaicher:
Also related discussion: https://github.com/doctrine/DoctrineMigrationsBundle/pull/295/files#r362401350

PR #729, opened 2020-01-30
Recipe for upcoming doctrine/migrations 3.0 and 2.2

goetas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Closes https://github.com/symfony/recipes/issues/718

### DIFF
--- a/doctrine/doctrine-migrations-bundle/2.2/config/packages/doctrine_migrations.yaml
+++ b/doctrine/doctrine-migrations-bundle/2.2/config/packages/doctrine_migrations.yaml
@@ -1,0 +1,5 @@
+doctrine_migrations:
+    migrations_paths:
+        # namespace is arbitrary but should be different from App\Migrations
+        # as migrations classes should NOT be autoloaded
+        'DoctrineMigrations': '%kernel.project_dir%/migrations'
--- a/doctrine/doctrine-migrations-bundle/2.2/manifest.json
+++ b/doctrine/doctrine-migrations-bundle/2.2/manifest.json
@@ -1,0 +1,10 @@
+{
+    "bundles": {
+        "Doctrine\\Bundle\\MigrationsBundle\\DoctrineMigrationsBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/",
+        "migrations/": "migrations/"
+    },
+    "aliases": ["doctrine-migrations", "migrations"]
+}


doctrine/doctrine-migrations-bundle/3.0/config/packages/doctrine_migrations.yaml
```diff
@@ -0,0 +1,5 @@
+doctrine_migrations:
+    migrations_paths:
+        # namespace is arbitrary but should be different from App\Migrations
+        # as migrations classes should NOT be autoloaded
+        'App\DoctrineMigrations': '%kernel.project_dir%/Migrations'
```
OskarStark: Shouldn't we keep the same namespace as before?
https://github.com/symfony/recipes/blob/24cc9f60acdb5fd2331b588152acc6c6b2a18cc6/doctrine/doctrine-migrations-bundle/1.2/config/packages/doctrine_migrations.yaml#L5

goetas: There was this https://github.com/doctrine/DoctrineMigrationsBundle/pull/278#discussion_r331081231 conversation suggesting to change it to more recent practices as using the app name space

OskarStark: It in this case they should not be under App\ neither App\Migrations (comment) nor App\DoctrineMigr.... 

TLDR:
Should the comment talk about App\DoctrineMigrat... too? 🧐

goetas: I do not see any bc breaks with it since recipes are mainly used for new installations. Do you have a particular reason to keep the old namespace?

BoShurik: Maybe use `migrations` not `Migrations` as all other root dirs

goetas: Shouldn't this be inside the src dir? Did I choose the wrong kernel parameter?

BoShurik: If you want it inside `src` it should be:
`%kernel.project_dir%/src/Migrations`

BoShurik: See previous [recipe](https://github.com/symfony/recipes/blob/master/doctrine/doctrine-migrations-bundle/1.2/config/packages/doctrine_migrations.yaml#L2)

goetas: Hmm, you are right!

Now you have raised an interesting point, does it make sense to have migrations outside of the src? 

BTW, to be honest I do not remember why migrations should not be autoloadable... To not polute the class map?

BoShurik: See previous discussions https://github.com/symfony/recipes/pull/87#issuecomment-305150705 (for `src `) and https://github.com/symfony/recipes/issues/718 (for `root`)
I personally vote for `root` and already change it in my projects

> to be honest I do not remember why migrations should not be autoloadable

IIRC, to prevent using it in the application

nicolas-grekas:
Hello, how can we move forward here?
Creating a new SF5.1 website-skeleton fails currently because this patch (or a similar one) is missing.
Sorry, I'm not deep enough into the topic to be helpful here...

goetas:
The reason why is failing is related to https://github.com/symfony/orm-pack/issues/18 and https://github.com/symfony/orm-pack/pull/22

goetas:
Orm pack should have a major release each time a major release is merged into doctrine

goetas:
As a quick solution you can revert https://github.com/symfony/orm-pack/pull/22 and address some of the concerns I've expressed into my comments

nicolas-grekas:
The reason is that I installed doctrine/annotations v3 :)
The pack is just ... a pack

goetas:
The reason is that I installed doctrine/doctrine-migrations-bundle v3 :)
The pack is just ... a pack

Ah... my bad. :+1: 

doctrine/doctrine-migrations-bundle v3 is not yet released, so some error can still happen

goetas:
applied proposal in https://github.com/symfony/recipes/issues/718, now migrations are in root `migrations`

doctrine/doctrine-migrations-bundle/3.0/config/packages/doctrine_migrations.yaml
goetas: ive kept the same ns

goetas: migrations are as suggested in https://github.com/symfony/recipes/issues/718

nicolas-grekas:
@goetas do you consider it still WIP? If not can you please update the PR title. to confirm?

goetas:
@nicolas-grekas  this is ready

goetas:
What are the next steps to see this merged?

GPHemsley:
It looks like the recipe for 3.0 didn't actually make it into this PR? https://github.com/doctrine/DoctrineMigrationsBundle/blob/3.0.x/UPGRADE.md reports changes that are not covered here.

PR #731, opened 2020-01-31
Replace .gitignore by a .gitkeep

Th3Mouk:
I just install a new Symfony project with `composer create-project symfony/skeleton`

Add I found a `.gitignore` in the `src/Controller` folder that is empty.

IMO the usage of `.gitkeep` is more appropriate here.


fabpot:
This has been discussed in the past. I don't find the discussion right now, but we won't change the convention. Thank you for proposing.

Th3Mouk:
Thanks for your time @fabpot.

If you find it, I'd be happy to read it.

PR #747, opened 2020-03-24
Set NODE_ENV var when running Encore scripts

javiereguiluz:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

In a project I needed to enable PurgeCSS only when in production. This is done in postcss.config.js and there's no way to know if Encore is running in production or not. I asked on Slack and @lyrkan proposed this solution to me. It worked for me nicely!

As I think this could be useful to others, I propose this change in case it makes sense to make it in general for all projects.

### DIFF
--- a/symfony/webpack-encore-bundle/1.0/package.json
+++ b/symfony/webpack-encore-bundle/1.0/package.json
@@ -8,9 +8,9 @@
     "license": "UNLICENSED",
     "private": true,
     "scripts": {
-        "dev-server": "encore dev-server",
-        "dev": "encore dev",
-        "watch": "encore dev --watch",
-        "build": "encore production --progress"
+        "dev-server": "NODE_ENV=development encore dev-server",
+        "dev": "NODE_ENV=development encore dev",
+        "watch": "NODE_ENV=development encore dev --watch",
+        "build": "NODE_ENV=production encore production --progress"
     }
 }
--- a/symfony/webpack-encore-bundle/1.0/package.json
+++ b/symfony/webpack-encore-bundle/1.0/package.json
@@ -8,9 +8,9 @@
     "license": "UNLICENSED",
     "private": true,
     "scripts": {
-        "dev-server": "NODE_ENV=development encore dev-server",
-        "dev": "NODE_ENV=development encore dev",
-        "watch": "NODE_ENV=development encore dev --watch",
-        "build": "NODE_ENV=production encore production --progress"
+        "dev-server": "encore dev-server",
+        "dev": "encore dev",
+        "watch": "encore dev --watch",
+        "build": "encore production --progress"
     }
 }
--- a/symfony/webpack-encore-bundle/1.0/webpack.config.js
+++ b/symfony/webpack-encore-bundle/1.0/webpack.config.js
@@ -1,7 +1,7 @@
 var Encore = require('@symfony/webpack-encore');
 
-// Optional: sync the NODE_ENV value with the current webpack environment
-// It's useful when you configure tools like PostCSS in their own config files instead of in webpack.config.js.
+// Optional: sync the NODE_ENV value with the current Webpack Encore environment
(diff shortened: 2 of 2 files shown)


symfony/webpack-encore-bundle/1.0/package.json
```diff
@@ -8,9 +8,9 @@
...
     "private": true,
     "scripts": {
-        "dev-server": "encore dev-server",
-        "dev": "encore dev",
-        "watch": "encore dev --watch",
-        "build": "encore production --progress"
+        "dev-server": "NODE_ENV=development encore dev-server",
+        "dev": "NODE_ENV=development encore dev",
+        "watch": "NODE_ENV=development encore dev --watch",
+        "build": "NODE_ENV=production encore production --progress"
```
nicolas-grekas: how does this work when the var already exists?
should it be `NODE_ENV=${NODE_ENV:-production}`?

Lyrkan: For portability reasons something like [`cross-env`](https://github.com/kentcdodds/cross-env#readme) should also probably be used.

weaverryan: Yes, if we do this, then we need to use `cross-env`. I was always hoping to avoid that. 

An alternative might be documentation? Couldn't we advise people to set this env var (of some other env var) inside `webpack.config.js` if they want to? Like:

```
// webpack.config.js
process.env. NODE_ENV = Encore.isProduction() ? 'production' : 'development';
```

WDYT?

javiereguiluz: I made some changes according to your comments. What do you think? Thanks.

Pierstoval: What question me the most is: will this be compatible with all platforms?

symfony/webpack-encore-bundle/1.0/webpack.config.js
```diff
@@ -1,5 +1,9 @@
 var Encore = require('@symfony/webpack-encore');
 
+// Optional: sync the NODE_ENV value with the current webpack environment
+// It's useful when you configure tools like PostCSS in their own config files instead of in webpack.config.js.
```
weaverryan, line 4: I’d change this to:

It’s useful if you need to read the environment from some other config file

The key difference is that I want people to know that this change doesn’t *cause* any behavior to change, it’s just something you can use if you want yourself.

javiereguiluz: I made the change you proposed. But please, note that if you think this is a rare edge-case and that we shouldn't add this to the config file because it will be confusing to newcomers ... that's perfectly find for me.

You are an expert on Webpack/Encore/JS/etc. and I'm not ... so your decision is the important one. Thanks!

javiereguiluz:
@Lyrkan @weaverryan what do you think about this proposal after the latest changes that you suggested? Should we add this or close it as "won't fix"? Thanks!

noniagriconomie (approved):

symfony/webpack-encore-bundle/1.0/webpack.config.js
```diff
@@ -1,5 +1,9 @@
 var Encore = require('@symfony/webpack-encore');
 
+// Optional: sync the NODE_ENV value with the current Webpack Encore environment
+// It’s useful if you need to read the Webpack Encore environment from some other config file
+// process.env.NODE_ENV = Encore.isProduction() ? 'production' : 'development';
```
stof, line 5: this should be moved after the `Manually configure the runtime environment` case

stof (changes requested):

seb-jean:
I love this idea :)

nicolas-grekas:
Please rebase so that the new checks can run.
See also @stof's comment.

PR #749, opened 2020-04-03
Comment deprecation logs (Monolog)

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

I suggest that we comment the logging of deprecations in production. Depending on your application, it can fill your logs pretty fast, and we don't want such a thing to happen by default. Logging deprecations in prod should be a conscious decision IMHO.


fabpot:
merge now

PR #751, opened 2020-04-06
[easy-log-handler] remove recipe

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT

@javiereguiluz maybe you know about this?
@fabpot otherwise?
This looks scary for now :)


javiereguiluz:
Instead of removing this FIXME ... what if we:

* Remove this recipe entirely
* Deprecate the package

I created this package because log files were a mess for me ... but recently PhpStorm added syntax highlighting and other improvements to logs ... so I no longer need this package. What do you think?

nicolas-grekas:
Works for me, recipe is now removed.

nicolas-grekas:
Here is the sidekick PR to merge+tag: https://github.com/symfony/debug-pack/pull/11

Nyholm:
Could you please elaborate why this is removed?

B-Galati:
@javiereguiluz could you share with us the way you're going through logs with PhpStorm? It looks very interesting :-) :pray:

javiereguiluz:
@B-Galati the best of all is that I don't do anything special :)  What I tried to say in [my previous comment](https://github.com/symfony/recipes/pull/751#issuecomment-609857218) is that in the past, PhpStorm showed logs as an endless stream of black & white text. I couldn't find anything there.

But in recent versions, logs are shown like this (by default, no plugins installed):

phpstorm-logs

For me this is more than enough. The highlighting makes everything very easy to find. I know that the Monolog handler that we removed is a bit better sometimes because it formats some information in several lines, etc. But 95% of the times, PhpStorm is enough for me. That's why I suggested removing this.

B-Galati:
Thanks @javiereguiluz, super interesting :+1: 

(It was not working for me because I had Ideolog plugin enabled)

B-Galati:
I configured Ideolog and it's also giving me something easier to read FYI:

image

cristi-contiu:
For projects started before this release and using the EasyLogHandler from the official [symfony/debug-pack](https://github.com/symfony/debug-pack), this will actually break the dev build after running `composer update` , as the `easycorp/easy-log-handler` package will be removed but not the config file added by the original recipe (which references the `EasyCorp\EasyLog\EasyLogHandler` service).

I am also using the latest PhpStorm 2020.1 but I have no syntax highlighting on the logs, with our without Ideolog.

javiereguiluz:
@cristi-contiu I'm using PhpStorm 2019.3 and I haven't installed any plugin related to logs. In fact, I only have 4 custom plugins, so that's why I think this must be some built-in behavior:

image

cristi-contiu:
@javiereguiluz Thank you very much for your reply, I managed to fix it!

In case anyone else has the same issue, in my case the `.log` extension was associated with "ANSI Aware" file type the from the "ANSI Highlighter" plugin. Removing the association from PHPStorm Settings > Editor > File Types > ANSI Aware or uninstalling the plugin will do the trick.

Before (notice the `.dev.log` file icon vs the `.env` icon):
image

After:
image

ostrolucky:
PHPStorm 2020.1, removed association of *.log for ANSI Aware plugin, still can't see behaviour you are describing

Screenshot 2020-04-26 at 15 43 10

cristi-contiu:
@ostrolucky Do you see the color scheme on the `.env` file?

ostrolucky:
@cristi-contiu WDYM? If I have some sort of syntax highlighting in .env file? No, I don't

cristi-contiu:
@ostrolucky My bad, I thought that the `.env` files color scheme was the same as the `.log` files. It turns out the `.env` file color scheme was given by a PHPStorm plugin.

As a test, I disabled all downloaded plugins and still got syntax highlighting for `.log` files (but not in `.env` files). I left all bundled plugins enabled.

In my case, the `.log` files are identified as Text file type (using the `*.log` pattern), but even if I remove the association, I still get highlighting. My color scheme is Darcula (but the highlighting also work with Classic Light and other themes).

dkarlovi:
@ostrolucky you might have Ideolog plugin installed. Interestingly, it will not detect the log format here, if I **uninstall** it, it starts working for me.

ostrolucky:
Fresh PC and installation, here is how it looks first time opening these log files for me:

Screenshot 2020-06-15 at 17 13 41

Installing either of these plugins doesn't help. I really don't know how this works for everybody else out of the box :(

PR #765, opened 2020-04-25
Use PHP to define the services and routes of the app

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

This is #721 again, after it's been reverted in #750.
Let's build a plan to make this happen smoothly.


nicolas-grekas:
I've realized we can drop the services and routes files entirely! See #872

PR #769, opened 2020-05-01
Update mattermost-notifier: add documentation link

thePanz:
Mattermost notifier: add documentation link and use `CHANNEL_ID` as example value for channel, to avoid confusions with the parameter value needed there


thePanz:
Thank you @fabpot

PR #770, opened 2020-05-04
Property info

Saphyel:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | https://symfony.com/doc/current/components/property_info.html

### DIFF
--- a/symfony/property-info/4.0/config/packages/property_info.yaml
+++ b/symfony/property-info/4.0/config/packages/property_info.yaml
@@ -1,0 +1,3 @@
+framework:
+    property_info:
+        enabled: true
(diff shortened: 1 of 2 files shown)


stof:
I don't think that's necessary. The default value of the `enabled` flag is a smart default depending on whether the package is installed.

Saphyel:
I found sometimes with prod config that is not always enabled. In dev mode is as you said

stof:
@Saphyel in prod mode, be careful to clear your cache. But the logic for the default is the same.

nicolas-grekas:
Closing as explained by @stof

Issue #497, opened 2018-11-22
--env no longer equivalent to APP_ENV


bendavies:
Related to https://github.com/symfony/recipes/pull/491

These two should be equivalent, shouldn't they?

```bash
> APP_ENV=prod APP_DEBUG=0 bin/console cache:clear --no-warmup
> bin/console cache:clear --no-warmup --env=prod --no-debug
```

Reproducer:

```bash
> composer create-project symfony/skeleton skeleton 
> cd skeleton
> rm -rf vendor
> composer install --no-dev #error will be here too.
> bin/console cache:clear --no-warmup --env=prod --no-debug

> Fatal error: Uncaught RuntimeException: The "APP_ENV" environment variable is not set to "prod". Please run "composer require symfony/dotenv" to load the ".env" files configuring the application. in project/src/.bootstrap.php on line 13
```

Thanks

nicolas-grekas:
We did that on purpose: defining the real APP_ENV to prod means opting out from loading .env files.
--env prod loads .env file. The fatal error is bad though :) We should replace it with a better behavior, and also merge https://github.com/symfony/skeleton/pull/122 IMHO (and you can move Dotenv to "require" meanwhile).
ping @weaverryan FYI, WDYT?

bendavies:
right, so --env and --no-debug are **not** equivalent to APP_ENV and APP_DEBUG any more?
That's a pretty big gotcha.

nicolas-grekas:
Hum, the exception is right, you should require dotenv.
We should move it by default to "require" to me, please support https://github.com/symfony/recipes/issues/497 if you agree.

bendavies:
I'm afraid I can't agree since (repeating myself - sorry), it seems really poor from a DX perspective that `--env` and `APP_ENV` are not equivalent, as @Tobion pointed out.

It seems like breaking change to me. what is the BC policy here?

FYI, we had mostly migrated an app to completely use env vars, but had some older supervisord jobs using `--env=prod`, and they broke as soon as this change was deployed.

nicolas-grekas:
I understand the current behavior might not be unanimously OK, but the BC break argument makes no sense to me: you don't apply new recipes on existing code automatically. It's your responsibility to commit the resulting file and once done, they won't change unless you accept the change.

bendavies:
I get what you are saying about BC, but from my point of view:

[...]

PR #771, opened 2020-05-06
fix(console): stop to require autoload twice

Oliboy50:
| Q             | A
| ------------- | ---
| License       | MIT 

- because `config/boostrap.php` also requires autoload
- no need to apply this in `5.1` recipe because `config/bootstrap.php` has already been replaced with `Dotenv::bootEnv()`


nicolas-grekas:
`.env` files must be loaded *after* the logic on L21 to L28

This PR reintroduces https://github.com/symfony/recipes/issues/497

I'm therefor closing, thanks for submitting.

PR #775, opened 2020-05-28
Update conf to log deprecations in a dedicated channel

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | https://github.com/symfony/symfony-docs/pull/13642

Follows https://github.com/symfony/symfony/pull/36621


symfony/framework-bundle/5.1/config/packages/prod/deprecations.yaml
```diff
@@ -0,0 +1,8 @@
+# Uncomment to log deprecations in a dedicated channel when using monolog-bundle
+#monolog:
```
Nyholm, line 2: Why not keep this in the recipe of the Monolog-bundle?

As far as I can tell, this is not useable if the monolog-bundle is not installed.

nicolas-grekas: Because this works only when fwb 5.1 is installed, and we have no way to bind a recipe to two packages at the same time. I thought putting this in fwb is the best place, especially because it's commented.

Nyholm: What about the other way around?

Put the comments in the monolog bundle and say "This only works on Framework bundle 5.1"

I'm fine with the changes. I let next reviewer give their opinion about this. Im 50/50.

nicolas-grekas: > But... that channels: [deprecation] config sorta ruins things

that's why I put this in a separate file yes.
I can put the file in the recipe for mobolog yes.

nicolas-grekas: updated,please check again

GPHemsley:
What effect does this have on Symfony versions prior to 5.1, such as LTS 4.4?

l-vo:
There is no effect for versions before 5.1. If you enable the configuration to log deprecation in a separate file for these versions, the file should stay empty since nothing is sent to the `deprecation` channel.

GPHemsley:
> There is no effect for versions before 5.1. If you enable the configuration to log deprecation in a separate file for these versions, the file should stay empty since nothing is sent to the `deprecation` channel.

That makes sense for the new (commented out) code, but what about the code that was _removed_ as part of this PR? If I want to log deprecations in 4.4, is it advised that I keep that intact?

l-vo:
Yes, you can keep the old code, it will work in 4.4 and also in next versions. Anyway the recipe update should not be proposed in 4.4.

GPHemsley:
> Yes, you can keep the old code, it will work in 4.4 and also in next versions.
OK, thanks.

> Anyway the recipe update should not be proposed in 4.4.
Does that mean there's a bug? Because it is being proposed in 4.4.

l-vo:
Oh indeed, good catch !

`deprecation_filter` config should not be removed, I think it's related to the hesitation about [where to put the recipe](https://github.com/symfony/recipes/pull/775#discussion_r431798948).

Do you want to create a PR for fixing it ? I would do it otherwise.

GPHemsley:
@l-vo It's all yours.

PR #781, opened 2020-06-08
Add secret env to framework configuration

alexander-schranz:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

The `APP_SECRET` is defined but never set to the framework configuration.

### DIFF
--- a/symfony/framework-bundle/5.1/config/packages/framework.yaml
+++ b/symfony/framework-bundle/5.1/config/packages/framework.yaml
@@ -1,4 +1,5 @@
 framework:
+    secret: '%env(APP_SECRET)%'
     #csrf_protection: true
     #http_method_override: true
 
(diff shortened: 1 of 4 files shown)


symfony/framework-bundle/5.1/config/packages/framework.yaml
```diff
@@ -1,4 +1,5 @@
 framework:
+    secret: '%env(APP_SECRET)%'
```
nicolas-grekas, line 2: with this change, the while `config/packages` folder becomes identical to the one in 4.2
we shoud thus put a symlink to the directory in the recipe for 5.1

alexander-schranz: done!

nicolas-grekas:
I meant the whole config/packages directory should be linked :)

alexander-schranz:
@nicolas-grekas oh sorry I see. its fixed now!

PR #787, opened 2020-06-13
Leverage the new Runtime component

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

This recipe-update is a sidekick of https://github.com/symfony/symfony/pull/36652
As you can see, much of the logic has been removed from the recipe and moved to vendor/. This means it will be easier to upgrade such apps.

### DIFF
--- a/symfony/console/5.3/bin/console
+++ b/symfony/console/5.3/bin/console
@@ -1,0 +1,17 @@
+#!/usr/bin/env php
+<?php
+
+use App\Kernel;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+
+if (!is_file(dirname(__DIR__).'/vendor/autoload_runtime.php')) {
+    throw new LogicException('Symfony Runtime is missing. Try running "composer require symfony/runtime".');
+}
+
+require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
+
+return function (array $context) {
+    $kernel = new Kernel($context['APP_ENV'], (bool) $context['APP_DEBUG']);
+
+    return new Application($kernel);
+};
--- a/symfony/console/5.3/manifest.json
+++ b/symfony/console/5.3/manifest.json
@@ -1,0 +1,6 @@
+{
+    "copy-from-recipe": {
+        "bin/": "%BIN_DIR%/"
+    },
+    "aliases": ["cli"]
+}
--- a/symfony/framework-bundle/5.3/public/index.php
+++ b/symfony/framework-bundle/5.3/public/index.php
@@ -1,22 +1,9 @@
 <?php
 
 use App\Kernel;
-use Symfony\Component\Dotenv\Dotenv;
-use Symfony\Component\ErrorHandler\Debug;
-use Symfony\Component\HttpFoundation\Request;
 
-require dirname(__DIR__).'/vendor/autoload.php';
+require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
 
-(new Dotenv())->bootEnv(dirname(__DIR__).'/.env');
-
-if ($_SERVER['APP_DEBUG']) {
-    umask(0000);
-
-    Debug::enable();
-}
-
-$kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
-$request = Request::createFromGlobals();
-$response = $kernel->handle($request);
-$response->send();
-$kernel->terminate($request, $response);
+return function (array $context) {
+    return new Kernel($context['APP_ENV'], (bool) $context['APP_DEBUG']);
+};


symfony/console/5.2/bin/console
```diff
@@ -0,0 +1,14 @@
...
+
+if (!is_file(dirname(__DIR__).'/vendor/symfony/runtime/autoload.php')) {
+    throw new LogicException('Symfony Runtime is missing. Try running "composer require symfony/runtime".');
+}
+
+$kernel = require dirname(__DIR__).'/public/index.php';
+
+return function (array $context) use ($kernel) {
+    return new Application($kernel($context));
+};
```
weaverryan, line 14: What about console only apps? It’s an edge case, but I had a friend recently build a symfony app without a public directory. It just looks weird to have the console depend on the front controller.

symfony/framework-bundle/5.2/public/index.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+use App\Kernel;
+use Symfony\Component\HttpFoundation\Request;
+
+require_once ($_SERVER['APP_VENDOR_DIR'] = dirname(__DIR__).'/vendor').'/symfony/runtime/autoload.php';
```
weaverryan, line 6: Sorry if I missed it - how is the APP_VENDOR_DIR used? I mean, why do we need to set it?

symfony/framework-bundle/5.2/public/index.php
```diff
@@ -0,0 +1,18 @@
...
+return function (array $context) {
+    if ($trustedProxies = $context['TRUSTED_PROXIES'] ?? false) {
+        Request::setTrustedProxies(explode(',', $trustedProxies), Request::HEADER_X_FORWARDED_ALL ^ Request::HEADER_X_FORWARDED_HOST);
+    }
+
+    if ($trustedHosts = $context['TRUSTED_HOSTS'] ?? false) {
+        Request::setTrustedHosts([$trustedHosts]);
+    }
+
+    return new Kernel($context['APP_ENV'], (bool) $context['APP_DEBUG']);
```
weaverryan, line 17: I’m still learning about the runtime component. Is it the return value of the closure that dictates which runtime is used?

Maybe a link to some runtime Part of the docs? The new runtime stuff is cool, but untraceable (and still a mostly unfamiliar concept in our world). If I wanted to change the behavior, I’d like some way to find how to do that.

Related questions:

* where is the Debug::enable() - is that in the runtime and is always enabled for a certain environment? How could I control that?

What if I need to run some code *right* after each Request object is created, but before it’s handled? We have some complex, hacky code that does that related to trusted proxies.

Thanks!

symfony/console/5.2/bin/console
nicolas-grekas: They would create the Kernel inline as usual, right inside the close.
We could to the same here also, but I propose to reuse the definition in the index.php file here.

symfony/framework-bundle/5.2/public/index.php
nicolas-grekas: `APP_VENDOR_DIR` is used to locate the autoload.php file.
It's usually not needed because the runtime can figure by doing `../../` from inside the vendor dir - but vendors are not always there. (e.g. when using the `link` util provided in the `symfony/symfony` repo.

symfony/framework-bundle/5.2/public/index.php
nicolas-grekas: There is always only one runtime; and it knows how to deal with a range of possible return values. You can check the [`SymfonyRuntime::start()`](https://github.com/symfony/symfony/pull/36652/files#diff-18578a7bf9e590e1557d876ecf806063R100) method to know which ones are supported right now.

If you want to change the behavior, either it's related to some global state and you might want to write your own runtime or it's something that should fit right into the close. The runtime should contain as little logic as possible. Its role is to provide a consistent runtime environement by configuring the appropriate global state.

> where is the Debug::enable

It's part of this consistent runtime environment configured by the SymfonyRuntime. You can't control it unless you configure/decorate/write the runtime. If there is something that should be commonly configured, the `$_SERVER['APP_RUNTIME_OPTIONS']` is here to help (the only key supported is `dotenv` right now, see [this discussion](https://github.com/symfony/symfony/pull/36652#discussion_r419387901) where this has been asked for, see [this one](https://github.com/symfony/symfony/pull/36652#discussion_r420604922) about the error handler.)

> What if I need to run some code right after each Request object is created, but before it’s handled? We have some complex, hacky code that does that related to trusted proxies.

I think the best way would be to decorate the `http_kernel` service. Or you override the `handle()` method in your `App\Kernel`. Or you decorate your `App\Kernel` with an anonymous class in the returned closure. etc. :)

symfony/console/5.2/bin/console
nicolas-grekas: I removed created the kernel from a require, it's now declared inline.

nicolas-grekas:
PR updated to account for https://github.com/symfony/symfony/pull/37351 and https://github.com/symfony/symfony/pull/37357

symfony/console/5.2/bin/console
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env php
+<?php
+
+use App\Kernel;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+
+if (!is_file(dirname(__DIR__).'/vendor/symfony/runtime/autoload.php')) {
+    throw new LogicException('Symfony Runtime is missing. Try running "composer require symfony/runtime".');
```
yceruto, line 8: Should we do the same in `index.php`?

nicolas-grekas: I think we can skip this in index.php: it will save some perf, and since the console is run when composer install/update runs also, the message will be displayed.

symfony/framework-bundle/5.2/manifest.json
```diff
@@ -0,0 +1,27 @@
...
+        "public/": "%PUBLIC_DIR%/",
+        "src/": "%SRC_DIR%/"
+    },
+    "composer-scripts": {
+        "cache:clear": "symfony-cmd",
+        "assets:install %PUBLIC_DIR%": "symfony-cmd"
+    },
+    "env": {
+        "APP_ENV": "dev",
+        "APP_SECRET": "%generate(secret)%"
```
nicolas-grekas, line 16: The difference with the 5.1 recipe is that we don't need to list `TRUSTED_PROXIES` nor `TRUSTED_HOSTS` anymore.

symfony/framework-bundle/5.2/public/index.php
nicolas-grekas: I removed using `APP_VENDOR_DIR` since that's useful only to ppl that use the `./link` script, which is far from the majority.

javiereguiluz:
I haven't followed the related discussions about Runtime component, but let me add this pseudo-random comment here. Looking at this code:

```php
return function (array $context) {
    $kernel = new Kernel($context['APP_ENV'], (bool) $context['APP_DEBUG']);

    return new Application($kernel);
};
```

I wonder if it would be possible and/or desirable to use a proper object instead of a PHP array:

```php
return function (Runtime $runtime) {
    $kernel = new Kernel($runtime->env(), $runtime->debug());

    return new Application($kernel);
};
```

Method names would be up to be decided:

```php
$kernel = new Kernel($runtime->env(), $runtime->debug());
$kernel = new Kernel($runtime->environment(), $runtime->debugMode());
$kernel = new Kernel($runtime->env(), $runtime->isDebug());
// ...
```

nicolas-grekas:
@javiereguiluz I'm not sure creating a DTO for accessing two scalars is worth it. And I'm not sure what else this DTO could do.

Nyholm (commented):
Im happy with this PR.

symfony/console/5.3/bin/console
```diff
@@ -0,0 +1,17 @@
...
+
+use App\Kernel;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+
+if (!is_file(dirname(__DIR__).'/vendor/autoload_runtime.php')) {
+    throw new LogicException('Symfony Runtime is missing. Try running "composer require symfony/runtime".');
+}
+
+require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
+
```
noniagriconomie, line 12: Related to https://github.com/symfony/recipes/pull/787#pullrequestreview-430223298,
what about adding a comment `@see SymfonyRuntime from the symfony/runtime component`?

In `bin/console` and `public/index.php` file

---

Also related to the docbloc from the component readme:

```
The SymfonyRuntime can resolve and handle many types related to the
symfony/http-foundation and symfony/console components.
Check its source code for more information.
```

nicolas-grekas: I don't like adding comments in such files. There are other places to maintain such docs.

chalasr:
Although this may seem to add some extra mystification/complexity around the boostraping logic (which is currently super simple for the minimal/basic use case), it's really cool to see how it can reduce the learning curve when it comes to bootstrapping "special" applications like e.g. lambda or roadrunner (thanks @php-runtime), which become more and more common nowadays.

I'm 👍 here.

Nyholm (approved):
Thank you Nicolas and all the reviewers. 

@javiereguiluz That is still possible, but we need to make changed in the component if we want to add that DTO. I dont think that is a blocker for merging this.

PR #788, opened 2020-06-16
check for PHP files to exist before requiring them

xabbuh:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  |

### DIFF
--- a/symfony/framework-bundle/5.1/src/Kernel.php
+++ b/symfony/framework-bundle/5.1/src/Kernel.php
@@ -16,11 +16,10 @@ class Kernel extends BaseKernel
         $container->import('../config/{packages}/*.yaml');
         $container->import('../config/{packages}/'.$this->environment.'/*.yaml');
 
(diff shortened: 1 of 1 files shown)


symfony/framework-bundle/5.1/src/Kernel.php
```diff
@@ -21,7 +21,10 @@ protected function configureContainer(ContainerConfigurator $container): void
             $container->import('../config/{services}_'.$this->environment.'.yaml');
         } else {
             $path = \dirname(__DIR__).'/config/services.php';
-            (require $path)($container->withPath($path), $this);
+
+            if (is_file($path)) {
```
nicolas-grekas, line 25: can youmerge this if with the previous else please?

nicolas-grekas: also we should be consistent with using is_file / file_exists (I'm fine using is_file above also)

Issue #398, opened 2018-04-13
framework-bundle Do not install assets with "--symlink --relative" by default


gaiaz-iusipov:
For some reason, there are "--symlink --relative" by default.

- Docker with Nginx has mounted only `public` directory, but not `vendor`, so vendor bundles with assets become unavailable for nginx;
- Windows symlink issues.

`recipes/symfony/framework-bundle/3.3/manifest.json`
```
    "composer-scripts": {
        "cache:clear": "symfony-cmd",
        "assets:install --symlink --relative %PUBLIC_DIR%": "symfony-cmd"
    },
```

covex-nn:
I agree with @gaiaz-iusipov, and without modification in `composer.json` this could be fixed only by sharing `vendor` directory from `php` container to `nginx` container, but it is not right! Nginx container shoud only has access to assets.

I work under Windows, and symlinks are not created as relative. They are absolute, and point to`C:\ ...`. So, even if i would share `vendor` dir, Nginx will not have access to these files: inside containers symlinks still point to somewhere `C:\ ...`. And only when i run `composer auto-scripts` inside container, symlinks start to point to `../../vendor/ ...`

PR #793, opened 2020-07-08
Add symlink for installed assets

alexander-schranz:
| Q             | A
| ------------- | ---
| License       | MIT

I could not find out why this was changed so I thought it was just forgotten in the recipes that it most cases it better to install the assets with symlink as creating hard copies, like it was done in the older symfony [skeleton](https://github.com/symfony/symfony-standard/blob/cd07a8df72d611ac32aacde72b3cce2ac2f49783/composer.json#L61).

### DIFF
--- a/symfony/framework-bundle/5.4/manifest.json
+++ b/symfony/framework-bundle/5.4/manifest.json
@@ -9,7 +9,7 @@
     },
     "composer-scripts": {
         "cache:clear": "symfony-cmd",
-        "assets:install %PUBLIC_DIR%": "symfony-cmd"
+        "assets:install %PUBLIC_DIR% --symlink --relative": "symfony-cmd"
     },
     "env": {
         "APP_ENV": "dev",


nicolas-grekas:

alexander-schranz:
@nicolas-grekas rebased!

nicolas-grekas:
Can you please rebase and update only the recipe for 5.4?

alexander-schranz:
I found some related issue why maybe the symlink was removed: https://github.com/symfony/recipes/issues/398 but don't understand when this is the case and if its a good default.

nicolas-grekas:
the case is when public/ is mounted to a docker container: the symlinks will point to nowhere if vendor/ is not also mounted (and mounting it doesn't make much sense I guess.)
So maybe we should close this PR.

alexander-schranz:
@nicolas-grekas Is fine for me then to close it.

Issue #784, opened 2020-06-12
Symfony Framework Kernel should use $this->getProjectDir() instead of ../ for config/


7ochem:
The Symfony Framework Kernel should use `$this->getProjectDir()` to locate the project root and access config files from there instead of `../config/` which is being used now since Symfony 5.1.

In 4.4 (4.2) Kernel:
https://github.com/symfony/recipes/blob/81543c64382d0a8898479182e18b460ef196ab15/symfony/framework-bundle/4.2/src/Kernel.php#L38-L43

In 5.1 Kernel:
https://github.com/symfony/recipes/blob/81543c64382d0a8898479182e18b460ef196ab15/symfony/framework-bundle/5.1/src/Kernel.php#L16-L19

In all other places (`MicroKernelTrait` and `\Symfony\Component\HttpKernel\Kernel`) `$this->getProjectDir()` is being used consistently.

nicolas-grekas:
There is no need to use getProjectDir, all is fine with `..` now, the path is resolved relative to the Kernel's file.

7ochem:
I have one project where we have placed the Kernel.php file into `src/Framework/`. Before 5.1 this worked just fine. Now I have updated to 5.1, and ran `symfony:recipes:install`, moved `src/Kernel.php` to `src/Framework/Kernel.php` and ran into this issue.

Like described in the text above, before 5.1 it used `getProjectDir()` and in other places, this is also used. Why was this changed in the `Kernel.php` file overnight?

nicolas-grekas:
Will be closed by #794

PR #794, opened 2020-07-10
Remove usages of relative paths from Kernel

Jean85:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

This PR changes the Kernel for two reasons:
 * consistency
 * using relative paths breaks when extending the Kernel

I was doing the latter for testing purposes (I want to import some configurations only in certain tests, not in all the `test` environment) and it breaks under this situation.

If this gets merged, maybe we could consider backporting it to older recipes too?

### DIFF
--- a/symfony/framework-bundle/5.4/src/Kernel.php
+++ b/symfony/framework-bundle/5.4/src/Kernel.php
@@ -13,26 +13,30 @@ class Kernel extends BaseKernel
 
     protected function configureContainer(ContainerConfigurator $container): void
     {
-        $container->import('../config/{packages}/*.yaml');
-        $container->import('../config/{packages}/'.$this->environment.'/*.yaml');
+        $projectDir = $this->getProjectDir();
+
+        $container->import($projectDir.'/config/{packages}/*.yaml');
+        $container->import($projectDir.'/config/{packages}/'.$this->environment.'/*.yaml');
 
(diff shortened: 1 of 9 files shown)


nicolas-grekas:
Very related to https://github.com/symfony/recipes/issues/784

nicolas-grekas:
Please rebase and use getProjectDir instead maybe?

nicolas-grekas:
Oh, and target 5.4 now.

Jean85:
Done @nicolas-grekas. Q&A is failing due to no 5.4 stable release, should I change anything else?

symfony/framework-bundle/5.4/src/Kernel.php
```diff
@@ -0,0 +1,42 @@
...
+    use MicroKernelTrait;
+
+    protected function configureContainer(ContainerConfigurator $container): void
+    {
+        $projectDir = $this->getProjectDir();
+
+        $container->import($projectDir.'/config/{packages}/*.yaml');
+        $container->import($projectDir.'/config/{packages}/'.$this->environment.'/*.yaml');
+
+        if (is_file(\dirname(__DIR__).'/config/services.yaml')) {
```
nicolas-grekas, line 21: let's change this line also (same for routes)

nicolas-grekas (approved):

PR #795, opened 2020-07-13
Introduce APP_VAULT to select the set of secrets to use for a specific deployment

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

The Symfony env is not enough to precisely configure the set of secrets to use for a specific deployment.
E.g. you might want to deploy a staging target in "prod" mode, but with a test-mode set of API credentials.

The proposed configuration allows doing so via a new `APP_VAULT` env var. It defaults to `APP_ENV` so that nothing changes without explicit opt-in.

### DIFF
--- a/symfony/framework-bundle/5.1/config/packages/framework.yaml
+++ b/symfony/framework-bundle/5.1/config/packages/framework.yaml
@@ -1,0 +1,21 @@
+# see https://symfony.com/doc/current/reference/configuration/framework.html
+framework:
+    secret: '%env(APP_SECRET)%'
+    #csrf_protection: true
+    #http_method_override: true
+
+    secrets:
+        # set the APP_VAULT env var to override the directory where secrets are stored
+        vault_directory: '%kernel.project_dir%/config/secrets/%env(default:kernel.environment:APP_VAULT)%'
+
+    # Enables session support. Note that the session will ONLY be started if you read or write from it.
+    # Remove or comment this section to explicitly disable session support.
+    session:
+        handler_id: null
+        cookie_secure: auto
+        cookie_samesite: lax
+
+    #esi: true
+    #fragments: true
+    php_errors:
+        log: true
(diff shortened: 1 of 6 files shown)


symfony/framework-bundle/4.4/config/packages/framework.yaml
```diff
@@ -0,0 +1,20 @@
+# see https://symfony.com/doc/current/reference/configuration/framework.html
+framework:
+    secret: '%env(APP_SECRET)%'
+    #csrf_protection: true
+    #http_method_override: true
+
+    secrets:
+        vault_directory: '%kernel.project_dir%/config/secrets/%env(APP_VAULT)%'
```
ogizanagi, line 8: What about making this the default in the [fwb config](https://github.com/symfony/symfony/blob/dfc32678c186f93f1c3ead8b02231aa3732d869b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L147 ) (adding the `env(APP_VAULT): '%env(APP_ENV)%'` parameter as default in the fwb as well)?

nicolas-grekas: I'm not sure we should reference any env vars in Configuration files - we don't have any right now at least. We could add `kernel.vault` and make it default to APP_VAULT in `MicroKernelTrait` (which is the only place where we have some hardcoded env var reference.

For 4.4, the current is the only option thought.

symfony/framework-bundle/5.1/config/packages/framework.yaml
```diff
@@ -0,0 +1,20 @@
+# see https://symfony.com/doc/current/reference/configuration/framework.html
+framework:
+    secret: '%env(APP_SECRET)%'
+    #csrf_protection: true
+    #http_method_override: true
+
+    secrets:
+        vault_directory: '%kernel.project_dir%/config/secrets/%env(default:kernel.environment:APP_VAULT)%'
```
weaverryan, line 8: Comment above:

```
# set the APP_VAULT env var to override the vault directory
```

nicolas-grekas:
Closing in favor of https://github.com/symfony/symfony/pull/37584

PR #796, opened 2020-07-14
Put migrations back in src/Migrations/

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Changing this directory made the recipes incompatible with the-fast-track book.
Since the book is an important asset for the community, we'd better not break it.

Changes https://github.com/symfony/recipes/pull/729
Invalidates https://github.com/symfony/recipes/issues/718


nicolas-grekas:
Note that I don't know what the exact impact of "fixing" the book would be. Printed ones are not fixable for sure, but maybe that's not critical enough to require this PR tobe merged?
on @fabpot knows :)

fabpot:
Using '/migrations' for all versions of Doctrine seems like a great idea (which we've already talked about in the past).

PR #797, opened 2020-07-14
Move src/Migrations/ to migrations/

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Is there anything that prevents doing so, if the book can be fixed?
Replaces #796


GPHemsley:
This creates a new empty directory, leaving existing migrations in the old directory instead of moving them. Is that intentional? (Do recipes have the power to move files outside the recipe?)

nicolas-grekas:
You're correct, recipes don't move files.

PR #799, opened 2020-07-15
Update services.yaml

abdounikarim:
Hello,

I just remove the `Migrations` directory from the exclude. 
Since 5.1, this directory has been moved outside of `src`.
I just talked with @javiereguiluz and he agrees.

I can also update the docs if you want.

Cheers 😉


nicolas-grekas:
This can target the v4.2 of the recipe thanks to https://github.com/symfony/recipes/pull/797

PR #802, opened 2020-07-16
[FrameworkBundle] Use an array for exclude directive

lyrixx:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  |

IMHO, this notation is easier to understand. More over it scales really
better. Finally this notation is not documented on the [following
page](https://symfony.com/doc/current/service_container.html)

### DIFF
--- a/symfony/framework-bundle/4.2/config/services.yaml
+++ b/symfony/framework-bundle/4.2/config/services.yaml
@@ -14,7 +14,7 @@ services:
     # makes classes in src/ available to be used as services
     # this creates a service per class whose id is the fully-qualified class name
     App\:
-        resource: '../src/*'
+        resource: '../src/'
         exclude:
             - '../src/DependencyInjection/'
             - '../src/Entity/'
@@ -24,7 +24,7 @@ services:
     # controllers are imported separately to make sure services can be injected
     # as action arguments even if you don't extend any base controller class
     App\Controller\:
-        resource: '../src/Controller'
+        resource: '../src/Controller/'
         tags: ['controller.service_arguments']
 
     # add more service definitions when explicit configuration is needed
--- a/symfony/framework-bundle/4.2/config/services.yaml
+++ b/symfony/framework-bundle/4.2/config/services.yaml
@@ -15,7 +15,11 @@ services:
     # this creates a service per class whose id is the fully-qualified class name
     App\:
         resource: '../src/*'
-        exclude: '../src/{DependencyInjection,Entity,Tests,Kernel.php}'
+        exclude:
+            - '../src/DependencyInjection/'
+            - '../src/Entity/'
+            - '../src/Kernel.php'
+            - '../src/Tests/'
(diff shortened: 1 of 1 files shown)


damienalexandre:
Great change for DX because it simpler to add path than editing the glob :+1: Thanks!

stof:
this is not supported in 4.2 IIRC. It was added in a later version.

ogizanagi:
@stof I think it is: https://github.com/symfony/symfony/pull/27075 is part of 4.2

stof:
Oh indeed. `exclude` is older and 4.2 is the one expanding it.

symfony/framework-bundle/4.2/config/services.yaml
```diff
@@ -15,7 +15,11 @@ services:
     # this creates a service per class whose id is the fully-qualified class name
     App\:
         resource: '../src/*'
-        exclude: '../src/{DependencyInjection,Entity,Tests,Kernel.php}'
+        exclude:
+            - '../src/DependencyInjection/*'
```
javiereguiluz, line 19: Do we need the asterisk here? Would `../src/DependencyInjection/` work?

If I'm right, in `glob`, a single `*` means contents in that dir only and double `**` means "recursive" and goes into all subdirectories.

So `../src/DependencyInjection/**` could be more precise, but `../src/DependencyInjection/` looks easier to understand to me. But maybe I'm confusing things.

lyrixx: I don't know :man_shrugging:  /cc @nicolas-grekas

lyrixx: Indeed, the `*` is not mandatory. I just checked.

Should I remove it?

javiereguiluz: In the `.gitignore` file, other Symfony recipes add entries like this:

```
/public/bundles/
/var/
/vendor/
```

Instead of this:

```
/public/bundles/*
/var/*
/vendor/*
```

So, if it's technically correct, removing the asterisk would be consistent with other Flex features. Thanks.

lyrixx: @javiereguiluz Done, But now we have an inconsistency with the `resource`:

```
    App\:
        resource: '../src/*'
        exclude:
            - '../src/DependencyInjection/'
            - '../src/Entity/'
            - '../src/Kernel.php'
            - '../src/Tests/'

    # controllers are imported separately to make sure services can be injected
    # as action arguments even if you don't extend any base controller class
    App\Controller\:
        resource: '../src/Controller'
```

So I fixed that too :)

javiereguiluz:
👍 I like this. Thanks Grég. I don't approve it because then it will have 2 official approves and it will be merged automatically. I'd prefer if Nico and/or Fabien approve this change. Thanks!

PR #815, opened 2020-08-31
[DoctrineFixturesBundle] Update ObjectManager old namespace

larzuk91:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

I just noticed that the recipe is pointing to the old namespace of the `ObjectManager` class


fabpot (approved):
merge now

fabpot:
I read it the wrong way. Current namespace is the good one. Reverting.

PR #818, opened 2020-09-10
Exclude src/preload.php from service discovery

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

As spotted in https://github.com/symfony/symfony/issues/38123


chalasr (approved):

fabpot (changes requested):
Probably not a good solution (as people upgrading to latest 4.4/5.1 won't get this change).

PR #820, opened 2020-09-10
Resolve deprecation message for symfony/dotenv 5.1

andrea-cristaudo:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

Dotenv component doesn't allow anymore constructor to be called with a boolean to set on/off the use of putenv.
This PR avoid the deprecation to be triggered as it is done in this commit from symfony flex (https://github.com/symfony/flex/commit/1bf83d4678ed937a77438ae681aa97d6ad4c1094#diff-b0bd3a5ec4048cca5e142d072c4a290b).


xabbuh:
I don't think that's necessary. The recipe for 5.1 already doesn't pass a boolean anymore.

nicolas-grekas:
@xabbuh is right, closing as such. Of course, please explain what we missed, if we did miss anything in your opinion.

PR #825, opened 2020-09-29
[symfony/framework-bundle] Add config/preload.php

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Follows https://github.com/symfony/symfony/issues/38334

### DIFF
--- a/symfony/framework-bundle/4.2/manifest.json
+++ b/symfony/framework-bundle/4.2/manifest.json
@@ -23,7 +23,6 @@
         "/.env.*.local",
         "/%CONFIG_DIR%/secrets/prod/prod.decrypt.private.php",
         "/%PUBLIC_DIR%/bundles/",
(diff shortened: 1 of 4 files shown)


wouterj (approved):
Nice solution, I like it this way!

chalasr (approved):

fabpot (approved):

GPHemsley:
It appears that the auto-generated {{src/.preload.php}} used the current environment (e.g. dev) to create the path of the files, while this recipe hardcodes prod. (It also seems to add a second path.) Is that intentional? Is this something app developers should be configuring?

nicolas-grekas:
> Is this something app developers should be configuring?

yes, absolutely: like everything in `config/`, this is yours to tune.

PR #826, opened 2020-10-01
[FrameworkBundle] Simplify the service and route registration

lyrixx:
Bracket are not needed since we can be sure the file exist

| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  |


nicolas-grekas:
IIRC, the brackets allow the Kernel to not fail when the file is missing.

lyrixx:
@nicolas-grekas yes, indeed. But the file **do** exist (look at the line just above). This is why removing brackets are OK.
But I kept the brackets on others line

nicolas-grekas (approved):

yceruto (approved):

PR #831, opened 2020-10-07
Restore deprecation filter for Symfony versions prior 5.1

l-vo:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  |    

Related to https://github.com/symfony/recipes/pull/775#issuecomment-697953963

Deprecations handler with filter should not be removed since it's still needed for Kernel/Fwb versions < 5.1.


nicolas-grekas:
I'd prefer not doing this: it oriented towards legacy deps. Better optimize for newest versions.

l-vo:
Yes I understand the will, but for folks that migrate from 3.4 to 4.4 (for instance), the recipe update suggests to remove the deprecation handler. It involves a minimum understanding of what is done in symfony/symfony#36621 to not remove the handlers. Are we ok with that ?

PR #839, opened 2020-11-04
[Messenger] Let failed transport DNS depend on environment variable

jeroennoten:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | https://github.com/symfony/symfony-docs/pull/14519

To me, it didn't make sense to have the async transport be based on an environment variable, while having the failed transport hardcoded. So I thought, having both depend on the same environment variable, with the queue name added for the failed transport made more sense.

### DIFF
--- a/symfony/messenger/4.3/config/packages/messenger.yaml
+++ b/symfony/messenger/4.3/config/packages/messenger.yaml
@@ -6,7 +6,7 @@ framework:
         transports:
             # https://symfony.com/doc/current/messenger.html#transport-configuration
             # async: '%env(MESSENGER_TRANSPORT_DSN)%'
-            # failed: 'doctrine://default?queue_name=failed'
+            # failed: '%env(MESSENGER_TRANSPORT_DSN)%?queue_name=failed'
             # sync: 'sync://'
 
         routing:


symfony/messenger/4.3/config/packages/messenger.yaml
```diff
@@ -6,7 +6,7 @@ framework:
         transports:
             # https://symfony.com/doc/current/messenger.html#transport-configuration
             # async: '%env(MESSENGER_TRANSPORT_DSN)%'
-            # failed: 'doctrine://default?queue_name=failed'
+            # failed: '%env(MESSENGER_TRANSPORT_DSN)%?queue_name=failed'
```
jderusse, line 9: Using the same env variable for both the "working queue" and the "failed queue" is confusing.
I' rather use something like "MESSENGER_TRANSPORT_FAILED_DSN"

jeroennoten: In my opinion, using the same environment variable for both queues is actually a good default, because most of the times, only the queue name is different. I think that it makes sense to configure the queue name in the configuration file, and the transport 'technology' with an environment variable.

jderusse: nvm, I missed the `?queue_name=failed` in the parameter

fabpot (changes requested):
We did it that way as the failed queue is mostly useful when using the DB. Using the same transport is not likely, so reusing the same DSN is probably the exception instead of being the rule.

:-1: for me

jeroennoten:
@fabpot makes sense. I'll close this PR and the doc PR.

Issue #805, opened 2020-07-29
Allow PHP config in packages/ and routes/ folders


ghost:
[The current recipe for framework-bundle](https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/5.1/src/Kernel.php) does'nt take into account the .php configuration files in the `config/packages/` folder.

In the same way, it's configured to only receive routes from a single `routes.php` file in the `config/` folder, but it should be possible to have multiple .php files to configure routes inside the `config/routes/` folder.

Since the use of php file-based configuration will be promoted in future versions of Symfony, this feature should be possible from now on.

TavoNiievez:
Working config...
```diff
class Kernel extends BaseKernel
{
    //...

    protected function configureContainer(ContainerConfigurator $container): void
    {
+        $container->import('../config/{packages}/*.php');
         $container->import('../config/{packages}/*.yaml');
+        $container->import('../config/{packages}/'.$this->environment.'/*.php');
         $container->import('../config/{packages}/'.$this->environment.'/*.yaml');

		//...
    }

    protected function configureRoutes(RoutingConfigurator $routes): void
    {
+        $routes->import('../config/{routes}/'.$this->environment.'/*.php');
         $routes->import('../config/{routes}/'.$this->environment.'/*.yaml');
+        $routes->import('../config/{routes}/*.php');
         $routes->import('../config/{routes}/*.yaml');

		//...
    }
}
```

Tobion:
See https://github.com/symfony/recipes/pull/840#issuecomment-722363810 and https://github.com/symfony/recipes/pull/840#issuecomment-735235216

PR #840, opened 2020-11-05
Allow PHP config in /packages and /routes folders

TavoNiievez:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | #805

The kernel has logic to load yaml and php config files, but ignores the `.php` config in the `packages/*` and `routes/*` folders.

### DIFF
--- a/symfony/framework-bundle/5.1/src/Kernel.php
+++ b/symfony/framework-bundle/5.1/src/Kernel.php
@@ -13,7 +13,9 @@ class Kernel extends BaseKernel
 
     protected function configureContainer(ContainerConfigurator $container): void
     {
+        $container->import('../config/{packages}/*.php');
         $container->import('../config/{packages}/*.yaml');
+        $container->import('../config/{packages}/'.$this->environment.'/*.php');
         $container->import('../config/{packages}/'.$this->environment.'/*.yaml');
 
         if (is_file(\dirname(__DIR__).'/config/services.yaml')) {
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
We don't recommend using php config inside packages/ Also, no recipes use the php format for this folder. The reason is that php config doesn't work well with php.

If you want to use php, this is PR shows how you can patch your kernel to to so.

But I'm not sure we should make this the default for eveybody for now at least.

TavoNiievez:
@nicolas-grekas , thanks for your quick response
It's just that i find it strange that the Symfony documentation offers you php as a configuration option, but if you try to add a .php file in those folders it just doesn't load.
In fact, the order in which the files are loaded appears to be important. The `.php` must be loaded first and then the `.yaml`, if i only have php configuration files and i reverse the order of these two lines the application breaks.
```php
		// this works
        $container->import('../config/{packages}/*.php');

		// but... this doesn't
        $container->import('../config/{packages}/*.php');

		// and replacing it doesn't work either
        $container->import('../config/{packages}/*.php');
```
i'm not sure this is at least documented. Similarly, Kernel seems to ignore the `services_test.php` or `services_dev.php` files.
https://github.com/symfony/recipes/blob/882d50ba26960d918e1d98aa04a6d885f67f5c28/symfony/framework-bundle/5.1/src/Kernel.php#L21
and i'm not sure how to add it as apparently loading yaml by environment seems to be different than loading php:
https://github.com/symfony/recipes/blob/882d50ba26960d918e1d98aa04a6d885f67f5c28/symfony/framework-bundle/5.1/src/Kernel.php#L22-L23

nicolas-grekas:
I'm closing as explained. Feel free to do these changes in your project of course.

PR #844, opened 2020-11-17
Always log application specific logs by default

ostrolucky:
| Q             | A
| ------------- | ---
| License       | MIT

Current default ignores ALL logs where there was no error in a HTTP request. This is not good, because it means any application specific logging, eg. the ones tied to business logic, is not recorded.

I would like to change that and hence have more sane logging by default. I don't think people expect that if they do `$this->logger->log(...)` in their service, it won't be logged in production. I would think that only framework/library diagnostic logs should not logged by default.

This PR solves that. Application using this proposed config will have all logs recorded that was created via `$this->logger->log(...)`, but old logging behaviour for rest of the logs is preserved.

### DIFF
--- a/symfony/monolog-bundle/3.1/config/packages/prod/monolog.yaml
+++ b/symfony/monolog-bundle/3.1/config/packages/prod/monolog.yaml
@@ -8,6 +8,7 @@ monolog:
             excluded_404s:
                 # regex: exclude all 404 errors from the logs
                 - ^/
+            channels: ["!app"]
         nested:
             type: stream
             path: "%kernel.logs_dir%/%kernel.environment%.log"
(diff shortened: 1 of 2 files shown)


ostrolucky:
Alternative would be to apply this handler only for >= info levels, but we would then have to remove exclusion of `app` channel from fingers_crossed

Tobion:
I don't agree with this. The default should be simple. Now it gets more complicated.
E.g. for people using explicit channels, this won't apply again. But those are likely also business logs.

jderusse:
> Current default ignores ALL logs where there was no error in a HTTP request

? what do you mean? in a standard application all my log are recorded,.

ostrolucky:
>  for people using explicit channels, this won't apply again.

This config is a default, for initial use of monolog. In beginning there obviously aren't more channels.

> ? what do you mean? in a standard application all my log are recorded,.

Current default for `prod` logs to file only on error.

nicolas-grekas:

ostrolucky:
No point maintaining this if I'm not getting any feedback

PR #846, opened 2020-11-19
remove unsued mysql configuration

abdounikarim:
| Q             | A
| ------------- | ---
| License       | MIT

Hello,

I just remove old mysql configuration.


nicolas-grekas:
I think we should keep these lines because they're critical when one wants to use MySQL.

abdounikarim:
@nicolas-grekas we can just comment them ?

fabpot:
When not using MySQL, is there any issues with these lines?

nicolas-grekas:
They're ignored when not using mysql.

PR #823, opened 2020-09-25
Added API Platform 2.6 recipe

mtarld:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | 

Added API Platform 2.6 recipe in order to prevent deprecations that will be triggered by https://github.com/api-platform/core/pull/3436.

Recipe is the same as the 2.5 one but with `defaults.stateless` set to `true`

PR #648, opened 2019-09-17
[api-platform] Update recipe for 2.5

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

Enables `PATCH` support and switch to OpenAPI v3. Will be the defaults in API Platform 3.

PR #853, opened 2020-12-02
Revert "[api-platform] Update recipe for 2.5"

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

This reverts commit 1ed3ff9f848948903bcc58a9b3571c24431af635. The `stateless` option cannot be enabled by default because the firewall with the default configuration uses the session, so it conflicts.


chalasr (approved):

fabpot (approved):

guilliamxavier:
I think you reverted the wrong commit! This reverted #648 while I guess the intent was to revert #823

dunglas:
@guilliamxavier 🤦‍♂️ you're right! I was too tired I guess...

PR #854, opened 2020-12-03
Add Stimulus in Webpack Encore recipe

tgalopin:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Add Stimulus in Webpack Encore recipe for Symfony UX support

### DIFF
--- a/symfony/webpack-encore-bundle/1.0/assets/controllers/hello_controller.js
+++ b/symfony/webpack-encore-bundle/1.0/assets/controllers/hello_controller.js
@@ -1,0 +1,16 @@
+import { Controller } from 'stimulus';
+
+/*
+ * This is an example Stimulus controller!
+ *
+ * Any element with a data-controller="hello" attribute will cause
+ * this controller to be executed. The name "hello" comes from the filename:
+ * hello_controller.js -> "hello"
+ *
+ * Delete this file or adapt it for your use!
+ */
+export default class extends Controller {
+    connect() {
+        this.element.textContent = 'Hello Stimulus! Edit me in assets/controllers/hello_controller.js';
+    }
+}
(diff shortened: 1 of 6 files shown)


symfony/webpack-encore-bundle/1.0/package.json
```diff
@@ -1,7 +1,9 @@
 {
     "devDependencies": {
+        "@symfony/stimulus-bridge": "^1.0.0",
         "@symfony/webpack-encore": "^0.31.0",
```
jmsche, line 4: Requires ^0.32, right?

symfony/webpack-encore-bundle/1.0/assets/controllers/hello_controller.js
```diff
@@ -0,0 +1,16 @@
+import { Controller } from 'stimulus';
+
+/*
+ * This is an example Stimulus controller!
+ *
+ * Any element with with a data-controller="hello" attribute will cause
```
nicolas-grekas, line 6: double "with"

fabpot (approved):

fabpot (commented):
merge now

alexander-schranz:
@tgalopin @weaverryan any possibility to make this into an own recipe? We in sulu have our own implementation how we recommend implementing [js components](https://github.com/sulu/web-js-twig-demo). And we still want to use webpack encore but it make it harder for our users that they need now remove always the stimulus integration.

tgalopin:
Perhaps what you're looking for is a dedicated entrypoint for your JS components? Let's chart about it in Symfony Devs Slack?

alexander-schranz:
Thx for your time 🙏  again, for the public as discussed we probably will use only `webpack encore js library` and not the bundle at current state as stimilus is to integrated at current state when using the bundle and I understand your point why it is this way.

PR #868, opened 2021-01-07
Bump to phpunit 8.5

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

PHPUnit 8.5 is an LTS version with PHP 7.2 as the minimum supported version. Perfect for 5.*


wouterj (approved):
:+1: PHPUnit 7.5 also doesn't support PHP 8, so this is a much safer default

fabpot (approved):

PR #872, opened 2021-01-15
Define services and routes in App\Kernel

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Going one step further than #765, we can just drop the `config/services.yaml` and `config/routes.yaml` files entirely.


BoShurik:
With this change part of the container and router configuration will be in `src` but semantically should be in `config`

fabpot:
Not sure I like this change. It makes configuration less visible and buried into a PHP class.

fabpot (changes requested):
Having thought a bit more about this one, and I definitely don't like this change. :-1: for me.

nicolas-grekas:
Sorry, I posted this from 2025, I forgot we were still in 2021 ;)

PR #874, opened 2021-01-15
Moving scripts to <head> by default

weaverryan:
Hi!

This is "kind of" a big change from a documentation POV, though the docs PR to update things was very small (and I'll check SymfonyCasts for any notes needed).

Having `script` tags in the `head` is more performant (and works well with Turbo) AND has the same behavior as putting the scripts at the bottom as long as you use `<script defer`. In WebpackEncore 1.9, you can now specify to add `defer` automatically on all script tags and the new 1.9 recipe enables this by default

So, things should "just work" like before, but some docs will need to be updated.

| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | symfony/symfony-docs#14836 

TODO

* [x] Give this a quick test
* [X] Make a docs PR


dunglas (approved):

nicolas-grekas (approved):

fabpot:
Besides the docs and videos, it also has a big impact (like any recipe change that does not target the next Symfony version) on the book.

I've mentioned that in the past, but I think we should not make such changes in past version and limit them to the next minor version.

fabpot:
I forgot to mention that it also makes some blog posts not working anymore. If one tries to follow the book examples, the patches won't apply cleanly anymore. That should not happen IMHO. We should not break the past.

PR #877, opened 2021-01-18
Cleanup Kernel & services.yaml files

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

As shown in #872 and #765, we're not ready yet for defaulting to PHP config files.

This PR reverts #785 (/cc @weaverryan), which was motivated by https://github.com/symfony/maker-bundle/pull/604, which's got closed eventually.

Let's keep things simple.

### DIFF
--- a/symfony/framework-bundle/5.3/config/services.yaml
+++ b/symfony/framework-bundle/5.3/config/services.yaml
@@ -1,0 +1,24 @@
+# This file is the entry point to configure your own services.
+# Files in the packages/ subdirectory configure your dependencies.
+
+# See https://symfony.com/doc/current/best_practices.html#configuration
+parameters:
+
+services:
+    # default configuration for services in *this* file
+    _defaults:
+        autowire: true      # Automatically injects dependencies in your services.
+        autoconfigure: true # Automatically registers your services as commands, event subscribers, etc.
+
+    # makes classes in src/ available to be used as services
+    # this creates a service per class whose id is the fully-qualified class name
+    App\:
+        resource: '../src/'
+        exclude:
+            - '../src/DependencyInjection/'
+            - '../src/Entity/'
+            - '../src/Kernel.php'
+            - '../src/Tests/'
+
+    # add more service definitions when explicit configuration is needed
+    # please note that last definitions always *replace* previous ones
--- a/symfony/framework-bundle/5.3/config/packages
+++ b/symfony/framework-bundle/5.3/config/packages
@@ -1,0 +1,1 @@
+../../4.2/config/packages
--- a/symfony/framework-bundle/5.3/config/preload.php
+++ b/symfony/framework-bundle/5.3/config/preload.php
@@ -1,0 +1,1 @@
+../../5.1/config/preload.php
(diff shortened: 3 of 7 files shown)


symfony/framework-bundle/4.4/config/services.yaml
```diff
@@ -21,11 +20,5 @@ services:
             - '../src/Kernel.php'
             - '../src/Tests/'
 
-    # controllers are imported separately to make sure services can be injected
-    # as action arguments even if you don't extend any base controller class
-    App\Controller\:
-        resource: '../src/Controller/'
-        tags: ['controller.service_arguments']
```
nicolas-grekas, line 28: we don't have to support this by default, KISS

symfony/framework-bundle/5.3/config/services.yaml
```diff
@@ -0,0 +1,24 @@
...
+    # this creates a service per class whose id is the fully-qualified class name
+    App\:
+        resource: '../src/'
+        exclude:
+            - '../src/DependencyInjection/'
+            - '../src/Entity/'
+            - '../src/Kernel.php'
+            - '../src/Tests/'
+
+    # add more service definitions when explicit configuration is needed
```
nicolas-grekas, line 23: No more autodiscovery for the `App\Controller\` namespace, and not more `src/Controller/.gitignore` file in the recipe.

nicolas-grekas (commented):
Updated for 5.3 only, in accordance with  #876

symfony/framework-bundle/5.3/config/services.yaml
xabbuh: This means controllers will only be discovered automatically if they extend the `AbstractController` class, right?

nicolas-grekas: Yes!

xabbuh: I suppose that requires updating the documentation first then.

weaverryan: It *is* an ugly spot in the config file. But yes, we would definitely need to cover this in the docs - this would be a big WTF possibly.

javiereguiluz: I don't understand this change. We used this sine Symfony 4.4:

```yaml
    # controllers are imported separately to make sure services can be injected
    # as action arguments even if you don't extend any base controller class
    App\Controller\:
        resource: '../src/Controller/'
```

It worked beautifully all these years. If there isn't a strong technical reason to change this, we should keep it. Thanks!

stof: IIRC, this was added to support controllers not using the `AbstractController` as base class

nicolas-grekas: @stof is correct. In my opinion, this is supporting an uncommon use case, which we don't have to by default. Better keep config files lean.

javiereguiluz: It's OK then if we consider that it's an uncommon use case.

But please, let's try to minimize changes because it hurts docs maintenance and it's a pain for some users 🙏  Thanks!

PR #880, opened 2021-01-20
Remove "public: false"

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Not needed anymore, 3.4 is EOLed.

### DIFF
--- a/doctrine/doctrine-bundle/1.6/config/packages/prod/doctrine.yaml
+++ b/doctrine/doctrine-bundle/1.6/config/packages/prod/doctrine.yaml
@@ -14,12 +14,10 @@ doctrine:
 services:
     doctrine.result_cache_provider:
         class: Symfony\Component\Cache\DoctrineProvider
-        public: false
         arguments:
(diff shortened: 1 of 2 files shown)


wouterj:
I don't think this change is allowed according to the new recipe guidelines:

> Creating a new project with a set of dependencies should always use the same recipes to avoid differences between the generated code and the existing documentation, blog posts, videos for these versions.
> https://github.com/symfony/recipes#updating-recipes

nicolas-grekas:
This might be a problem because it'd mean we cannot clean up anything.

/cc @fabpot please advise.

wouterj:
I think we can do it for 2.3+ only, unless this qualifies as a bug (or we reword the policy, I see benefits at both sides of this decision). Let's wait for Fabien response.

derrabus (approved):

nicolas-grekas:
Both recipes are about outdated packages, closing therefore.

PR #883, opened 2021-01-23
Updating package.json and bootstrap.js for upcoming stimulus-bridge

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | It appears nothing is needed

This is waiting on:
* [x] Waiting on the 1.2.0 tags to make it to the sub-tree splits of the UX packages - e.g. https://github.com/symfony/ux-chartjs/tags
* [x] Merge of https://github.com/symfony/stimulus-bridge/pull/19
* [x] Release of symfony/stimulus-bridge
* [x] Release of @symfony/webpack-encore

This affects all version of WebpackEncoreBundle... but it's an odd case where the bundle and Webpack encore itself are not very "bound" together. We could update this recipe in only the latest WebpackEncoreBundle version... or even bump a new minor version of the bundle and add it there. This relates to https://github.com/symfony/recipes#updating-recipes

### DIFF
--- a/symfony/webpack-encore-bundle/1.0/package.json
+++ b/symfony/webpack-encore-bundle/1.0/package.json
@@ -1,6 +1,6 @@
 {
     "devDependencies": {
-        "@symfony/stimulus-bridge": "^1.1.0",
+        "@symfony/stimulus-bridge": "^2.0.0",
         "@symfony/webpack-encore": "^1.0.0",
         "core-js": "^3.0.0",
         "regenerator-runtime": "^0.13.2",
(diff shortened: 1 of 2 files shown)


symfony/webpack-encore-bundle/1.0/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "devDependencies": {
-        "@symfony/stimulus-bridge": "^1.1.0",
-        "@symfony/webpack-encore": "^0.32.0",
+        "@symfony/stimulus-bridge": "^1.2.0",
+        "@symfony/webpack-encore": "^1.0.0",
```
OskarStark, line 4: Isn't this considering a BC break if we now require a new major version? Or does it contain no breaks?

weaverryan: Yes, the new version does contain BC breaks. But this is the awkward things with the Encore recipe: the bundle does not have or need a new version. But we will want new projects to begin using `@symfony/webpack-encore` 1.0.

fabpot (approved):

jmsche:
Hi, is it possible to merge this? Bridge 2.0 was tagged yesterday, and so was UX 1.2.0, so it would help a lot for new installs.

weaverryan:
I'm getting it merged now - the UX individual packages (despite the tag timestamp) weren't pushed until a few hours ago. I was waiting until then - but we're good now!

Nyholm (approved):
Thank you

PR #887, opened 2021-02-03
panther: display extra installation steps and enable the screenshot on failure feature

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

Enables https://github.com/symfony/panther/pull/392.


symfony/phpunit-bridge/4.1/.env.test
```diff
@@ -3,3 +3,4 @@ KERNEL_CLASS='App\Kernel'
 APP_SECRET='$ecretf0rt3st'
 SYMFONY_DEPRECATIONS_HELPER=999999
 PANTHER_APP_ENV=panther
+PANTHER_ERROR_SCREENSHOT_DIR=./tests/error-screenshots
```
weaverryan, line 6: What about `./var/error-screenshots`? It's already ignored

dunglas: Isn't it harder to discover? I was hesitating with simply `./error-screenshots`, at the root.

nicolas-grekas: I like `var/something` also. Doc will tell where this is anyway.

dunglas: @nicolas-grekas yes, I updated the PR accordingly. It looks like my comment saying that I changed my mind has been lost :/ I have issues with GitHub these days.

chalasr (approved):

nicolas-grekas (approved):

PR #888, opened 2021-02-05
Set display_errors for phpunit

BackEndTea:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | N/A

When phpunit runs into an error, with `display_errors` set to 0, you will not see any output, and phpunit will just exit. This is somewhat of a misconfiguration for a dev env, but i've ran into this a few times. By forcing display_errors to 1, you will get an error output, which may point you into the right direction.

I've opted to add this to all versions, instead of from the newest version, as the difference is minimal, and shouldn't be too confusing when tutorials etc may show an older version.

### DIFF
--- a/phpunit/phpunit/4.7/phpunit.xml.dist
+++ b/phpunit/phpunit/4.7/phpunit.xml.dist
@@ -8,6 +8,7 @@
          bootstrap="tests/bootstrap.php"
 >
     <php>
+        <ini name="display_errors" value="1" />
         <ini name="error_reporting" value="-1" />
         <server name="APP_ENV" value="test" force="true" />
         <server name="SHELL_VERBOSITY" value="-1" />
--- a/phpunit/phpunit/9.3/phpunit.xml.dist
+++ b/phpunit/phpunit/9.3/phpunit.xml.dist
@@ -9,6 +9,7 @@
          convertDeprecationsToExceptions="false"
 >
     <php>
+        <ini name="display_errors" value="1" />
         <ini name="error_reporting" value="-1" />
         <server name="APP_ENV" value="test" force="true" />
         <server name="SHELL_VERBOSITY" value="-1" />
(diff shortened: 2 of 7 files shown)


nicolas-grekas:
I've never experienced this issue even though I have `display_errors=0`. When does this make a difference? Can you provide a small app to see this?

BackEndTea:
Took me a bit to figure out what went wrong originally. The issue is that you don't get any error output when both `display_errors` is off, and `log_errors` is off. If either one is  on you should get a message.

Example with a class making a static method non static (any php script that produces an error is fine):

```php
<?php
//foo.php 
class One
{
    public static function foo()
    {

    }
}

class Two extends One
{
    public function foo()
    {

    }
}
```

```sh
echo "============================"
echo "1 display, 1 log"
php -ddisplay_errors=1 -dlog_errors=1 foo.php
echo "done"
echo "============================"
echo "0 display, 1 log"
php -ddisplay_errors=0 -dlog_errors=1 foo.php
echo "done"
echo "============================"
echo "1 dislay 0 log"
php -ddisplay_errors=1 -dlog_errors=0 foo.php
echo "done"
echo "============================"
echo "0 dislay 0 log"
php -ddisplay_errors=0 -dlog_errors=0 foo.php
echo "done"
echo "============================"
```


```
$ bash run.sh
============================
1 display, 1 log
PHP Fatal error:  Cannot make static method One::foo() non static in class Two in /var/www/foo.php on line 13

Fatal error: Cannot make static method One::foo() non static in class Two in /var/www/foo.php on line 13
done
============================
0 display, 1 log
PHP Fatal error:  Cannot make static method One::foo() non static in class Two in /var/www/foo.php on line 13
done
============================
1 dislay 0 log

Fatal error: Cannot make static method One::foo() non static in class Two in /var/www/foo.php on line 13
done
============================
0 dislay 0 log
done
============================

```

Note that the last one has no error output.

nicolas-grekas:

BackEndTea:
rebased

nicolas-grekas:
Maybe we can update only the latest ofeach major?

nicolas-grekas (approved):

derrabus (approved):

PR #889, opened 2021-02-05
symfony/phpunit-bridge: set umask(0000) when debug is true

fancyweb:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

When running functional tests with an empty cache (no directory), the cache directory is created. I run my tests with `APP_DEBUG=1`. I expect the cache directory to be created with the same permissions whether it is created from the tests or from a real request on the front controller. Currently, running the tests and then accessing the application lead to unexpected permissions errors.

Ref https://github.com/symfony/recipes/pull/210

### DIFF
--- a/symfony/phpunit-bridge/3.3/tests/bootstrap.php
+++ b/symfony/phpunit-bridge/3.3/tests/bootstrap.php
@@ -9,3 +9,7 @@ if (file_exists(dirname(__DIR__).'/config/bootstrap.php')) {
 } elseif (method_exists(Dotenv::class, 'bootEnv')) {
     (new Dotenv())->bootEnv(dirname(__DIR__).'/.env');
 }
+
+if ($_SERVER['APP_DEBUG']) {
+    umask(0000);
+}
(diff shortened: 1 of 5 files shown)


symfony/phpunit-bridge/3.3/tests/bootstrap.php
```diff
@@ -9,3 +9,7 @@
 } elseif (method_exists(Dotenv::class, 'bootEnv')) {
     (new Dotenv())->bootEnv(dirname(__DIR__).'/.env');
 }
+
+if ($_SERVER['APP_DEBUG']) {
+    umask(0000);
+}
```
nicolas-grekas, line 15: missing EOL

nicolas-grekas (approved):

nicolas-grekas:
(please rebase so that we can run the new checks)

nicolas-grekas:
This should target the recipe for 5.3 IMHO.

nicolas-grekas:
At least, since we don't use symlinks anymore, this should update the latest of each major.

nicolas-grekas:
Friendly ping @fancyweb

symfony/phpunit-bridge/4.1/tests/bootstrap.php
```diff
@@ -9,3 +9,7 @@
 } elseif (method_exists(Dotenv::class, 'bootEnv')) {
     (new Dotenv())->bootEnv(dirname(__DIR__).'/.env');
 }
+
+if ($_SERVER['APP_DEBUG']) {
+    umask(0000);
+}
```
fabpot, line 15: missing EOL (and in some other files as well)

fabpot (changes requested):

fabpot (approved):

PR #890, opened 2021-02-05
Improve onboarding when using Encore

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

I'm working on Symfony UX and I think the onboarding when using  Encore can be improved. Here is a first step.


weaverryan (approved):
The comment is a little ugly... but this works for me. I had just done this by hand today and it was, indeed, cumbersome, even with all of these functions memorized.

nicolas-grekas (approved):

PR #892, opened 2021-02-07
Disable "http_method_override" by default

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | https://github.com/symfony/symfony-docs/pull/14939

When this is turned on (the default) one can hit a route that accepts only PUT, but with a POST.
Shouldn't we disable this by default, as it's not that commonly needed, isn't it?

### DIFF
--- a/symfony/framework-bundle/5.3/config/packages/cache.yaml
+++ b/symfony/framework-bundle/5.3/config/packages/cache.yaml
@@ -1,0 +1,19 @@
+framework:
+    cache:
+        # Unique name of your app: used to compute stable namespaces for cache keys.
+        #prefix_seed: your_vendor_name/app_name
+
+        # The "app" cache stores to the filesystem by default.
+        # The data in this cache should persist between deploys.
+        # Other options include:
+
+        # Redis
+        #app: cache.adapter.redis
+        #default_redis_provider: redis://localhost
+
+        # APCu (not recommended with heavy random-write workloads as memory fragmentation can cause perf issues)
+        #app: cache.adapter.apcu
+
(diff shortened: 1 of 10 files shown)


dunglas:
I'm in favor of this change but shouldn't we target only 5.3?

xabbuh:
We need to be careful here as such a change means that applications simulating DELETE requests won't work anymore. I don't remember if EasyAdminBundle, for example, uses them by default (@javiereguiluz any idea about this?).

javiereguiluz:
@xabbuh thanks for the ping! Yes, we did that in the past, but we now use a normal form with `POST` method and a CSRF token (https://github.com/EasyCorp/EasyAdminBundle/blob/423c7f3b2dfd8fd3204bbe43a97390c4169d1d39/src/Resources/views/crud/includes/_delete_form.html.twig#L2-L4). This is the same as we do in the Symfony Demo app.

xabbuh:
Thanks for the reply Javier! What about CMS like solutions like Sulu?

nicolas-grekas:
> I'm in favor of this change but shouldn't we target only 5.3?

Updated, for 5.3 only it is now.

> applications simulating DELETE requests won't work anymore.

This won't break any apps since that's only for new apps.

> What about CMS like solutions like Sulu?

They already ship their own `framework.yaml`, so that flex won't override it.
https://github.com/sulu/skeleton/blob/2.x/config/packages/framework.yaml

fabpot (approved):

chalasr (approved):

PR #894, opened 2021-02-08
monolog: log on stderr and use JSON in prod

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | todo

Follows https://symfony.com/blog/logging-in-symfony-and-the-cloud and https://github.com/api-platform/api-platform/pull/1805.

Logging on `stderr` is considered a best practice in most containerized environments including SymfonyCloud, Docker and Kubernetes. It looks like the best default for new projects.

This PR also changes the default format used in production to JSON to improve the compatibility with modern logging systems such as ELK, EFK or DataDog (and even when using the command-line thanks to `jq`).

I targeted MonologBundle 3.7 to not impact existing installations.

### DIFF
--- a/symfony/monolog-bundle/3.7/config/packages/dev/monolog.yaml
+++ b/symfony/monolog-bundle/3.7/config/packages/dev/monolog.yaml
@@ -1,0 +1,19 @@
+monolog:
+    handlers:
+        main:
+            type: stream
+            path: php://stderr
+            level: debug
+            channels: ["!event"]
+        # uncomment to get logging in your browser
+        # you may have to allow bigger header sizes in your Web server configuration
+        #firephp:
+        #    type: firephp
+        #    level: info
+        #chromephp:
+        #    type: chromephp
+        #    level: info
+        console:
+            type: console
+            process_psr_3_messages: false
+            channels: ["!event", "!doctrine", "!console"]
--- a/symfony/monolog-bundle/3.7/config/packages/prod/deprecations.yaml
+++ b/symfony/monolog-bundle/3.7/config/packages/prod/deprecations.yaml
@@ -1,0 +1,8 @@
+# As of Symfony 5.1, deprecations are logged in the dedicated "deprecation" channel when it exists
+#monolog:
+#    channels: [deprecation]
+#    handlers:
+#        deprecation:
+#            type: stream
+#            channels: [deprecation]
+#            path: php://stderr
--- a/symfony/monolog-bundle/3.7/config/packages/prod/monolog.yaml
+++ b/symfony/monolog-bundle/3.7/config/packages/prod/monolog.yaml
@@ -1,0 +1,17 @@
+monolog:
+    handlers:
+        main:
+            type: fingers_crossed
+            action_level: error
+            handler: nested
+            excluded_http_codes: [404, 405]
+            buffer_size: 50 # How many messages should be saved? Prevent memory leaks
+        nested:
+            type: stream
+            path: php://stderr
+            level: debug
+            formatter: monolog.formatter.json
+        console:
+            type: console
+            process_psr_3_messages: false
+            channels: ["!event", "!doctrine"]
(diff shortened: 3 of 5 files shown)


chalasr (approved):
Looks good to me

BoShurik:
Is there some statistic on this? How many symfony apps use containerized environments?

dunglas:
@BoShurik I don't think so, Symfony doesn't have a telemetry system. Anyway this PR is convenient even for non-containairized apps (ELK has been an industry standard for a while, and storing structured logs is considered a good practice).

BoShurik:
I don't think `stderr` is a good default. E.g. `php-fpm` ignores it. You need to set `catch_workers_output`. And looks like there is a bug https://bugs.php.net/bug.php?id=73886

derrabus (approved):

maartendekeizer:
When running on IIS (via FastCGI) this will trigger the following error "errno=9 Bad file descriptor".
I'm not sure if I support the change.
I think we need at leaste keep a comment in the `config/packages/(dev|prod)/monolog.yaml` referencing the 'old' value `"%kernel.logs_dir%/%kernel.environment%.log"`

nicolas-grekas:
The part about the dev env should be reverted, see #938 for details.

TomaszGasior:
**This is very bad decision.**

While I understand that most app deployments nowadays happen in containerized environment or using dedicated servers with full access to system logs, this changes makes developer experience much worse for people  who use shared hostings without ability to access full system logs.

For people like me this new configuration adds extra work of changing configuration before deployment. If I forget about this little thing for my new app, all logs from production instance of my application are basically lost or, at least, very hard to access. 

Defaulting to **predictable file on the disk** is much better for developer experience, especially while debugging errors on production instance.

IMHO the best way to solve the problem tried to solved by this PR is to move logs target setting to environment variable, making it possible to redefine the path to stderr where it's applicable, defaulting to file on the disk — which is much more predictable for each server.

dunglas:
Nowadays most new apps are deployed in the cloud, and not being cloud native by default hurts the developer experience.

Logging on the standard output is part of the "12 factor" best practices (https://12factor.net/logs) and required to deploy on almost all modern deployment (Docker, Kubernetes, Heroku, AppEngine, App Platform, Symfony Cloud...).
13 factor also states that this behavior must not be configurable.

Being "cloud native" by default is mandatory in 2021 for a framework like Symfony.

If you want to write your logs on disk, it's just a single YAML file to change (it's why this change is part of Flex, in the code you own and can change, and not of the framework itself).

TomaszGasior:
The design of the change made in this PR  makes it very easy to forget about logs default configuration. As I wrote, if developer forgets about this little thing, the logs are lost in some cases. **Being safe for data for most use cases is much more important than "being cloud native".**

Still, it is possible to solve the issue by moving configuration to environment variable:
1. Move the value of monolog.handlers.nested.path to the environment variable like LOGGER_URL.
2. Add LOGGER_URL="%kernel.logs_dir%/%kernel.environment%.log" to `.env` and LOGGER_URL="php://stderr" to `.env.prod` from symfony recipe.
3. Developers deploying the app to shared hostings have to inspect .env.local.php generated from `composer symfony:dump-env`. Now, it's very hard to forget about changing logging path and it's even easier to do — another deployment is not needed!

What do you think?

stof:
@TomaszGasior and writing to a local file means that logs are lost for any cloud-native deployment, because you won't have access to the local filesystem to read the file.

maartendekeizer:
Maybe we can write to stdout/stderr and file by default?

TomaszGasior:
@stof That's the reason why my suggestion focuses on **discoverability**, not on reverting to the change back to file.

dunglas:
Discoverability is better done by writing docs than by bloating the generated code for every users.

Shared hosting environments is a thing from the past, let's not annoy every users with this.

Gemorroj:
I think we shoul add comment to the config at lest.
I lost 30 minutes while searching for why there is no data in the logs.

PR #895, opened 2021-02-09
Remove all symlinks

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Using symlinks doesn't play well with our new [policy for updating recipes](https://github.com/symfony/recipes#updating-recipes).

Also, symlinks don't work with versioning based on tree-hashes, as flex does.
I was the first one to use them, but now I propose to drop them all and ban them from recipes.
Ideally, we should enforce this rule with @symfony-flex-server if you agree.

/cc @fabpot I might need your help to implement this... :)


nicolas-grekas:
(CI failure is a false-positive)

weaverryan:
Are we waiting on anything for this? 

> Also, symlinks don't work with versioning based on tree-hashes, as flex does

I don't understand what this is referring to exactly, but I trust you that it's an issue. I'm 👍 for doin this. It's unfortunate in a sense, but, as you said, it promotes "keeping recipes for old versions alone".

fabpot:
I think we're mainly waiting for me :)
The biggest issue here is that diffs in PRs are now mostly irrelevant. Ideally, we should find a way to create a diff between an old version of a recipe and the new one.

nicolas-grekas:
Maybe flex-server could post a link to this diff? Or better: post the diff as a comment inside a `<details>` tag?

stof:
The flex server bot should ignore the `.github` folder when validating recipes. that folder does not contain recipes.

nicolas-grekas:
I added a check for symlinks in https://github.com/symfony/recipes/commit/b570ddec6796f8907a57bfc4365cac75814bbf28

PR #899, opened 2021-02-16
Added a recipe for psr-http-message-bridge

derrabus:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | N/A

A recipe to integrate symfony/psr-http-message-bridge#89

### DIFF
--- a/symfony/psr-http-message-bridge/2.1/config/packages/psr_http_message_bridge.yaml
+++ b/symfony/psr-http-message-bridge/2.1/config/packages/psr_http_message_bridge.yaml
@@ -1,0 +1,21 @@
+services:
+    _defaults:
+        autowire: true
+        autoconfigure: true
+
+    Symfony\Bridge\PsrHttpMessage\HttpFoundationFactoryInterface:
+        '@Symfony\Bridge\PsrHttpMessage\Factory\HttpFoundationFactory'
+
+    Symfony\Bridge\PsrHttpMessage\HttpMessageFactoryInterface:
+        '@Symfony\Bridge\PsrHttpMessage\Factory\PsrHttpFactory'
+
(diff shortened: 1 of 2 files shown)


symfony/psr-http-message-bridge/2.1/config/packages/psr_http_message_bridge.yaml
```diff
@@ -0,0 +1,20 @@
+services:
+    _defaults:
+        autowire: true
+        autoconfigure: true
+
+    Symfony\Bridge\PsrHttpMessage\HttpFoundationFactoryInterface:
+        class: Symfony\Bridge\PsrHttpMessage\Factory\HttpFoundationFactory
+
+    Symfony\Bridge\PsrHttpMessage\HttpMessageFactoryInterface:
+        class: Symfony\Bridge\PsrHttpMessage\Factory\PsrHttpFactory
```
nicolas-grekas, line 10: it'd be great to find a way to remove those two lines (see my comments on the PR on the bridge)

derrabus: Okay, let's keep the discussion on the bridge PR.

nicolas-grekas: You made a good point in the bridge PR: these are useful services on their own.
To really label them as useful services, the interfaces should be *aliases* to the actual implementation. This is the criteria that `debug:autowiring` needs to identify the "useful services".

derrabus: Both interfaces are aliases now. I've also changed the services in the comment and assigned service IDs to them.

nicolas-grekas (approved):

chalasr (approved):

PR #900, opened 2021-02-18
add support for override_url

jrushlow:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | https://github.com/doctrine/DoctrineBundle/pull/1290    

Depends on:
- [x] PR #1290 Merged into `2.3`

1290 proposes to add a new `doctrine.dbal.override_url` param that will allow you to override parts of a `DATABASE_URL`.

### DIFF
--- a/doctrine/doctrine-bundle/2.3/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/2.3/config/packages/doctrine.yaml
@@ -1,0 +1,19 @@
+doctrine:
+    dbal:
+        override_url: true
+        url: '%env(resolve:DATABASE_URL)%'
+
+        # IMPORTANT: You MUST configure your server version,
+        # either here or in the DATABASE_URL env var (see .env file)
+        #server_version: '13'
+    orm:
+        auto_generate_proxy_classes: true
+        naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
+        auto_mapping: true
+        mappings:
+            App:
+                is_bundle: false
+                type: annotation
+                dir: '%kernel.project_dir%/src/Entity'
+                prefix: 'App\Entity'
+                alias: App
--- a/doctrine/doctrine-bundle/2.3/config/packages/test/doctrine.yaml
+++ b/doctrine/doctrine-bundle/2.3/config/packages/test/doctrine.yaml
@@ -1,3 +1,5 @@
 doctrine:
     dbal:
+        # Overrides the database name in the test environment only
+        # "host", "port", "username", & "password" can also be set to override their respective url parts
         dbname: main_test
--- a/doctrine/doctrine-bundle/2.3/config/packages/test/doctrine.yaml
+++ b/doctrine/doctrine-bundle/2.3/config/packages/test/doctrine.yaml
@@ -1,5 +1,7 @@
-doctrine:
-    dbal:
-        # Overrides the database name in the test environment only
-        # "host", "port", "username", & "password" can also be set to override their respective url parts
-        dbname: main_test
+#doctrine:
+#    dbal:
+#        # Overrides the database name in the test environment only
+#        # "host", "port", "username", & "password" can also be set to override their respective url parts
+#        #
+#        # If you're using ParaTest, "TEST_TOKEN" is set by ParaTest otherwise nothing is appended to the database name.
+#        dbname: main_test%env(default::TEST_TOKEN)%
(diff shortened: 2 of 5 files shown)


doctrine/doctrine-bundle/2.3/config/packages/doctrine.yaml
```diff
@@ -0,0 +1,19 @@
...
+        auto_generate_proxy_classes: true
+        naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
+        auto_mapping: true
+        mappings:
+            App:
+                is_bundle: false
+                type: annotation
+                dir: '%kernel.project_dir%/src/Entity'
+                prefix: 'App\Entity'
+                alias: App
```
weaverryan, line 19: If a file is identical to a file in a previous version, we symlink the new version file to the old one to avoid duplication :)

jrushlow: Waiting on the outcome of https://github.com/symfony/recipes/pull/895 to see where this should go.. I figured that has a better chance of getting merged in before the doctrine pr does.

weaverryan: you're totally right - keep it - I didn't see that until now

jrushlow:
https://github.com/doctrine/DoctrineBundle/pull/1290 was merged in to 2.3 today - this recipe i believe is safe to merge in now

doctrine/doctrine-bundle/2.3/config/packages/doctrine.yaml
```diff
@@ -0,0 +1,19 @@
+doctrine:
+    dbal:
+        override_url: true
```
weaverryan, line 3: I don't think we need this here... just because the user won't be overriding anything.

But what we *should* have, I think, is a `config/packages/test/doctrine.yaml` that has this setting. AND... I'm tempted to actually set the `dbname: ` there it something... maybe even something that has `TEST_TOKEN` in it for Paratest.

OR, we could actually do nothing (not even make any changes to the recipe) and instead we rely on documentation to guide the user to create the `config/packages/test/doctrine.yaml` file.

jrushlow: not setting `override_url: true` while using a `url` triggers the new deprecation in the bundle.

`Not setting doctrine.dbal.override_url to true is deprecated. True is the only value that will be supported in doctrine-bundle 3.0.`

weaverryan: duh. I totally forgot :p.

But should we also have a `config/packages/test/doctrine.yaml` file? I kind think we should...

jrushlow: done.


>maybe even something that has TEST_TOKEN in it for Paratest.


I'm not familiar with Paratest, but I gather the `TEST_TOKEN` has to be unique in every test? I don't see how we could achieve that in the config - but I am probably missing something obvious.

doctrine/doctrine-bundle/2.3/config/packages/test/doctrine.yaml
```diff
@@ -0,0 +1,3 @@
+doctrine:
+    dbal:
+        dbname: main_test
```
weaverryan, line 3: Maybe add a comment?

```
# Overrides the database name in the test environment only
```

doctrine/doctrine-bundle/2.3/config/packages/doctrine.yaml
weaverryan: `TEST_TOKEN` is *provided* by ParaTest - they set it automatically. All we need to do is *read* that environment variable with a default value. If it's not there, then it would be empty and... who cares :). The idea would be that we have something like `main_test%env('TEST_TOKEN')%`

doctrine/doctrine-bundle/2.3/config/packages/doctrine.yaml
Nyholm: I think we should keep this.

Nyholm (approved):
Thank you. 

This diff is really small. 

```diff
 # /config/packages/doctrine.yaml
 doctrine:
     dbal:
+        override_url: true
         url: '%env(resolve:DATABASE_URL)%'
 
 # /config/packages/test/doctrine.yaml
+#doctrine:
+#    dbal:
+#        # Overrides the database name in the test environment only
+#        # "host", "port", "username", & "password" can also be set to override their respective url parts
+#        #
+#        # If you're using ParaTest, "TEST_TOKEN" is set by ParaTest otherwise nothing is appended to the database name.
+#        dbname: main_test%env(default::TEST_TOKEN)%
```

derrabus (approved):

Nyholm:
Wohoo. Thank you @jrushlow

PR #905, opened 2021-02-23
get mock storage from the factory

jrushlow:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | https://github.com/symfony/symfony/issues/40282

Get the session storage mock file from the storage factory. Possible solution for #40282

### DIFF
--- a/symfony/framework-bundle/5.3/config/packages/test/framework.yaml
+++ b/symfony/framework-bundle/5.3/config/packages/test/framework.yaml
@@ -1,4 +1,4 @@
 framework:
     test: true
     session:
-        storage_factory_id: session.storage.mock_file
+        storage_factory_id: session.storage.factory.mock_file


nicolas-grekas:
Not a solution for the linked issue, since recipe updates don't apply to existing projects.
But maybe a good default for new 5.3 projects of course.

/cc @jderusse

symfony/framework-bundle/5.3/config/packages/test/framework.yaml
```diff
@@ -1,4 +1,4 @@
 framework:
     test: true
     session:
-        storage_factory_id: session.storage.mock_file
+        storage_factory_id: session.storage.factory.mock_file
```
jderusse, line 4: good catch. That's a bug

jderusse (approved):

chalasr (approved):

PR #906, opened 2021-02-27
Starting recommending to run PHPUnit directly

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | TODO - important

Hi!

This is tricky, and probably controversial. My goal is to make running PHPUnit in Symfony projects "not special". This means that users can run `./vendor/bin/phpunit` like any other project (though, we will keep `bin/phpunit` as a shortcut and smooth mechanism to not break docs, etc).

My understanding (please correct me if I'm wrong) is that as long as users have `<listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener"/>` in their `phpunit.xml.dist` file, all of the features of `phpunit-bridge` are available to them. The one exception is the ability to run your tests against multiple PHPUnit versions, which is not a need that many projects have.

On a high level, here is how the new system would work:

1) When you `composer require test --dev`, you actually download `symfony/test-pack` (this is already the case). In https://github.com/symfony/test-pack/pull/10, I've added `phpunit/phpunit`. This means you'll get both `symfony/phpunit-bridge` and `phpunit/phpunit`.

2) When the recipes are executed, the `symfony/phpunit-bridge` runs first. This gives the user the `phpunit.xml.dist` file from `phpunit-bridge`, which is good because that has the `<listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener"/>` already in it.

3) At this point, the user can run `./vendor/bin/phpunit` directly. But, of course, they could also choose to continue executing `simple-phpunit` directly. But, as a shortcut (and to make this change smoother), the `bin/phpunit` file still exists, but now executes PHPUnit directly if it's installed directly.

**Details**

A) The experience has been optimized for the "main" user that just wants to run PHPUnit directly. In the `phpunit.xml.dist` file from the `phpunit-bridge` recipe, we have 2 spots where we favor config that assumes you're running things directly.

Cheers!


phpunit/phpunit/4.7/phpunit.xml.dist
```diff
@@ -24,4 +24,18 @@
...
+        <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener" />
+    </listeners>
+    -->
+
+    <!-- Run `composer require symfony/panther` before enabling this extension -->
+    <!--
+    <extensions>
+        <extension class="Symfony\Component\Panther\ServerExtension" />
+    </extensions>
+    -->
```
weaverryan, line 40: This is just to make this file consistent with the `phpunit.xml.dist` file from the `phpunit-bridge` recipe, in case someone installs `phpunit` directly (and not via the pack, where you will get the `phpunit.xml.dist` file from `phpunit-bridge`)

symfony/phpunit-bridge/5.3/bin/phpunit
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env php
+<?php
+if (file_exists(dirname(__DIR__).'/vendor/phpunit/phpunit/phpunit')) {
+    require(dirname(__DIR__).'/vendor/phpunit/phpunit/phpunit');
```
weaverryan, line 4: This is the new part: if PHPUnit is install directly, use it.

kbond:
I recently switched a large (4.4) app over to using paratest (with excellent results btw). Since paratest (and I believe we want to start promoting this) requires phpunit I switched to requiring phpunit directly in my app without issue.

wouterj:
:+1: I like this approach! I'm not sure what my approval triggers on flex recipes, so I'll leave this as a normal comment for now :)

A couple important takeaways from this PR imho:

* The simple way is now targeted at application developers, rather than library maintainers.
* This removes Symfony being "special with PHPunit". This allows removing some docs about PHPunit, and also allows referencing official PHPunit docs and blog posts. 
* This opens up using many community packages build for/on-top-of PHPunit in a Symfony application, such as [Paratest](https://github.com/paratestphp/paratest), [Collision](https://github.com/nunomaduro/collision) or [PestPHP](https://pestphp.com/).
* The "old way" is still supported (`vendor/bin/simple-test`), so no feature is lost with this change

wouterj:
Btw, I'm going to work on a testing docs rewrite soon (see https://github.com/symfony/symfony-docs/pull/14731). It would be great to coordinate merging this together with the doc changes, so the less resources is spent and a bigger "announcement" can be made.

javiereguiluz:
In my opinion, this hits the sweet spot: it allows to use standard PHPUnit practices while enjoying all the cool features of Symfony's PHPUnit Bridge. A big 👍  form me. Thanks Ryan!

phpunit/phpunit/4.7/phpunit.xml.dist
dunglas: A nice addition would be to have a PHPUnit configurator in Flex.

Nyholm:
This PR is marked as "[WIP]", what is the current blocker?

symfony/phpunit-bridge/5.3/phpunit.xml.dist
```diff
@@ -0,0 +1,45 @@
...
+         bootstrap="tests/bootstrap.php"
+>
+    <php>
+        <ini name="error_reporting" value="-1" />
+        <server name="APP_ENV" value="test" force="true" />
+        <server name="SHELL_VERBOSITY" value="-1" />
+        <!-- If you use phpunit-bridge directly, you can control extra behavior -->
+        <!--
+        <server name="SYMFONY_PHPUNIT_REMOVE" value="" />
+        <server name="SYMFONY_PHPUNIT_VERSION" value="8.5" />
```
noniagriconomie, line 19: what about defining `9.5` as value?
related to https://phpunit.de/supported-versions.html

derrabus: `8.5` is the safer default because it works for all PHP versions that Symfony 5 can be run on.

nicolas-grekas:
I'm certainly not happy with this PR, because requiring phpunit/phpunit as a direct dep is conceptually so wrong. Sebastian Bergmann himself discourages doing so. Yet everybody happily does it and :see_no_evil: with the issues it creates. Dependency hell is the first and major one, but since the majority doesn't use `--no-dev` when installing deps on prod servers, it also means runtime overhead to load functions declared by phpunit et al., exposition to possible unserialize()+destructor exploits, etc.

I don't have the will to block this move, because I also understand ppl that complain about simple-phpunit - it's not a perfect solution either. I think the tradeoffs it offers are worth it, but if the majority doesn't agree, let's merge.

:+1: from the purely technical aspect for the patch.

nicolas-grekas:
For the record, see https://github.com/composer/composer/issues/9636, which lokks for a better solution to these issues. No ready, so not a blocker for this PR.

Nyholm:
I do see your point and I agree that this is an issue that should be solved. But it is not a "symfony issue", it is more an issue of how the php community is using tools like phpunit. 

I think symfony would benefit to do "like everybody else" in this case, even though it is a technically worse solution.

nicolas-grekas:
See also https://github.com/composer/composer/pull/9792

weaverryan:
Here's the plan:

A) Merge this recipe. It applies to phpunit-bridge 5.3 only.

B) When 5.3 comes out, THEN merge and tag adding phpunit/phpunit to test-pack https://github.com/symfony/test-pack/pull/10

The timing on (B) is just to reduce (but definitely not eliminate, think 4.4. users) the number of people who would install test-pack but not get the new recipe. This recipe updates `bin/phpunit` to use `phpunit/phpunit` directly. So if you don't get this recipe, then you will still be using `phpunit-bridge`, which is fine... except that it could cause some class versioning issues where `phpunit-bridge` downloads one version of PHPUnit, then the user's app has another version autoloading from their own `vendor/`.

I have just tested this recipe, both with and without `phpunit/phpunit` directly installed.

nicolas-grekas (approved):

wouterj:
> The timing on (B) is just to reduce (but definitely not eliminate, think 4.4. users) the number of people who would install test-pack but not get the new recipe.

Note that we unpack packs by default. So you would need to install a fresh `test-pack` in a <5.3 application to get this situation.

Nyholm (approved):

weaverryan:
I'm coordinating with Wouter to make sure the docs will be ready for 5.3

symfony/phpunit-bridge/5.3/bin/phpunit
VincentLanglet: I tried adding this to my bin/phpunit file and I'm getting 
```
PHP Fatal error:  strict_types declaration must be the very first statement in the script
```

Nyholm: Did you try without strict type declaration?
This file don't include any.

nicolas-grekas: Already fixed in https://github.com/symfony/recipes/pull/952

VincentLanglet: > Already fixed in #952

Indeed, thanks

PR #908, opened 2021-03-05
Add zenstruck/foundry

kbond:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | symfony/symfony-docs#14731

Based on slack discussions with @weaverryan & @wouterj.

Foundry has the concept of [auto-refreshing entities](https://github.com/zenstruck/foundry#auto-refresh) - prior to 1.9, this was disabled by default as there are caveats (entity changes silently lost). In 1.9, the caveats still exist but we were [able to throw a helpful exception](https://github.com/zenstruck/foundry/pull/131). We would now like auto-refreshing entities enabled by default but this would be a BC break in 1.x. A recipe to enable in 1.x was decided as the next best thing.

The double config is unfortunate but we couldn't think of a better way.

Likely we'll want a `post-install.txt` and probably some config comments?

### DIFF
--- a/zenstruck/foundry/1.9/config/packages/test/zenstruck_foundry.yaml
+++ b/zenstruck/foundry/1.9/config/packages/test/zenstruck_foundry.yaml
@@ -1,0 +1,5 @@
+# Unless you want different configuration for test/dev environments,
+# add configuration to config/packages/dev/zenstruck_foundry.yml
+# and this will be synced to your test environment.
+imports:
+    - { resource: ../dev/zenstruck_foundry.yaml }
--- a/zenstruck/foundry/1.9/config/packages/dev/zenstruck_foundry.yaml
+++ b/zenstruck/foundry/1.9/config/packages/dev/zenstruck_foundry.yaml
@@ -1,0 +1,4 @@
+# See full configuration: https://github.com/zenstruck/foundry#full-default-bundle-configuration
+zenstruck_foundry:
+    # Whether to auto-refresh proxies by default (https://github.com/zenstruck/foundry#auto-refresh)
+    auto_refresh_proxies: true
--- a/zenstruck/foundry/1.9/manifest.json
+++ b/zenstruck/foundry/1.9/manifest.json
@@ -1,0 +1,8 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "bundles": {
+        "Zenstruck\\Foundry\\ZenstruckFoundryBundle": ["dev", "test"]
+    }
+}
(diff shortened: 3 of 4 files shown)


Nyholm:
Thank you @kbond, Is this still "WIP"?

I would be happy to merge this recipe.

zenstruck/foundry/1.9/config/packages/test/zenstruck_foundry.yaml
```diff
@@ -0,0 +1,2 @@
+imports:
+    - { resource: ../dev/zenstruck_foundry.yaml }
```
OskarStark, line 2: I would not use an import here, the complexity for one option is not worth.

Any reasons why we should do it this way @kbond ?

kbond: This was done to avoid users having to duplicate their config in test and dev. There are several other configuration options for foundry and I think the common scenario would be for them to be the same in test/dev.

What do you suggest?

kbond:
I removed the WIP/RFC. I believe this will work as expected as is. Just a few outstanding questions:
1. Should I add a `post-install.txt`?
2. Should the double config be handled differently or comments added for an explaination?
3. Any other comments in the config?

kbond:
FYI, `zenstruck/foundry` [1.9](https://github.com/zenstruck/foundry/releases/tag/v1.9.0) is now released.

Nyholm:
> Should I add a post-install.txt?

If there is a reason to add one, then yes. Im not sure what that reason is. What would you like to tell the user?

> Should the double config be handled differently or comments added for an explaination?

Will the config always be the same? If so, we can keep it. We are currently only using one config options, so we can also keep it simple. 
It does not matter too much for me. I would do what @OskarStark suggestes.

> Any other comments in the config?

You could add a link to your bundles docs.

zenstruck/foundry/1.9/config/packages/test/zenstruck_foundry.yaml
kbond: What I'm trying to avoid is: when a user decides they need to add additional config, they don't miss adding it in two places (as I believe almost always the user would want the same config for test/dev).

kbond:
I added some config comments and a `post-install.txt` for review.

zenstruck/foundry/1.9/config/packages/test/zenstruck_foundry.yaml
OskarStark: Makes sense, thanks for your explanation 👍

OskarStark (approved):

Nyholm (approved):
Thank you

PR #910, opened 2021-03-10
Integrate the PsrHttpMessage bundle

derrabus:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | N/A

Integrates symfony/psr-http-message-bridge#94

### DIFF
--- a/symfony/psr-http-message-bridge/2.2/config/packages/psr_http_message_bridge.yaml
+++ b/symfony/psr-http-message-bridge/2.2/config/packages/psr_http_message_bridge.yaml
@@ -1,0 +1,5 @@
+psr_http_message:
+
+    # Message converters that allow controllers to operate on PSR-7 requests/responses
+    message_converters:
+        enabled: false
--- a/symfony/psr-http-message-bridge/2.2/manifest.json
+++ b/symfony/psr-http-message-bridge/2.2/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "Symfony\\Bridge\\PsrHttpMessage\\Bundle\\PsrHttpMessageBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    }
+}


symfony/psr-http-message-bridge/2.2/config/packages/psr_http_message_bridge.yaml
```diff
@@ -0,0 +1,11 @@
+psr_http_message:
+
+    # PSR-17 factories using the selected implementation (nyholm or diactoros)
+    # This functionality is enabled by default if nyholm/psr7 is available.
+    #message_factories:
+    #    enabled: true
+    #    implementation: nyholm
```
nicolas-grekas, line 7: Can we remove this option altogether and default to what is available? nyholm's by default when both are found? Maybe we can keep the option, but I would not advertise it in the recipe if we can.

symfony/psr-http-message-bridge/2.2/config/packages/psr_http_message_bridge.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+    # Message converters that allow controllers to operate on PSR-7 requests/responses instead of HttpFoundation
```
nicolas-grekas, line 9: not "instead of" - we can remove the end of the sentence

derrabus: Agreed, changed.

symfony/psr-http-message-bridge/2.2/config/packages/psr_http_message_bridge.yaml
derrabus: I'll have a look.

derrabus: I've removed the block from the recipe. The bundle should now enable and configure diactoros if it's the only available implementation.

Nyholm (approved):
Thank you

Nyholm:
@symfony-flex-server review please

nicolas-grekas:
Maybe we should drop the recipe altogether? I don't see a big value in having it. Don't we have this policy?

Nyholm:
If you install the bridge, you are very likely to use the `psr_http_message.message_converter.enabled: true`. That is why it should be a recipe. 

I also agree that the default should be `false`. 

Ie, a recipe should have sensible default and show options that are likely to change.

nicolas-grekas:
Are you sure? Eg I use sentry, which will require the bridge, but I'm never going to use the message converters :)

Nyholm:
Correct. That is why the default is `false`. 

The question is: Should we use a recipe or not?
Which is the same as: Should we create an config file with the default config or not?

I believe many people would like to change the default, that is why I think the config file should be created.

symfony/psr-http-message-bridge/2.2/config/packages/psr_http_message_bridge.yaml
```diff
@@ -0,0 +1,5 @@
+psr_http_message:
+
+    # Message converters that allow controllers to operate on PSR-7 requests/responses
+    message_converters:
+        enabled: false
```
weaverryan, line 5: Do we need any config file? In practice, of a library depends on `symfony/psr-http-message-bridge`... they would also get the bundle, which would give them this config file. New config files cause bloat. I'd prefer to not ship any config file. It already looks like it's usually not needed. I also just suggested the possibility of even enabling this part by default (https://github.com/symfony/psr-http-message-bridge/pull/94/files#r592669310) but I don't know if there are downsides to making this the default :)

nicolas-grekas: Enabling by default means registering an event listener on the hotpath, better not.
But I share the feeling about the recipe, it might not be needed.

weaverryan: Ok, I'd vote for disabling the listeners by default (as they are done already) but no config file in the recipe.

derrabus:
The reason why I proposed a new recipe for my bridge PR were:

* I wanted to have a new recipe that replaces the current problematic one
* I wanted the bundle to be registered automatically
* I wanted to make the option for enabling the message converters discoverable as this is a common reason for installing the bridge.

nicolas-grekas:
> I wanted to have a new recipe that replaces the current problematic one
> I wanted the bundle to be registered automatically

If we remove the recipe for the bridge altogether, for 2.1 I mean, both items will be resolved.

Item 3 remains, and on my side, I prefer empty config folders to begin with. There are so many potential packages to configure - they all compete for attention. Having just the strict minimum is my preference, to prevent building a deep tree of commented option.

derrabus:
> If we remove the recipe for the bridge altogether, for 2.1 I mean, both items will be resolved.

Is there anything I have to do on the other PR to make the bundle discoverable for Flex?

Nyholm:
> Is there anything I have to do on the other PR to make the bundle discoverable for Flex?

Generally, no. 
But in this package: yes. 

I think Flex is only looking for a bundle in the "default" location. You need to have a manifest with the `bundles` key. You can remove the config file. 

I think this will work even tough the package is a "library".

Issue #909, opened 2021-03-09
New symfony/psr-http-message-bridge recipe KO's out-of-the-box with sentry/symfony


weaverryan:
Hi!

If you `composer require symfony/psr-http-message-bridge`, you get the new https://github.com/symfony/recipes/blob/master/symfony/psr-http-message-bridge/2.1/config/packages/psr_http_message_bridge.yaml config file.

This service config is, sort of, "invalid" when you first receive it. For example, if you actually tried to use the `PsrHttpFactory` service, you would get this error:

> Cannot autowire service "Symfony\Bridge\PsrHttpMessage\Factory\PsrHttpFactory":
> argument "$serverRequestFactory" of method "__construct()" references   
> interface "Psr\Http\Message\ServerRequestFactoryInterface" but no such service exists.
> Did you create a class that implements this interface?

The reason - which is included in the documentation - is that you also need to install an implementation - like `nyholm/psr7`.

That's not a great error messagae. But the *real* problem is when another package depends on `symfony/psr-http-message-bridge` - for example `sentry/sentry-symfony`. If you install it right now, it's currently broken (with the above error). This is what happens:

A) The symfony/psr-http-message-bridge recipe adds the (sort of) incomplete services file, which includes a non-functional `Symfony\Bridge\PsrHttpMessage\HttpMessageFactoryInterface` service, because it's an alias to the non-functional `Symfony\Bridge\PsrHttpMessage\Factory\PsrHttpFactory` service.

B) sentry/sentry-symfony uses `Symfony\Bridge\PsrHttpMessage\HttpMessageFactoryInterface` as an optional dependency - https://github.com/getsentry/sentry-symfony/blob/master/src/Resources/config/services.xml#L68

Everything works fine if there IS no `Symfony\Bridge\PsrHttpMessage\HttpMessageFactoryInterface` service. However, there now IS one... but it's broken, which gets you KO'ed on `composer install sentry/sentry-symfony`.

Sorry for the long explanation - it took me awhile to sort out the problem. 

**Possible Solution**

I would prefer if `symfony/psr-http-message-bridge` either came with no recipe (and the needed config is documented) or if it came with the config file, but commented-out. The user already needs to do more work to *actually* make the services functional.

I also don't use this library... so I don't know if its main use-case will be as a library that people directly install (and so we want to make that experience as smooth as possible) or if its an indirect dependency because people are installing things like sentry/sentry-symfony. In either case, adding the new services "broke the experience" of installing that library.

Thanks!

[...]

PR #911, opened 2021-03-17
Wire symfony/psr7-pack

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Fix https://github.com/symfony/psr-http-message-bridge/pull/94
Fix https://github.com/symfony/recipes/pull/910
Fix https://github.com/symfony/recipes/issues/909

Requires registering https://github.com/symfony/psr7-pack on packagist /cc @fabpot 

/cc @derrabus @Nyholm 

I think this approach good enough: less code to maintain and solves the issue of keeping things decoupled.
See discussions on the linked PRs.


stof:
I don't think that works, actually. As packs are unpacked automatically by Flex, you won't have any recipes to apply for them. Packs are only getting aliases in their recipes, not install-time content

Nyholm (changes requested):
The manifest `psr-http-message-bridge/2.1/manifest.json` is now invalid. 

Did you really mean to move the `config/packages/psr_http_message_bridge.yaml`?

nicolas-grekas:
> The manifest psr-http-message-bridge/2.1/manifest.json is now invalid.

It's been moved to `psr7-pack` also, there is no recipe for `psr-http-message-bridge` anymore in this PR. This follows https://github.com/symfony/recipes/issues/909#issuecomment-795446904

Nyholm (approved):
Thank you

fabpot (approved):

PR #913, opened 2021-03-22
Allow .php config files (and others) to be loaded in the Kernel by default.

dreadnip:
| Q             | A
| ------------- | ---
| License       | MIT

I'd like to suggest a partial rollback of this [PR](https://github.com/symfony/recipes/pull/705) by @nicolas-grekas. Since both Symfony [internally](https://github.com/symfony/symfony/issues/37186) and the [community](https://tomasvotruba.com/blog/2020/07/16/10-cool-features-you-get-after-switching-from-yaml-to-php-configs/) are moving in the direction of PHP configs, I feel like it'd be nice to allow these out-of-the-box when using the framework bundle. 

Right now if you switch to PHP config files, either manually or by using a [tool](https://github.com/symplify/config-transformer), you end up with a broken Symfony app. If we import .php config files in the default kernel, this would work straight away.

My apologies if this PR is misplaced in any way, it is my first attempt to contribute and I couldn't quite figure out which repo I had to submit the PR to.

### DIFF
--- a/symfony/framework-bundle/5.3/src/Kernel.php
+++ b/symfony/framework-bundle/5.3/src/Kernel.php
@@ -11,10 +11,12 @@ class Kernel extends BaseKernel
 {
     use MicroKernelTrait;
 
+    private const CONFIG_EXTS = '{php,xml,yaml,yml}';
+
     protected function configureContainer(ContainerConfigurator $container): void
     {
-        $container->import('../config/{packages}/*.yaml');
-        $container->import('../config/{packages}/'.$this->environment.'/*.yaml');
+        $container->import('../config/{packages}/*.' . self::CONFIG_EXTS);
+        $container->import('../config/{packages}/'.$this->environment.'/*.' . self::CONFIG_EXTS);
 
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
I think we're not ready to make this move. Yaml is still the best format in `packages/`
Tomas is doing nice experiments, but the drawbacks are still huge and we're still looking for a smoother way.

:-1: on my side sorry.

Nyholm:
Thank you. 
I think this PR is correct, but the timing is not right as Nicolas says, but could we maybe find a compromise? Could we add a comment like: 

```php
// Uncomment this line to allow configuration with PHP files
// $container->import('../config/{packages}/*.php');
```

nicolas-grekas:
> could we maybe find a compromise? Could we add a comment like:

We try to keep recipes as clean as possible, that's why I'd prefer not...

Nyholm:
Hm. Okey. @dreadnip. I'll close this PR now. Could you open a similar PR in 5-6 months? I think we are more mature with the idea then. It is also a few issues we should take care of before fully embracing php config. 


I appreciate that you took your time to make this PR. Thank you.

PR #916, opened 2021-03-25
Routing: don't require to install doctrine/annotations to use attributes

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

Currently, installing `doctrine/annotations` is needed to be able to use the `#[Route()]` PHP 8 attribute. It'unfortunate because attributes can work out of the box and this dependency is not needed with PHP 8.

This PR enables "annotation" support (actually attribute support) by default, even if `doctrine/annotations` isn't installed.
This PR targets Symfony 5.3 as it's a new feature. A side effect will be that the default skeleton will not work by default with PHP 7, but IMHO it's not a big issue because most new projects using Symfony 5.3 should use PHP 8. Also, this moves the ecosystem forward.

### DIFF
--- a/symfony/routing/6.0/config/packages/routing.yaml
+++ b/symfony/routing/6.0/config/packages/routing.yaml
@@ -1,0 +1,12 @@
+framework:
+    router:
+        utf8: true
+
+        # Configure how to generate URLs in non-HTTP contexts, such as CLI commands.
+        # See https://symfony.com/doc/current/routing.html#generating-urls-in-commands
+        #default_uri: http://localhost
+
+when@prod:
+    framework:
+        router:
+            strict_requirements: null
--- a/symfony/routing/6.0/config/routes.yaml
+++ b/symfony/routing/6.0/config/routes.yaml
@@ -1,0 +1,7 @@
+controllers:
+    resource: ../src/Controller/
+    type: annotation
+
+kernel:
+    resource: ../src/Kernel.php
+    type: annotation
--- a/symfony/routing/6.0/manifest.json
+++ b/symfony/routing/6.0/manifest.json
@@ -1,0 +1,6 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "aliases": ["router"]
+}


chalasr (dismissed):
Better DX for the 80% use case. 👍

nicolas-grekas:
I think @derrabus had a plan about this, let's ask him if there are any missing parts?

weaverryan:
> A side effect will be that the default skeleton will not work by default with PHP 7

We need to make this PR happen, but this is problem. What exactly would happen (what error) if someone uses PHP 7 on 5.3 after this recipe? Could we only error (and with a clear error) if they started trying to use annotations without doctrine/annotations installed?

Thanks!

nicolas-grekas:
According to our past policies, I'm :-1: to break the experience of PHP 7 users.

This is 6 months too early. Let's wait for Symfony 6 to do this.

nicolas-grekas:
I don't know if that's going to be acceptable, but technically, we achieve something condition by using the kernel instead here.

dunglas:
The current experience is not acceptable either. You need to install a totally useless library just to generate some config to use a native feature.

With this PR, users wanting to use PHP 7 will just have to comment two blocks of configuration if they really want to do so. We're just talking about new projects here. Will users start a new project with Symfony 5.3 (to be released only in May) with a version of PHP that will not be supported in November of the same year? We should make the experience good for the most common setup for new projects (latest PHP version and latest Symfony version), not degrade it for everyone to support legacy configurations.

An alternative (less good, but that can allow reaching consensus) could be to just comment the lines added by this PR. So it will work out of the box for PHP 7, and PHP 8 users will be able to discover how to make the app working for them. In my opinion it's less good than making the experience good for PHP 8 users, but it's still better than the current situation (and this could be done even for 5.2).

nicolas-grekas:
Most ppl will have to install doctrine/annotations for entities anyway, so I wouldn't call it useless (yet). What about my proposal to use the Kernel instead? We can wait 6 months to do this in Symfony 6 IMHO. Doing this in 5.4 LTS is too "aggressive" IMHO.

dunglas:
Only those using Doctrine ORM, and hopefully attributes support will land in the ORM before the release of Symfony 5.3.

Regarding adding code in the kernel, I'm not a fond of "polluting"/making more complex the generated code for supporting things like this. Adding commented configs is less intrusive IMHO.

nicolas-grekas:
I just ran:
```sh
SYMFONY_ENDPOINT=https://flex.symfony.com/r/github.com/symfony/recipes/916 symfony new --version=dev test
```

and here is the output. This is a blocker on my side. :-1: as is.


SYMFONY_ENDPOINT=https://flex.symfony.com/r/github.com/symfony/recipes/916 symfony new --version=dev test

nicolas-grekas:
image

dunglas:
Are you using PHP 8? Here is what I get:

image

To prevent this error message with PHP 7 we can either... comment the config I added in this PR or make PHP 8 mandatory in the `composer.json` of the skeleton.

Also, I opened a PR to improve the error message: https://github.com/symfony/symfony/pull/40586

derrabus:
I agree with Nicolas that we should not add a recipe that results in an instant KO on PHP 7. An alternative would be to make `resource: annotation` an no-op on PHP 7 without `doctrine/annotations`.

Another problem I see is that the `doctrine/annotations` recipe installs yet another routes configuration for annotations. Maybe we should synchronize the file names of both recipes so we don't end up with redundant route configurations.

nicolas-grekas (changes requested):
(trying to block any accidental merge, see discussion)

dunglas:
+1 regarding the filename, I'll update the PR.

Regarding the no-op, I fear that it will be very hard to debug. Isn't a proper error message better than a no-op?

derrabus:
> Regarding the no-op, I fear that it will be very hard to debug.

I share that concern.

> Isn't a proper error message better than a no-op?

I agree, but what would be the proper channel for that error message? The exception breaks `composer create-project symfony/skeleton` on PHP 7 if we merged your recipe.

dunglas:
Sure. I don't think that we many choices here. Either we display an error message for PHP 7 users telling them to upgrade their PHP version or to run `composer req annotations` (it's what I suggest, and that's will be the case when my my PR on SF will be merged), or we comment the new configuration so the installation will not fail for PHP 7 users, and PHP 8 users will be able to discover this feature (even if it's far from ideal).

I'm in favor of the first solution (because I think that our main target should be PHP 8 users). The second solution is less good but acceptable too imho (it's still improvement compared to the current situation).

Tobion:
I would propose to not throw an annotation loader exception when there are no php files in the directory yet.
So on a new install which does not have any controller classes yet, there won't be an exception. So it won't block people from installing symfony on php 7. Only when they add a controller, they will get an exception telling them to either use php 8 attributes or install doctrine/annotations.

dunglas:
We now have another related problem: DoctrineBundle supports attributes, but unlike Symfony components, it isn't able to load attributes and annotations at the same time.
IMHO, this another good reason to switch to attributes by default as PHP 7 is getting less and less popular (especially for new projects, which are the main target of Flex).

derrabus:
@dunglas We can do so safely for 6.0 at least.

weaverryan:
Should we change this to a 6.0 recipe and merge now?

fabpot:
Let's do it for 6.0.

dunglas:
PR updated to target 6.0.

weaverryan (approved):
Wonderful. I am SO excited about attribute-routing out-of-the-box (with no extra packages) ❤️

derrabus (approved):

derrabus:
Looks like the CI cannot handle 6.0 recipes yet.

PR #924, opened 2021-04-03
Add alias for turbo

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/ux-turbo-mercure/1.3/manifest.json
+++ b/symfony/ux-turbo-mercure/1.3/manifest.json
@@ -1,0 +1,3 @@
+{
+    "aliases": ["turbo", "hotwire"]
+}


nicolas-grekas (approved):

derrabus (approved):

chalasr (approved):

OskarStark (approved):

fabpot (changes requested):
Turbo is not linked to Mercure. This is just part of Turbo. `composer require ux-turbo` already works as this is the Symfony package name of the main Turbo integration. I think there is nothing to do.

dunglas:
Without the Mercure part, the recipe and the bundle are almost useless. To use Turbo Drive, Turbo Frames and the basic features of Turbo Streams all you need is to add this to your page:

```html
<script src="https://cdn.skypack.dev/@hotwired/turbo"></script>
```

The added value of this bundle is the integration with Mercure and Doctrine. The integration with MakerBundle also require Mercure (or another transport) to be installed.

For instance, the official Rails gem comes with ActionCable (the equivalent of MercureBundle in the Rails ecosystem) by default.

In terms of discoverability and usability, I think that it's better for the user to have the Turbo Streams integration + the default transport enabled when you run `composer req turbo`.

tgalopin:
If I'm not missing something, currently, the alias ux-turbo already exists and is wired to symfony/ux-turbo due to the convention of Symfony Flex of trying to resolve packages with the symfony prefix.

This PR would override this behavior and ship a different package instead, I don't think that's wise.

I agree with Fabien: the plug and play nature of Flex and UX is enough IMO to justify having ux-turbo as an independent package. Turbo is very useful by itself (I'll mostly use it without any transport I think) and relying on a CDN is obviously not a good idea when you want to profit from the benefits of having a package manager and a build pipeline.

I'd be in favor of having:
* ux-turbo and turbo pointing to ux-turbo
* ux-turbo-mercure, turbo-mercure and also maybe turbo-streams pointing to ux-turbo-mercure

dunglas:
> HRrelying on a CDN is obviously not a good idea when you want to profit from the benefits of having a package manager and a build pipeline.

If you have Webpack (Encore) can just do `yarn add @hotwired/turbo` and add `import * as Turbo from "@hotwired/turbo"` to your main JS file then. And you don't get the added complexity of the bundle.
As I explained privately, the main benefit in terms of DX of this bundle is the Streams part. Having the ability of installing NPM packages through Composer has little benefits and doesn't worth the added complexity (it's also why I wouldn't have split these two packages).

PR #925, opened 2021-04-05
Load environment configured services and routes for php extension config

Ferror:
## Description
Documentation states that application loads `config/services_<environment-name>.yaml (and services_<environment-name>.xml and services_<environment-name>.php files too).` https://symfony.com/doc/current/configuration.html, but for me, it looks like it does not auto load PHP configuration - I know that Symfony does no support XML configuration anymore, but as far as I know we would like to keep PHP config files.

## Checks
| Q             | A
| -------------- | ---
| License            | MIT
| Doc issue/PR   | PR 

## Example
* For `test` environment it will load `services.php`, `services_test.php`
* For `dev` environment it will load `services.php`, `services_dev.php`
etc...


Tobion:
The idea is that you can check for the env inside the file. See https://github.com/symfony/recipes/pull/721/files 
Also considering https://symfony.com/blog/new-in-symfony-5-3-configure-multiple-environments-in-a-single-file I don't think this will be relevant anymore. 
Please fix the documentation instead.

PR #926, opened 2021-04-05
Fix deprecation in 5.3' session config

chalasr:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  |  -

This fixes the following notice on symfony/framework-bundle:5.3-dev

 > Since symfony/framework-bundle 5.3: Not setting the "framework.session.storage_factory_id" configuration option is deprecated, it will default to "session.storage.factory.native" and will replace the "framework.session.storage_id" configuration option in version 6.0.

/cc @jderusse


derrabus (approved):

jderusse (approved):
good catch @chalasr

PR #938, opened 2021-04-18
Log in var/log in dev

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

This PR partially reverts a change done in #894: logging on stderr by default in the dev env breaks the DX, because the output of console commands becomes a mess, with log line mangling the command's output.


dunglas (approved):

ogizanagi (approved):

dunglas:
For the record, logging in the terminal in the dev env is part of 12 factor (and mandatory when developing using Docker / Docker Compose): https://12factor.net/logs

PR #943, opened 2021-05-06
Suggest data-turbo-track="reload" attribute for Turbo users in Encore

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | TODO

This should be... just this simple! We should have a `data-turbo-track` on our `script` and `link` tags. The result is that, when the versioned filename changes, the full page will refresh.

Ref: https://turbo.hotwire.dev/handbook/drive#reloading-when-assets-change

Thanks!

### DIFF
--- a/symfony/webpack-encore-bundle/1.9/config/packages/webpack_encore.yaml
+++ b/symfony/webpack-encore-bundle/1.9/config/packages/webpack_encore.yaml
@@ -7,7 +7,12 @@ webpack_encore:
     # Set attributes that will be rendered on all script and link tags
     script_attributes:
         defer: true
+        # Uncomment (also under link_attributes) if using Turbo Drive
+        # https://turbo.hotwire.dev/handbook/drive#reloading-when-assets-change
+        # 'data-turbo-track': reload
     # link_attributes:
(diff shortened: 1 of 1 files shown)


dunglas (approved):

symfony/webpack-encore-bundle/1.9/config/packages/webpack_encore.yaml
```diff
@@ -7,7 +7,12 @@ webpack_encore:
     # Set attributes that will be rendered on all script and link tags
     script_attributes:
         defer: true
+        # Uncomment (also under link_attributes) if using Turbo Drive
+        # https://turbo.hotwire.dev/handbook/drive#reloading-when-assets-change
+        # 'data-turbo-track': reload
```
nicolas-grekas, line 12: why not uncomment by default?

weaverryan: Feels weird to sprinkle this attribute for everyone, including on non-Turbo projects

Nyholm (approved):
Thank you

PR #947, opened 2021-05-10
Update note about server version

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

As discussed in https://github.com/symfony/recipes/pull/939#discussion_r629363629

### DIFF
--- a/doctrine/doctrine-bundle/2.4/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/2.4/config/packages/doctrine.yaml
@@ -1,7 +1,10 @@
 doctrine:
     dbal:
-        # IMPORTANT: DATABASE_URL *must* define the server version
         url: '%env(resolve:DATABASE_URL)%'
+
+        # IMPORTANT: You MUST configure your server version,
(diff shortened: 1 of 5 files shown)


chalasr (approved):

doctrine/doctrine-bundle/2.4/config/packages/doctrine.yaml
```diff
@@ -1,6 +1,6 @@
 doctrine:
     dbal:
-        # IMPORTANT: DATABASE_URL *must* define the server version
+        # IMPORTANT: this "url" setting *must* define the "serverVersion" if you want to build the cache offline
```
stof, line 3: The alternative is to configure the `server_version` key separately if you cannot have it in the URL (which is a lot more maintainable than trying to concatenate extra query string parameters here if the env variable does not contain it).

guilliamxavier: Yes, that's what I meant in the referenced comment, but I feel like there is a will to stop "publicizing" the separate `server_version` setting?

stof (approved):

PR #948, opened 2021-05-12
[twig-bundle] uncomment `encore_*_tags()` when 5.4 is used

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | https://github.com/symfony/symfony-docs/issues/15457

Needs https://github.com/symfony/symfony/pull/41205

### DIFF
--- a/symfony/twig-bundle/5.4/config/packages/twig.yaml
+++ b/symfony/twig-bundle/5.4/config/packages/twig.yaml
@@ -1,0 +1,6 @@
+twig:
+    default_path: '%kernel.project_dir%/templates'
+
+when@test:
+    twig:
+        strict_variables: true
--- a/symfony/twig-bundle/5.4/templates/base.html.twig
+++ b/symfony/twig-bundle/5.4/templates/base.html.twig
@@ -1,0 +1,18 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="UTF-8">
+        <title>{% block title %}Welcome!{% endblock %}</title>
+        {# Run `composer require symfony/webpack-encore-bundle` to start using Symfony UX #}
+        {% block stylesheets %}
+            {{ encore_entry_link_tags('app') }}
+        {% endblock %}
+
+        {% block javascripts %}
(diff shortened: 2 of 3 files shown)


symfony/twig-bundle/5.4/config/packages/twig.yaml
```diff
@@ -0,0 +1,4 @@
+twig:
+    default_path: '%kernel.project_dir%/templates'
+    debug: '%kernel.debug%'
+    strict_variables: '%kernel.debug%'
```
weaverryan, line 4: It looks like this file is a bit simpler in the current, 5.0 recipe: https://github.com/symfony/recipes/blob/master/symfony/twig-bundle/5.0/config/packages/twig.yaml

Everything else looks good

nicolas-grekas: updated, thanks

weaverryan (approved):

nicolas-grekas:
PR ready!

javiereguiluz (approved):
Nice one! Thanks.

chalasr (approved):

stof:
@nicolas-grekas something looks wrong in the comment of the github-actions-bot. There is no diff between 5.3 and 5.4, while it is the relevant one here as you are adding the 5.4 one.

symfony/twig-bundle/5.4/templates/base.html.twig
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="UTF-8">
+        <title>{% block title %}Welcome!{% endblock %}</title>
+        {# Run `composer require symfony/webpack-encore-bundle` to start using Symfony UX #}
```
dunglas, line 6: Shouldn't this comment be removed?

nicolas-grekas: we can keep it, it's a nice ad ;)

PR #949, opened 2021-05-20
Added the 3 first Translation Providers

welcoMattic:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | 

This PR adds recipes for the new Translation Providers, added in 5.3 (https://github.com/symfony/symfony/pull/38475)

cc @fabpot @nicolas-grekas @Nyholm

### DIFF
--- a/symfony/translation/5.3/config/packages/translation.yaml
+++ b/symfony/translation/5.3/config/packages/translation.yaml
@@ -1,0 +1,13 @@
+framework:
+    default_locale: en
+    translator:
+        default_path: '%kernel.project_dir%/translations'
+        fallbacks:
+            - en
+#        providers:
+#            crowdin:
+#                dsn: '%env(CROWDIN_DSN)%'
+#            loco:
+#                dsn: '%env(LOCO_DSN)%'
+#            lokalise:
+#                dsn: '%env(LOKALISE_DSN)%'
--- a/symfony/crowdin-translation-provider/5.3/manifest.json
+++ b/symfony/crowdin-translation-provider/5.3/manifest.json
@@ -1,0 +1,5 @@
+{
+    "env": {
+        "#1": "CROWDIN_DSN=crowdin://PROJECT_ID:API_TOKEN@ORGANIZATION_DOMAIN.default"
+    }
+}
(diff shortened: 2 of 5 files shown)


Nyholm:
Thank you. Have a look at the Mailer packages. 

Basically, We have some config for `symfony/mailer`: 

```yaml
framework:
    mailer:
        dsn: '%env(MAILER_DSN)%'
```

And for a bridge (`symfony/mailjet`) they just add the `MAILER_DSN` environment variable.

symfony/translation/5.3/config/packages/translation.yaml
```diff
@@ -0,0 +1,15 @@
+framework:
+    default_locale: en
+    translator:
+        default_path: '%kernel.project_dir%/translations'
+        fallbacks:
+            - en
+#        providers:
```
fabpot, line 7: I think most people will only ever use 1 provider, so having something similar to what we have in Mailer as mentioned by Tobia sounds better.

fabpot (changes requested):

symfony/translation/5.3/config/packages/translation.yaml
welcoMattic: Did you mean only 1 exemple provider, like:
```yaml
        providers:
            crowdin:
                dsn: '%env(CROWDIN_DSN)%'
```

If yes, which one do we want to make an example of?

fabpot: More something along the lines of:

```yaml
        provider: '%env(TRANSLATION_PROVIDER_DSN)%'
```

welcoMattic: Yes, you can add many providers, for example to manage `admin` domain in Loco, and `messages` domain in Crowdin. That's why I added an example for each. But, I'm ok to put only one example under `providers:`

OskarStark: I would add 2 but not all, like we do in notifier, but @fabpot please decide, thanks

welcoMattic: I just commit the removal of 2 providers in the example. Is it ok for you to keep 2 providers here?

welcoMattic:
Following up to https://github.com/symfony/symfony/issues/41374, I removed the PoEditor recipe

welcoMattic:
@Nyholm @OskarStark As 5.3 final has been released today, we need to merge

fabpot:

fabpot (approved):

fabpot (commented):
merge now

PR #952, opened 2021-06-02
Fix recipe for phpunit

### LINKED ISSUE #936: in phpunit recipe 5.3 bin/phpunit fails with phpunit 9.5
Hi!

I'm experimenting with phpunit and just find that in the 5.3 recipe bin/phpunit tries to require vendor/phpunit/phpunit/phpunit.

https://github.com/symfony/recipes/blob/master/symfony/phpunit-bridge/5.3/bin/phpunit#L4

But that fails with:

> Fatal error: strict_types declaration must be the very first statement in the script in /var/www/html/vendor/phpunit/phpunit/phpunit on line 2

looks like importing a file which has booth declare() AND an interpreter (`#!/usr/bin/env `) cannot be imported.
might be a php bug idk

https://github.com/sebastianbergmann/phpunit/blob/9.5/phpunit#L1

tried with phpunit 9.5

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Fix #936
Fix https://github.com/symfony/symfony/issues/41409 and https://github.com/symfony/symfony/issues/41471


phpunit/phpunit/4.7/phpunit.xml.dist
```diff
@@ -25,12 +25,9 @@
         </whitelist>
     </filter>
 
-    <!-- Run `composer require symfony/phpunit-bridge` before enabling this extension -->
```
nicolas-grekas, line 28: This removes any diff between phpunit.xml.dist shipped by simple-phpunit vs phpunit recipes.
Without this change, what ppl get is random, leading to the deprecation report being unavailable.

symfony/phpunit-bridge/5.3/bin/phpunit
```diff
@@ -1,16 +1,19 @@
...
-    require dirname(__DIR__).'/vendor/phpunit/phpunit/phpunit';
+
+if (!ini_get('date.timezone')) {
+    ini_set('date.timezone', 'UTC');
+}
+
+if (is_file(dirname(__DIR__).'/vendor/phpunit/phpunit/phpunit')) {
+    define('PHPUNIT_COMPOSER_INSTALL', dirname(__DIR__).'/vendor/autoload.php');
+    require PHPUNIT_COMPOSER_INSTALL;
+    PHPUnit\TextUI\Command::main();
```
weaverryan, line 11: Is this the real significant portion of this PR? Could you explain a bit further?

nicolas-grekas: This is the only possible fix for #936

weaverryan: I see how there is no PHP file in phpunit to directly require. And so, you've mimicked their phpunit executable.

phpunit/phpunit/9.3/phpunit.xml.dist
```diff
@@ -0,0 +1,40 @@
...
+        <testsuite name="Project Test Suite">
+            <directory>tests</directory>
+        </testsuite>
+    </testsuites>
+
+    <coverage processUncoveredFiles="true">
+        <include>
+            <directory suffix=".php">src</directory>
+        </include>
+    </coverage>
```
weaverryan, line 28: Note to others: changes made to the schema for PHPUnit 9.

weaverryan (approved):

chalasr (approved):

fmonts:
Why `.phpunit` has been removed from .gitignore? I don't think that directory should be committed to source code, in my case it contains 2600+ files...

symfony/phpunit-bridge/5.3/manifest.json
```diff
@@ -6,7 +6,6 @@
         "tests/": "tests/"
     },
     "gitignore": [
-        ".phpunit",
```
fmonts, line 9: Why `.phpunit` has been removed from .gitignore? I don't think that directory should be committed to source code, in my case it contains 2600+ files...

Kocal: Because now we prefer using the binary from PHPUnit (which does not create `.phpunit` folder) instead the binary from Symfony PHPUnit bridge (`simple-phpunit`).

So adding `.phpunit` in `.gitignore` is not useful anymore.

symfony/phpunit-bridge/5.3/phpunit.xml.dist
```diff
@@ -13,11 +11,8 @@
         <ini name="error_reporting" value="-1" />
         <server name="APP_ENV" value="test" force="true" />
         <server name="SHELL_VERBOSITY" value="-1" />
-        <!-- If you use phpunit-bridge directly, you can control extra behavior -->
-        <!--
         <server name="SYMFONY_PHPUNIT_REMOVE" value="" />
-        <server name="SYMFONY_PHPUNIT_VERSION" value="8.5" />
-        -->
+        <server name="SYMFONY_PHPUNIT_VERSION" value="9.5" />
```
Tobion, line 15: Symfony 5.3 requires php 7.2. But phpunit 9.5 requires php 7.3. So PHP 7.2 users won't be able to run phpunit by default or am I misunderstanding this?

fmonts: PHP 7.2 is EOL so it shouldn't be maintained anymore IMHO

PR #955, opened 2021-06-03
Commit vendor/autoload_runtime.php

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

I propose to commit the `vendor/autoload_runtime.php` file: this will allow installing apps with composer when the `--no-scripts` or `--no-plugins` flags are passed.

### DIFF
--- a/symfony/framework-bundle/5.3/manifest.json
+++ b/symfony/framework-bundle/5.3/manifest.json
@@ -22,6 +22,7 @@
         "/%CONFIG_DIR%/secrets/prod/prod.decrypt.private.php",
         "/%PUBLIC_DIR%/bundles/",
         "/%VAR_DIR%/",
-        "/vendor/"
+        "/vendor/*",
+        "!vendor/autoload_runtime.php"
     ]
 }


weaverryan:
Sorry for that mystery close and reopen ;).

What I meant to say was: probably a minority of users run composer with this flag. Perhaps we should instead document that they should commit the file? Ignoring vendor/ has been repeated so many times over the last 10 years that it’s basically a rule that every dev knows by heart. I’d like to not add an exception to that rule (and make symfony an outlier).

nicolas-grekas:
All SymfonyCloud users were installing their app with this flag to skip installing assets.
Same for our tests, where now have to install assets because we had to remove the flag.
I don't think this is a minority of users, given that's also the first comment that came up on the blog:
https://symfony.com/blog/new-in-symfony-5-3-runtime-component

This will still exclude vendor. It's just this file.

nicolas-grekas:
`--no-scripts` is very commonly used to skip running the `scripts` of the composer file. The issue is that this flag also disables plugins. Composer provides no way to skip only scripts while still running plugins :'( /cc @seldaek btw, this might be a missing feat of composer.

fabpot:
I'm very much against this change. What about moving this file elsewhere instead?

nicolas-grekas:
Closing in favor of https://github.com/composer/composer/pull/9942

PR #962, opened 2021-07-07
Check for .gitkeep & .yml files

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Those are two checks that the flex-server already does.
I'm submitting to ask: should we move all checks to GHA?
Also,the flex-server should skip the `.github` directory :)

### DIFF
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -18,9 +18,39 @@ jobs:
         uses: actions/checkout@v2
 
       - name: No symlinks
+        if: always()
         run: |
-          SYMLINKS=$(find -type l)
+          SYMLINKS=$(find * -type l)
           if [[ "" != "$SYMLINKS" ]]; then
             echo "::error::Symlinks are not allowed"
             echo "Found $SYMLINKS"
+            echo
+            exit 1
           fi
+
+      - name: No .yml, use .yaml
+        if: always()
+        run: |
+          YMLS=$(find * -name '*.yml')
+          if [[ "" != "$YMLS" ]]; then
+            echo "::error::*.yml files should renamed to *.yaml"
+            echo "Found $YMLS"
+            echo
+            exit 1
+          fi
+
+      - name: No .gitkeep, use .gitignore
+        if: always()
+        run: |
+          GITKEEPS=$(find * -name .gitkeep)
+          if [[ "" != "$GITKEEPS" ]]; then
+            echo "::error::.gitkeep files should be renamed to .gitignore"
+            echo "Found $GITKEEPS"
+            echo
+            exit 1
+          fi
+
+      - name: 4 spaces indentation
+        if: always()
+        run: |
+          find * -name '*.yaml' | xargs -n1 grep -P -H -n -v '^((    )*[^ \t]|$)' | cut -d: -f1-2 | sed 's/\(.*\):\([0-9]*\)$/::error file=\1,line=\2::Indendation must be a multiple of 4 spaces/'


.github/workflows/qa.yml
```diff
@@ -23,4 +23,23 @@ jobs:
           if [[ "" != "$SYMLINKS" ]]; then
             echo "::error::Symlinks are not allowed"
             echo "Found $SYMLINKS"
+            exit 1
+          fi
+
+      - name: No .gitkeep, use .gitignore
```
stof, line 29: Currently, each step runs only if the previous one is a success, meaning that we don't report all errors to the contributor at once. there is 2 ways to check them all:

1. use separate jobs
2. change the conditions to run the steps. see https://github.com/symfony/demo/blob/7ed6c4a801eb074ad4052563cf54bf0e380de816/.github/workflows/lint.yaml#L64 for an example (this requires setting an id on the last step of the setup, to be able to reference it for the one that should have succeeded before running the linting step)

.github/workflows/qa.yml
```diff
@@ -23,4 +23,23 @@ jobs:
...
+          GITKEEPS=$(find -name .gitkeep)
+          if [[ "" != "$GITKEEPS" ]]; then
+            echo "::error::.gitkeep files should be renamed to .gitignore"
+            echo "Found $GITKEEPS"
+            exit 1
+          fi
+
+      - name: No .yml, use .yaml
+        run: |
+          YMLS=$(find -mindepth 2 -name '*.yml')
```
stof, line 40: This check for `*.yml` must skip the `.github` folder

stof (changes requested):

stof:
@nicolas-grekas why using this weird `FAILED=1` trick (that currently produces an invalid YAML file breaking things) that would hide which step actually failed, instead of using the native way to have some steps running even if a previous step failed (see my comment) ?

nicolas-grekas:
@stof thanks for the hints:
> use separate jobs

Each job would require checking our the code. What a waste of resources.

> change the conditions to run the steps. see https://github.com/symfony/demo/blob/7ed6c4a801eb074ad4052563cf54bf0e380de816/.github/workflows/lint.yaml#L64 for an example (this requires setting an id on the last step of the setup, to be able to reference it for the one that should have succeeded before running the linting step)

But I want all steps to run, no matter their out come. How would that work here?

stof:
> > change the conditions to run the steps. see https://github.com/symfony/demo/blob/7ed6c4a801eb074ad4052563cf54bf0e380de816/.github/workflows/lint.yaml#L64 for an example (this requires setting an id on the last step of the setup, to be able to reference it for the one that should have succeeded before running the linting step)
> 
> But I want all steps to run, no matter their out come. How would that work here?

That's precisely what this condition is about. The default behavior for steps is to always have an implicit `success()` condition for running (checking the current state of the job at that point, which would skip them from running if a previous step failed), *unless the custom condition contains a state checking function*. That's why the condition calls `always()`. The second part is to check the outcome of the last step of the *setup* part, because we only want to run the lining steps (all of them, which will all have the same condition) if the setup was successful (for now, you only have 1 setup step: checking out the code. But we still need that one to succeed to run lining steps)

stof:
See https://github.com/symfony/demo/runs/2726554120?check_suite_focus=true for an example of how this gets reported in symfony/demo

nicolas-grekas:
Should be good now, thanks for your help @stof

PR #964, opened 2021-07-07
Move checks to GHA

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

- [x] No symlinks
- [x] No .yml, use .yaml
- [x] No .gitkeep, use .gitignore
- [x] 4 spaces indentation
- [x] manifest.json is found
- [x] JSON files can be linted
- [x] YAML files can be linted
- [x] Pull request should not have merge commits (please rebase)
- [x] Contributions must be licensed under MIT (add the pull request header in the description)
- [x] Alias "%s" cannot be used as it's a special alias used by Composer
- [x] Alias "%s" also defined for "%s"
- [x] Aliases not supported in the contrib repository
- [x] Unsupported key "%s" in manifest.json file
- [x] Package does not exist on Packagist
- [x] Version "%s" does not exist on Packagist
- [x] You should register the bundle in the manifest as the package is a (Symfony|Sylius|Sulu) bundle
- [x] Package must not depend on symfony/symfony; depend on explicit symfony/* packages instead
- [x] Package must not depend on symfony/security; depend on explicit symfony/security-* packages instead
- [x] Version "%s" not valid, should probably be "%s.%s" instead
- [x] Version "%s" not valid, format is "x.y" where x and y are numbers
- [x] Use https://symfony.com when referencing the Symfony website
- [x] Files should end with a newline Should end with a newline for ".yaml", ".yml", ".txt", ".md", ".markdown", ".json", ".rst"
- [x] Symfony console commands should not be wrapped in a Makefile task
- [x] File must be under a directory defined in the "copy-from-recipe" section of "manifest.json"
- [x] Recipe is not needed as it only registers bundle "%s" for all environments
- [x] "parameters" should be defined via the "container" configurator instead in config/packages/
- [x] Underscore notation is required for (directory|file) names under config/
- [x] "~" should be replaced with "null" in yaml files
- [x] "%s" (nil|string|map|seq) should be removed as it is empty in config/packages/


ogizanagi:
We should update the [editorconfig](https://github.com/symfony/recipes/blob/master/.editorconfig) file regarding the YAML indentation:

```editorconfig
[*.{yaml,yml}]
indent_style = space
indent_size = 4
```

.github/workflows/qa.yml
```diff
@@ -1,26 +1,70 @@
...
+        runs-on: Ubuntu-20.04
+
+        steps:
+            -
+                name: Checkout
+                uses: actions/checkout@v2
+
+            -
+                name: No symlinks
+                if: always()
```
stof, line 23: this should be `if: "always() and steps.end_of_setup.outcome == 'success'"` (with an id `end_of_setup` set on the `actions/checkout` step).
We don't want to run those linting steps if checking out the code failed, as that depends on it.

stof (changes requested):

.github/workflows/qa.yml
nicolas-grekas: updated

.github/workflows/qa.yml
```diff
@@ -1,26 +1,113 @@
...
+        steps:
+            -
+                name: Checkout
+                uses: actions/checkout@v2
+                id: checkout
+
+            -
+                name: Install tools
+                run: |
+                    (cd .github/tools && composer install --no-dev)
```
stof, line 25: This tool relies on PHP, so you should use the `setup-php` action

.github/workflows/qa.yml
```diff
@@ -1,26 +1,113 @@
...
+
+                    if [[ "" != "$ERRORS" ]]; then
+                        echo -e $ERRORS
+                        echo
+                        exit 1
+                    fi
+
+            -
+                name: YAML files are valid
+                if: "always() && steps.checkout.outcome == 'success'"
```
stof, line 105: this should check the state of the install of the tools, not of the checkout, as it relies on this setup step too

stof:
@fabpot could you update the symfony-flex-server to ignore the `.github` folder when looking for recipe folders ?

.github/workflows/qa.yml
nicolas-grekas: there is no need to, all works well already

.github/workflows/qa.yml
nicolas-grekas: Not sure we care: the install step is not expected to fail, and if it does, this one will too.

nicolas-grekas:
All checks are now implemented!

nicolas-grekas:
The code for the custom commands moved to https://github.com/symfony-tools/recipes-checker

PR #965, opened 2021-07-09
Compute diff between recipe versions

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

### DIFF
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -1,7 +1,7 @@
 name: Q&A
 
 on:
-    pull_request:
+    pull_request_target:
 
 defaults:
     run:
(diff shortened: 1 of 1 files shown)


.github/workflows/qa.yml
```diff
@@ -1,7 +1,7 @@
 name: Q&A
 
 on:
-    pull_request:
+    pull_request_target:
```
stof, line 4: This makes the workflow checkout the `master` branch rather than the code of the PR, making the CI useless (as it does not test the PR)

nicolas-grekas: Oops, fixed in 8269d3d

PR #971, opened 2021-07-11
Rely on defaults for doctrine mapping types

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Needs https://github.com/symfony/symfony/pull/42054 to be released before merging.

### DIFF
--- a/doctrine/doctrine-bundle/2.4/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/2.4/config/packages/doctrine.yaml
@@ -12,7 +12,6 @@ doctrine:
         mappings:
             App:
                 is_bundle: false
-                type: annotation
                 dir: '%kernel.project_dir%/src/Entity'
                 prefix: 'App\Entity'
                 alias: App


jrushlow:
We're wrapping up attribute support in MakerBundle now in https://github.com/symfony/maker-bundle/pull/920 We should have this merged and released pretty quick.

fabpot:
@jrushlow Correct if I'm wrong, but this change is not a requirement. I understand that it would be better, but as I said, I'm afraid that it will break a lot of things. I know for instance that the Symfony book will break instanenaously.
I would recommend waiting for 5.4 to be released so that we can have a version of the book with support for the new maker bundle.

weaverryan:
> I would recommend waiting for 5.4 to be released so that we can have a version of the book with support for the new maker bundle.

That makes sense to me. Indeed, this recipe is not a blocker for MakerBundle. Once it *is* merged, we'll just need to update some MakerBundle tests - no big deal :).

weaverryan:
Is it safe to merge this PR now? Without it, if you start a brand new Symfony 6 project / PHP 8, etc, you still get annotation-powered entities instead of attributes.

chalasr (approved):

derrabus (approved):

Tobion (approved):

PR #973, opened 2021-07-11
Also ignore secret private key by default for non-prod environments

Tobion:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | 

To me every private key should be excluded by default independent of the environment.
Esp. if you use other deployment environments like staging, having this unsecure default as gitignore is kinda dangerous.
The fact that custom env don't have the private key ignored by default and https://symfony.com/doc/current/configuration.html#creating-a-new-environment does not mention it that this should be done manually is a potential security problem.

If the secrets from dev or test are not really secrets (which is why you could commit the private key), then there is not point in setting them as secrets in the first place. Those can just be normal env vars.

For the test env this is already recommended in https://symfony.com/doc/current/configuration/secrets.html#secrets-in-the-test-environment
> If you add a secret in the dev and prod environments, it will be missing from the test environment. You could create a “vault” for the test environment and define the secrets there. But an easier way is to set the test values via the .env.test file.

For dev the recommendation seems different in https://symfony.com/doc/current/configuration/secrets.html#create-or-update-secrets
which does not make sense to me. Whether something needs to be encrypted does not depend on the environment, it depends on the value. If the dev environment contains the database_url for a local docker mysql server, that does not need to be encrypted. But if you configure a real, online smtp server to be used in the dev environment, than the mailer_dsn should be encrypted.

Reopened #972

### DIFF
--- a/symfony/framework-bundle/4.4/manifest.json
+++ b/symfony/framework-bundle/4.4/manifest.json
@@ -21,7 +21,7 @@
         "/.env.local",
         "/.env.local.php",
         "/.env.*.local",
-        "/%CONFIG_DIR%/secrets/prod/prod.decrypt.private.php",
+        "/%CONFIG_DIR%/secrets/*/*.decrypt.private.php",
         "/%PUBLIC_DIR%/bundles/",
         "/%VAR_DIR%/",
         "/vendor/"
--- a/symfony/framework-bundle/5.1/manifest.json
+++ b/symfony/framework-bundle/5.1/manifest.json
@@ -21,7 +21,7 @@
         "/.env.local",
         "/.env.local.php",
         "/.env.*.local",
-        "/%CONFIG_DIR%/secrets/prod/prod.decrypt.private.php",
+        "/%CONFIG_DIR%/secrets/*/*.decrypt.private.php",
         "/%PUBLIC_DIR%/bundles/",
         "/%VAR_DIR%/",
         "/vendor/"
--- a/symfony/framework-bundle/5.2/manifest.json
+++ b/symfony/framework-bundle/5.2/manifest.json
@@ -19,7 +19,7 @@
         "/.env.local",
         "/.env.local.php",
         "/.env.*.local",
-        "/%CONFIG_DIR%/secrets/prod/prod.decrypt.private.php",
+        "/%CONFIG_DIR%/secrets/*/*.decrypt.private.php",
         "/%PUBLIC_DIR%/bundles/",
         "/%VAR_DIR%/",
         "/vendor/"
--- a/symfony/framework-bundle/5.3/manifest.json
+++ b/symfony/framework-bundle/5.3/manifest.json
@@ -19,7 +19,7 @@
         "/.env.local",
         "/.env.local.php",
         "/.env.*.local",
-        "/%CONFIG_DIR%/secrets/prod/prod.decrypt.private.php",
+        "/%CONFIG_DIR%/secrets/*/*.decrypt.private.php",
         "/%PUBLIC_DIR%/bundles/",
         "/%VAR_DIR%/",
         "/vendor/"


nicolas-grekas (changes requested):
Adding a preemptive review to block auto-merging 😅

nicolas-grekas:
If we do this, it must be done for 5.4 only. That's because merging this on lower recipes will break the book:
https://symfony.com/doc/current/the-fast-track/en/16-spam.html#storing-secrets

> But an easier way is to set the test values via the .env.test file.

Then we have an inconsistency between the book and the doc, see:
https://symfony.com/doc/current/the-fast-track/en/17-tests.html#configuring-the-test-environment

I'm personally in favor of encouraging one way to manipulate secrets - aka through vaults. By telling  ppl to use the vault by default, we're going to teach then how it works via practicing. I think this is valuable for a smoother DX. /cc @weaverryan we should fix the doc if you agree.

There is another issue with the proposed change: it doesn't give any clue about a way to deal with dev/test keys. For prod, we do read the `SYMFONY_DECRYPTION_SECRET` env var by default, and we document how to handle it. How would this look like for dev/test? I've no idea.

I'm thus :-1: here: doing this as is would basically mean that we degrade the DX around dev/test vaults, by by making them unusable in practice. I don't see the value this would bring.

If ppl prefer to use .env.dev for storing secrets, nothing prevents them from doing so. And vaults don't prevent that.

Tobion:
> That's because merging this on lower recipes will break the book: https://symfony.com/doc/current/the-fast-track/en/16-spam.html#storing-secrets

I don't see why it would invalidate the book. It says
> For development secrets, you can decide to commit the vault and the keys that have been generated in the config/secret/dev/ directory.

That's totally correct. Depending on what values you encrypt and the context, you can DECIDE to commit the key.

> I'm personally in favor of encouraging one way to manipulate secrets - aka through vaults. 

A value that does not have sensitive information is not a secret. And an encrypted value whose key is committed alongside it, is also not a secret. So you are right, SECRETS should always go through the vault. But pretending that a value is a secret and at the same time committing the private key is totally wrong. Then it's not a secret anymore and there is no point in encrypting it.

> There is another issue with the proposed change: it doesn't give any clue about a way to deal with dev/test keys. For prod, we do read the SYMFONY_DECRYPTION_SECRET env var by default, and we document how to handle it. How would this look like for dev/test? I've no idea.

I don't really understand the question.
If dev/test keys are sensitive, encrypt them but keep the private key outside the repo (e.g. in bitwarden or what ever secret storage you use at the compony). This is the same thing that applies to the prod keys. You also need to store them somewhere.
If the dev/test keys are not sensitive, then put them in .env.test/.env.dev.

weaverryan:
> So you are right, SECRETS should always go through the vault. But pretending that a value is a secret and at the same time committing the private key is totally wrong

This is logical, I just don't see it as a big deal.

Overall, I'm seeing two very similar workflows and neither seem clearly superior. That's why I'm inclined to make no change: change is cost. I feel like we need more "win" to make this change.

I WILL say that, even though I was heavily involved in the original workflow for secrets, I find both the current AND alternative workflows proposed here as kind of a pain in the butt. In both cases, to "convert" something from a normal env var to a secret, I need to (A) completely REMOVE the env var from `.env`, (B) put it into the `prod` vault with `APP_ENV bin/console secrets:set` and also (C) either also put it into the `dev` + `test` vaults PR put it into `.env.dev` and `.env.test`.

If we're going to tweak the system, it should make that workflow easier if possible. For example, having only one vault (so not needing to prefix `secrets:set` with `APP_ENV=prod`) that's only activated in the `prod` environment (probably with a `when@prod` config in `framework.yaml`) AND where the secret WINS over "real" `.env` values so that you don't need to move an env var from `.env` to `.env.dev` and `.env.test`.

Tobion:
> This is logical, I just don't see it as a big deal. Overall, I'm seeing two very similar workflows and neither seem clearly superior.

To me it should be pretty clear that teaching bad security concepts is not the way to go. How would you explain someone that he needs to encrypt a value that is neither sensitive e.g. http://localhost:5432 for local development database? Ah and btw after you have encrypted it's not even secure in the end because we expose the private decryption key. 
Sorry but I just cannot understand this reasoning at all.

chalasr:
Not git-ignoring the private key for all environments seems wrong to me, and I agree it may have bad have consequences overall. Same reasoning about always using vaults: if a value is not really a secret, it should not go through vaults. 
This may complicates things, and we may need to find a  way to improve DX for the dev-to-prod workflow but breaking existing docs should not be a blocker for improving our defaults on something as critical as secret management. 
I'm 👍 for this change, at least as a "feature".

boedah:
After creating a `qa` env, I was also astonished as to why Symfony wants me to check in that new private key as it is not ignored. I changed the line to ignore all private decryption keys back then (after adding the one for `dev`).
IMHO the default should be to not accidentally check-in private keys - esp. for new users. There are SaaS tools to scan for checked-in secrets for a reason.

An option would be to give the user a hint in the output of the command after creating a `dev`/`test` secret (or always?) that she can force add that key via `git add -f config/secrets/dev/dev.decrypt.private.php` with an explanation of the implications.

nicolas-grekas:
I'm going due to lack of consensus.

PR #976, opened 2021-07-21
Set 'lang' attribute in 'html' node

yoannrenard:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | N/A

What do you think guys about setting the html `lang` attribute straight with the current locale?
IMO that would help us to make websites with better web accessibility.


Tobion:
If the website doesn't translate it's content this is not correct. Also developers might want to use Content-Language header instead. So I don't think we should put such assumptions as default.

yoannrenard:
I see what you mean, I'm closing the PR then
thanks @Tobion

PR #979, opened 2021-08-02
Link best practice doc part to secret management

noniagriconomie:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | 

Small addition, I find it usefull here

### DIFF
--- a/symfony/flex/1.0/.env
+++ b/symfony/flex/1.0/.env
@@ -9,6 +9,7 @@
 # Real environment variables win over .env files.
 #
 # DO NOT DEFINE PRODUCTION SECRETS IN THIS FILE NOR IN ANY OTHER COMMITTED FILES.
+# https://symfony.com/doc/current/best_practices.html#use-secret-for-sensitive-information
 #
 # Run "composer dump-env prod" to compile .env files for production use (requires symfony/flex >=1.2).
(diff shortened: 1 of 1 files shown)


symfony/flex/1.0/.env
```diff
@@ -9,6 +9,7 @@
 # Real environment variables win over .env files.
 #
 # DO NOT DEFINE PRODUCTION SECRETS IN THIS FILE NOR IN ANY OTHER COMMITTED FILES.
+# https://symfony.com/doc/current/best_practices.html#use-secret-for-sensitive-information
```
Nyholm, line 12: What about linking here directly?

https://symfony.com/doc/current/configuration/secrets.html

noniagriconomie: hi @Nyholm 
I wanted to point to the same page as the like below
But we can define the link that is more relevant for you :)

noniagriconomie: friendly ping @Nyholm

94noni: I’ll do it and swap and finish this PR
Thx for feed-back

noniagriconomie: done in https://github.com/symfony/recipes/pull/1074
closing here

PR #984, opened 2021-08-18
Use static app callable for the new Runtime component

SerheyDolgushev:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | N/A

### DIFF
--- a/symfony/console/5.3/bin/console
+++ b/symfony/console/5.3/bin/console
@@ -10,7 +10,7 @@ if (!is_file(dirname(__DIR__).'/vendor/autoload_runtime.php')) {
 
 require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
 
-return function (array $context) {
+return static function (array $context) {
     $kernel = new Kernel($context['APP_ENV'], (bool) $context['APP_DEBUG']);
 
     return new Application($kernel);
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
They are already static because they're not declared in a method.

SerheyDolgushev:
Sorry, but the anonymous function should use `static` keyword to be static. If the function is not declared as a method it does not make it static: https://www.php.net/manual/en/functions.anonymous.php#functions.anonymous-functions.static

nicolas-grekas:
yes it does, see eg https://3v4l.org/FT3gl

nicolas-grekas:
I suppose you mean static as in not allowing rebinding: https://3v4l.org/5fQMn
This makes zero difference here as there is no `$this`.
Let's save some boilerplate.

Nyholm:
Thank you for this @SerheyDolgushev 

Does this PR has an impact in performance, usability or any other way? Or is it just to please some good practices?

SerheyDolgushev:
@Nyholm 
Actually, this OR impact on performance can be neglected. Also, it has no effect on usability. 
In this case, we have anonymous functions that are not expecting to have any instance to be bind to it. And according to https://www.php.net/manual/en/functions.anonymous.php#functions.anonymous-functions.static such kind functions should use `static` keyword. So personally I think this PR makes the code a bit more cleaner.

Nyholm:
Since this PR is only about code style (subjective opinion) I prefer to keep things the way it is. Simply because it is less code and my subjective opinion believe the current code is more clean. 

Thank you for the references.

nicolas-grekas:
Resubmitted as #984, with benchmarks this time!

PR #986, opened 2021-08-20
Add scheb/2fa-bundle to official recipes

wouterj:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

I propose to move @scheb's 2fa bundle from the contrib repository to the official recipes:

- This bundle provides multi factor authentication, a feature which we likely won't ever cover fully in Symfony core and rather rely on this bundle.
- The bundle is stable, works great and @scheb is doing a great job of keeping up to date with Symfony versions (and even giving feedback on new security changes that affect the bundle).

For discoverability, I also propose to give the bundle the "2fa" alias.


fabpot (approved):

chalasr (approved):

scheb:
Thank you, feeling honored to have it in the club of noble bundles 🤴

PR #989, opened 2021-08-28
Add local Docker config for Mailer

fabpot:
When using the Mailer locally, let's use a mail catcher automatically.

/cc @dunglas 

| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

### DIFF
--- a/symfony/mailer/4.3/manifest.json
+++ b/symfony/mailer/4.3/manifest.json
@@ -5,5 +5,14 @@
     "env": {
         "#1": "MAILER_DSN=smtp://localhost"
     },
+    "docker-compose": {
+        "docker-compose.override.yml": {
+            "services": [
+                "mailer:",
+                "  image: schickling/mailcatcher",
+                "  ports: [1025, 1080]"
+            ]
+        }
+    },
     "aliases": ["mailer", "mail"]
 }


dunglas (approved):

nicolas-grekas (approved):

symfony/mailer/4.3/manifest.json
```diff
@@ -5,5 +5,14 @@
     "env": {
         "#1": "MAILER_DSN=smtp://localhost"
     },
+    "docker-compose": {
+        "docker-compose.override.yml": {
+            "services": [
```
OskarStark, line 10: It would be nice to move this to a separate file to instead of writing (maybe more complex) services in json 

This is more a feature request/idea for flex.

WDYT?

fabpot: We discuss this in the past, and IIRC, we did this as we already had all the infrastructure to update files with data from the manifest. We might want to revisit this indeed as doing it in a JSON file is not optimal to say the least :)

OskarStark: Thanks for the feedback 👌🏻

PR #991, opened 2021-08-30
Remove the header from post-install files (handled by Flex soon)

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

When installing a pack or when installing multiple packages at once, you might have a confusing output at the end of the `composer` run as many post install file might be displayed as most of the start with a blue "What's next" header.

In a Flex PR, I will propose to move the header to Flex and Flex will also add a small header between each post install to associate each post install file to its package. See https://github.com/symfony/flex/pull/790

I'm opening this PR so that I can test things and see how I can improve the output by making some experiments.

So, please NO MERGE.

### DIFF
--- a/api-platform/admin-pack/1.0/post-install.txt
+++ b/api-platform/admin-pack/1.0/post-install.txt
@@ -1,7 +1,3 @@
-<bg=blue;fg=white>                       </>
-<bg=blue;fg=white> To finish the install </>
-<bg=blue;fg=white>                       </>
-
   * Your admin is almost ready:
     1. Install the JavaScript dependencies by running <comment>yarn add @api-platform/admin @babel/preset-react</comment>
     2. Edit <comment>webpack.config.js</comment> and <info>uncomment</info> the following lines:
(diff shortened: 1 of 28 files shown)


fabpot:
Let me paste the new output here to get some feedback:

image

As you can see, the generic `Some files have been created` sentence is now under the `What's next` header, which is better IMO.

fabpot:
And here is the current output for reference:

image

javiereguiluz:
This proposal is fantastic! The new output is so much better than the current one 🎉 

Two minor comments:

* (Not related to this change but -->) Some packages show messages like *"You are ready to send emails"*, *"You're ready to use the Messenger component"*, etc. They might look nice when installing just one package, but with this multi-package list, they read a bit redundant/unnecessary.
* The only minor design thing that I'm not sure about is the yellow background to the entire `Instructions for Package XXXX` message. To me it's a bit hard to quickly spot the different package names. But it might be just me because this is the first time I see this.

OskarStark:
Nice idea 💡

Tobion:
👍

lyrixx (approved):

PR #995, opened 2021-08-31
[platformsh] Add a cron that automatically check for security issues in prod

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfonycorp/platformsh-meta/1.0/.platform.app.yaml
+++ b/symfonycorp/platformsh-meta/1.0/.platform.app.yaml
@@ -48,6 +48,16 @@ hooks:
 
         (>&2 symfony-deploy)
 
+crons:
+    security-check:
+        # Check that no security issues have been found for PHP packages deployed in production
+        # See https://github.com/fabpot/local-php-security-checker
+        spec: '50 23 * * *'
+        cmd: |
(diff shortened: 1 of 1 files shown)


symfonycorp/platformsh-meta/1.0/.platform.app.yaml
```diff
@@ -48,6 +48,16 @@ hooks:
...
         (>&2 symfony-deploy)
 
+crons:
+    security-check:
+        # Check that no security issues have been found for PHP packages deployed in production
+        # See https://github.com/fabpot/local-php-security-checker
+        spec: '50 23 * * *'
+        cmd: |
+            if [ "$SYMFONY_BRANCH" = "master" ]; then
+                croncape php-security-checker
```
OskarStark, line 58: Will it result in an email if it detects security issues?

fabpot: croncape will send an mail indeed is the exit code is not 0, which is exactly what we want

fabpot: Don't try it at home yet as I need to make a new release of the tools installed in containers (will probably do it later today).

PR #1001, opened 2021-09-19
Add recipe for runtime/bref

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | https://github.com/brefphp/bref/pull/1032

https://github.com/php-runtime/bref


runtime/bref/0.2/serverless.yaml
```diff
@@ -0,0 +1,50 @@
...
+        - '!assets/**'
+        - '!node_modules/**'
+        - '!public/build/**'
+        - '!tests/**'
+        - '!var/**'
+        # If you want to include files and folders that are part of excluded folders,
+        # add them at the end
+        - 'var/cache/prod/**'
+        - 'public/build/entrypoints.json'
+        - 'public/build/manifest.json'
```
fabpot, line 50: I needed to add `        - '.env*'` to make env vars work properly (at least with the latest serverless version).

runtime/bref/0.2/serverless.yaml
```diff
@@ -0,0 +1,50 @@
+# Read the documentation at https://www.serverless.com/framework/docs/providers/aws/guide/serverless.yml/
+service: my-symfony-app
+configValidationMode: error
+
+provider:
+    name: aws
+    # The AWS region in which to deploy (us-east-1 is the default)
+    region: us-east-1
+    # The stage of the application, e.g. dev, production, staging… ('dev' is the default)
+    stage: dev
```
fabpot, line 10: I would named it prod as APP_ENV below is set to prod.

runtime/bref/0.2/serverless.yaml
```diff
@@ -0,0 +1,50 @@
...
+    runtime: provided.al2
+    environment:
+        # Symfony environment variables
+        APP_ENV: prod
+        APP_RUNTIME: Runtime\Bref\Runtime
+        BREF_LOOP_MAX: 1 # Increase to keep the Lambda process alive between requests
```
fabpot, line 16: What about adding (commented): `SYMFONY_DECRYPTION_SECRET: ${ssm:/stripe-notifications/prod-decrypt-private-key}`

Nyholm: Great idea

runtime/bref/0.2/serverless.yaml
Nyholm: Oh. I've not updated to serverless 3.0 yet. Thank you

t-richard: I think the `.env` change in v3.0 you're referring to will soon be reverted https://github.com/serverless/serverless/issues/8566#issuecomment-921809620

runtime/bref/0.2/serverless.yaml
t-richard: Mentionning `stripe` here seems a bit odd to me. I would go with something more generic like `${ssm:/your-app/prod-decrypt-private-key}`

runtime/bref/0.2/serverless.yaml
t-richard: In my mind, Serverless stages and Symfony's `APP_ENV` are different. I'm often running several stages (dev/staging/prod) with `APP_ENV=prod`. But as `dev` is the default in Serverless framework, maybe this line could be totally removed ?

runtime/bref/0.2/serverless.yaml
```diff
@@ -0,0 +1,52 @@
...
+    region: us-east-1
+    # The stage of the application, e.g. dev, production, staging… ('dev' is the default)
+    stage: prod
+    runtime: provided.al2
+    environment:
+        # Symfony environment variables
+        APP_ENV: prod
+        APP_RUNTIME: Runtime\Bref\Runtime
+        BREF_LOOP_MAX: 1 # Increase to keep the Lambda process alive between requests
+        # SYMFONY_DECRYPTION_SECRET: ${ssm:/stripe-notifications/prod-decrypt-private-key}
```
fabpot, line 17:
```suggestion
        # SYMFONY_DECRYPTION_SECRET: ${ssm:/my-symfony-app/prod-decrypt-private-key}
```

to be consistent with the name of the service

fabpot (changes requested):

runtime/bref/0.2/bootstrap
```diff
@@ -0,0 +1,18 @@
...
+set -e
+
+LAMBDA_ARGV=(${_HANDLER//:/ })
+
+while true
+do
+    # We redirect stderr to stdout so that everything
+    # written on the output ends up in Cloudwatch automatically
+    /opt/bin/php "${LAMBDA_ARGV[0]}" 2>&1
+done
```
fabpot, line 18: missing EOL

Nyholm:
Oops, sorry, Updated now.

runtime/bref/0.2/serverless.yaml
fabpot: That's great news let's remove this and add `useDotenv: true` at the root instead.

fabpot (changes requested):

fabpot (approved):

runtime/bref/0.2/serverless.yaml
```diff
@@ -0,0 +1,52 @@
+# Read the documentation at https://www.serverless.com/framework/docs/providers/aws/guide/serverless.yml/
+service: my-symfony-app
+configValidationMode: error
+useDotenv: true
```
mnapoli, line 4: Unless I'm missing something you may not want to use `useDotenv: true`.

This will inject all `.env` env variables (application variables) into the `serverless` process (build/deploy process). And it will also automatically exclude `.env` files from being uploaded.

This feature was supposed to be enabled by default in Serverless Framework v3, but we've decided to keep it opt-in to avoid messing up applications. At the moment the error message says that this line should be added, but this will be changed soon.

If you want to control `.env` files separately from `serverless deploy` then you don't want to mix the 2.

runtime/bref/0.2/serverless.yaml
```diff
@@ -0,0 +1,52 @@
...
+        timeout: 120 # in seconds
+        layers:
+            - ${bref:layer.php-80}
+
+package:
+    patterns:
+        # Excluded files and folders for deployment
+        - '!assets/**'
+        - '!node_modules/**'
+        - '!public/build/**'
```
Kocal, line 45: I don't think we should exclude the whole `public/build` directory (even if you add `entrypoints.json` and `manifest.json` files back). 
What happens to files bundled by webpack Encore? They won't be deployed with this configuration right?

runtime/bref/0.2/serverless.yaml
```diff
@@ -0,0 +1,52 @@
...
+provider:
+    name: aws
+    # The AWS region in which to deploy (us-east-1 is the default)
+    region: us-east-1
+    # The stage of the application, e.g. dev, production, staging… ('dev' is the default)
+    stage: prod
+    runtime: provided.al2
+    environment:
+        # Symfony environment variables
+        APP_ENV: prod
```
Kocal, line 15:
```suggestion
    stage: ${opt:stage,'prod'} 
    runtime: provided.al2
    environment:
        # Symfony environment variables
        APP_ENV: ${self:provider.stage}
```

By doing this, the user can easily deploy a prod/env Symfony app by specifying `--stage` when deploying.

At work, we used to work on multiples stages/Symfony environments:
- a "staging" environement: we automatically deploy branch `develop` or pull requests with `--stage dev`, which will run a Symfony app in `dev` environment 
- a "production" environment: we automatically deploy branch `main `with `--stage prod`, which will run our Symfony app in `prod` environment

What do you think?

t-richard: This is not needed if you use the variable `${sls:stage}` which looks first for the cli option first, then the provider.stage value then defaults to dev if none is found.

runtime/bref/0.2/serverless.yaml
t-richard: These assets should be served through s3 or Cloudfront. Serving them from lambda is highly inefficient in terms of performance or costs. See https://bref.sh/docs/frameworks/symfony.html#assets and https://bref.sh/docs/websites.html

runtime/bref/0.2/serverless.yaml
mnapoli: +1 this is not needed, you should consider `stage` like a `defaultStage` configuration (we are considering renaming it to that in the future).

Also you might not want to mix Symfony env and stages, as you can have many stages (e.g. one per developer) which would not map to Symfony environments.

t-richard: Definitely agree here, those are two different concepts that I would not mix to avoid confusion.

Also see my previous comment here about that https://github.com/symfony/recipes/pull/1001#discussion_r711788971

t-richard: Definitely agree here, those are two different concepts that I would not mix to avoid confusion.

Also see my previous comment here about this https://github.com/symfony/recipes/pull/1001#discussion_r711788971

runtime/bref/0.2/serverless.yaml
Kocal: Oh good point, thx!

runtime/bref/0.2/serverless.yaml
Kocal: Alright, didn't know about that, thanks!

nicolas-grekas:
Friendly ping @Nyholm, just in case you forgot about it :)

fabpot:
Reading https://github.com/symfony/recipes-contrib/pull/1283, it looks like we should close both.
Closing now, feel free to reopen if you changed your mind.

PR #1005, opened 2021-09-23
[FrameworkBundle] Leverage vault to store APP_SECRET env var

noniagriconomie:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | Fix https://github.com/symfony/symfony/issues/38021 (https://github.com/symfony/symfony/issues/38021#issuecomment-884556420)

Hi, I've found this issue and wanted to give it a try for next v5.4
ℹ️ not tested for now, just code PR

Commands taken from https://symfony.com/doc/current/configuration/secrets.html
I also used the `--random` flag instead of the `generate()` feature to get a secret value


symfony/framework-bundle/5.4/manifest.json
```diff
@@ -9,11 +9,13 @@
     },
     "composer-scripts": {
         "cache:clear": "symfony-cmd",
-        "assets:install %PUBLIC_DIR%": "symfony-cmd"
+        "assets:install %PUBLIC_DIR%": "symfony-cmd",
+        "secrets:generate-keys": "symfony-cmd",
```
Tobion, line 13: the composer scripts run on each install/update. you don't want to regenerate keys every time

symfony/framework-bundle/5.4/manifest.json
```diff
@@ -9,11 +9,13 @@
...
+        "secrets:set APP_SECRET --random": "symfony-cmd",
+        "secrets:decrypt-to-local --force": "symfony-cmd"
```
Tobion, line 15: this should only be done for performance on production. it's nothing you'd always want to run.

Tobion (changes requested):

symfony/framework-bundle/5.4/manifest.json
noniagriconomie: accoding to the command doc:

```
  If encryption keys already exist, the command must be called with
  the --rotate option in order to override those keys and re-encrypt
  existing secrets.
  
      bin/console secrets:generate-keys --rotate
```

iiuc it is ok like this, after the first install, other updates will change nothing

symfony/framework-bundle/5.4/manifest.json
noniagriconomie: oki lets remove it

symfony/framework-bundle/5.4/manifest.json
```diff
@@ -9,11 +9,12 @@
     },
     "composer-scripts": {
         "cache:clear": "symfony-cmd",
-        "assets:install %PUBLIC_DIR%": "symfony-cmd"
+        "assets:install %PUBLIC_DIR%": "symfony-cmd",
+        "secrets:generate-keys": "symfony-cmd",
+        "secrets:set APP_SECRET --random": "symfony-cmd"
```
noniagriconomie, line 14: I do not know if I should use the `--local` option here

Tobion: doesn't this generate a new APP_SECRET every time?

noniagriconomie: indeed it does :s

2 options:

- either "do not care in dev" that this env is regenerated
- or we should add an option to this command so that, if already existing in the vault, it is not overriden if no flag provided (cf the `secrets:generate-keys --rotate`)

noniagriconomie: @nicolas-grekas friendly ping :)
related to https://github.com/symfony/symfony/issues/38021#issuecomment-685712783 and the creation of the secret logic, do you have an idea here?
many thanks!

noniagriconomie: PR at https://github.com/symfony/symfony/pull/43278 (personal account)

PR #1008, opened 2021-10-12
[docker-compose] add a pack to generate docker-compose files

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | todo

Add a new recipe generating the Docker Compose files needed by the Docker Compose configurator of Symfony flex.
We'll have to create the corresponding Git repo and composer package.

Usage:

```
symfony new my-project
cd my-project
symfony composer config --json extra.symfony.docker 'true'
symfony composer require docker-compose
docker-compose up -d
symfony serve -d
```

I also renamed the DB volume for consistency with the conventions we use elsewhere.

### DIFF
--- a/symfony/docker-compose-meta/1.0/docker-compose.override.yml
+++ b/symfony/docker-compose-meta/1.0/docker-compose.override.yml
@@ -1,4 +1,3 @@
 version: "3.8"
 
 # Development environment override
-services:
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
We don't attach recipes to packs anymore. We should use a metapackage instead. See eg symfonycorp/platformsh-meta

PR #1010, opened 2021-10-20
Add the webapp pack

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT

I'd like to deprecate the `website-skeleton` in favor of a new `webapp` pack.

The recommended pack is opinionated as it promotes the following:

* Doctrine ORM
* PostgreSQL as a default transport for Messenger (default recipes already makes PostgreSQL the default database choice)
* Symfony Webpack Encore / Symfony UX

We might add additional opinionated choices over time.

The idea is to get something that works out of the box for most applications.


fabpot:
I would not use `starter`, but I do understand where you're coming from. We could also have another pack for CLI command projects for instance, but that would only be about 0 more dep (console is already installed by default), so not really needed (the same goes for API).
So, it looks like this is the "only" case where we would need more is web applications, which is the core of what people build with the full-stack framework, hence "recommended".

fabpot:
PR updated accordingly.

PR #1018, opened 2021-11-05
Update spacing of doctrine yml files

alexander-schranz:
| Q             | A
| ------------- | ---
| License       | MIT

Use symfony coding standard of yml files with 4 spaces.

### DIFF
--- a/doctrine/doctrine-bundle/1.12/manifest.json
+++ b/doctrine/doctrine-bundle/1.12/manifest.json
@@ -24,24 +24,24 @@
         "docker-compose.yml": {
             "services": [
                 "database:",
-                "  image: postgres:${POSTGRES_VERSION:-13}-alpine",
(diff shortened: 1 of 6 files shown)


nicolas-grekas:
That could be worth a new check in the qa action. Would you like having a look?

alexander-schranz:
@nicolas-grekas how can we achieve this. Think the `lint:yaml` of symfony does not care about the yaml style? Are there any other tools used in symfony todo this?

alexander-schranz:
@fabpot do we need to change this offset value? https://github.com/symfony/flex/pull/788#discussion_r697849436

OskarStark:
Yamllint binary can check

nicolas-grekas:
Reverting as this 2-spaces indent was on purpose, docker-compose rulez.

PR #1019, opened 2021-11-05
[DoctrineBundle] Re-add the doctrine.orm.metadata_cache_driver option in production

javiereguiluz:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

This was removed in #996 but apparently Doctrine has un-reverted this removal. See https://github.com/doctrine/DoctrineBundle/pull/1405

### DIFF
--- a/doctrine/doctrine-bundle/2.4/config/packages/prod/doctrine.yaml
+++ b/doctrine/doctrine-bundle/2.4/config/packages/prod/doctrine.yaml
@@ -1,6 +1,9 @@
 doctrine:
     orm:
         auto_generate_proxy_classes: false
+        metadata_cache_driver:
+            type: pool
+            pool: doctrine.system_cache_pool
         query_cache_driver:
             type: pool
             pool: doctrine.system_cache_pool


stof:
DoctrineBundle has reverted the deprecation, to account for projects having special needs. But the recommendation is still to use the default behavior, so I would not re-add it in the recipe

javiereguiluz:
@stof OK, let's close this one then.

Just asking: which would be some of the criteria to maintain this in an app? I've re-added it to the Symfony Demo app, but I'm not sure if that is OK or not. Thanks!

stof:
I would also remove it from the demo. The existing default covers the 80% cases (and maybe even the 99% case).

The use case for keeping the metadata cache configurable was for projects that were so big that the default dev behavior was too slow for them, and they wanted to configure a prod-like cache even in the dev environment (requiring a manual cache clearing when changing the doctrine metadata, but this was the right trade-off from them as they were not changing them frequently and they would prefer the performance gained from the stronger caching)

javiereguiluz:
@stof thanks for the explanation. It's more clear now. I've removed this from the Symfony Demo app too.

Philosoft:
@stof what's the point to "recalculate" metadata for entities in prod environment on every request? and what about this https://github.com/doctrine/DoctrineBundle/issues/1393#issuecomment-938438463 in a thread that you refer to?

runephilosof-abtion:
@Philosoft As I understand it, the metadata caching now has sensible defaults in prod.
>  PHP Array cache is now automatically registered when `%kernel.debug%` is false.

stof:
@Philosoft this comment was about profiling the debug mode, not the prod mode.

loevgaard:
In production what cache is used? Doctrine [recommends](https://www.doctrine-project.org/projects/doctrine-orm/en/2.14/reference/caching.html) the `PhpFilesAdapter`, but I cannot see from a first glance that this is the case for production?

PR #1023, opened 2021-11-10
Add an embedded favicon to the base template

javiereguiluz:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

This adds an inlined/embedded favicon to avoid 404 errors when browsers try to get it.

We can change the emoji character if you like. E.g. the black circle --> ⚫️  could resemble the Symfony logo?


fabpot:
@javiereguiluz Let's use the black circle!

javiereguiluz:
I've updated to use the "black circle" emoji. It works and it looks OK on Chrome and Firefox (on macOS). It doesn't work on Safari. See:

favicon

PR #1026, opened 2021-11-22
[symfony/security-bundle] use lowest possible p/w hash "work factor" in test env

### LINKED ISSUE #1024: [security-bundle recipe] Use less resource-intensive password hasher for `test` environment
As explained in different blog posts ([here](https://maks-rafalko.github.io/blog/2021-11-21/symfony-tests-performance/#using-more-simple-password-hasher), [here](https://titouangalopin.com/posts/60edL3P43zwG6uGUiIlvPL/tips-for-a-reliable-and-fast-test-suite-with-symfony-and-doctrine)), and also in the docs - default password hasher (`auto`) consumes a lot of resources and takes too much time, especially when it's executed thousands of times in functional tests.

From https://symfony.com/doc/current/security/passwords.html#configuring-a-password-hasher
> Hashing passwords is resource intensive and takes time in order to generate secure password hashes. In general, this makes your password hashing more secure.
> 
> In tests however, secure hashes are not important, so you can change the password hasher configuration in test environment to run tests faster:

```yaml
# config/packages/test/security.yaml
password_hashers:
    # Use your user class name here
    App\Entity\User:
        algorithm: plaintext
```

So, what do you think to update an official recipe and in addition to [`config/packages/security.yaml`](https://github.com/symfony/recipes/blob/a624d3bf1def68bbd2d542b3a99216c0af090552/symfony/security-bundle/5.3/config/packages/security.yaml) add one more file - `config/packages/test/security.yaml` with the code above?

This will help many developers creating *new* projects to not waste time with functional tests.

Related to 

* https://github.com/api-platform/docs/pull/1472
* https://maks-rafalko.github.io/blog/2021-11-21/symfony-tests-performance/#using-more-simple-password-hasher

Thank you.

kbond:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | Fix #1024

Perhaps we could create a [`test-hasher`](https://github.com/symfony/recipes/issues/1024#issuecomment-975504242) in Symfony itself but this wouldn't be available until 6.1. Until then, I believe this is an easy win.

This is documented in the [4.4 docs](https://symfony.com/doc/4.4/testing/http_authentication.html) but was lost (I think) after a documentation refactor.

My argument for not using `plaintext` [is here](https://github.com/symfony/recipes/issues/1024#issuecomment-975015039) but I'm not strongly opposed to it.

### DIFF
--- a/symfony/security-bundle/5.3/config/packages/security.yaml
+++ b/symfony/security-bundle/5.3/config/packages/security.yaml
@@ -29,7 +29,7 @@ security:
 when@test:
     security:
         password_hashers:
-            # By default, password encoders are resource intensive and take time. This is
+            # By default, password hashers are resource intensive and take time. This is
             # important to generate secure password hashes. In tests however, secure hashes
             # are not important, waste resources and increase test times. The following
             # reduces the work factor to the lowest possible values.
--- a/symfony/security-bundle/5.3/config/packages/security.yaml
+++ b/symfony/security-bundle/5.3/config/packages/security.yaml
@@ -25,3 +25,16 @@ security:
     access_control:
         # - { path: ^/admin, roles: ROLE_ADMIN }
         # - { path: ^/profile, roles: ROLE_USER }
+
+when@test:
+    security:
+        password_hashers:
+            # By default, password encoders are resource intensive and take time. This is
+            # important to generate secure password hashes. In tests however, secure hashes
+            # are not important, waste resources and increase test times. The following
+            # reduces the work factor to the lowest possible values.
+            Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface:
+                algorithm: auto
+                cost: 4 # Lowest possible value for bcrypt
+                time_cost: 3 # Lowest possible value for argon
+                memory_cost: 10 # Lowest possible value for argon


symfony/security-bundle/5.3/config/packages/security.yaml
```diff
@@ -25,3 +25,12 @@ security:
     access_control:
         # - { path: ^/admin, roles: ROLE_ADMIN }
         # - { path: ^/profile, roles: ROLE_USER }
+
+when@test:
```
chalasr, line 29: Maybe add a small comment explaining that this is for performance purpose?

kbond: I copied the note from the old documentation.

symfony/security-bundle/5.3/config/packages/security.yaml
```diff
@@ -25,3 +25,16 @@ security:
...
```
weaverryan, line 29: Can we use `when@test` here? It's probably fine... but there's no guarantee the user has FWBundle 5.3 :/

symfony/security-bundle/5.3/config/packages/security.yaml
```diff
@@ -25,3 +25,16 @@ security:
...
+    security:
+        password_hashers:
+            # By default, password encoders are resource intensive and take time. This is
```
chalasr, line 32:
```suggestion
            # By default, password hashers are resource intensive and take time. This is
```

symfony/security-bundle/5.3/config/packages/security.yaml
stof: that's not related to FrameworkBundle. That's related to the version of the DI component and of HttpKernel. And for those, we have that guarantee

chalasr (approved):

kbond:
I confirmed with @maks-rafalko that there is virtually no performance difference between this method and plaintext.

nicolas-grekas (approved):

PR #1037, opened 2021-12-16
Copying the "archived/" directory from recipes-checker

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | TODO (mentioned on the main PR)

See https://github.com/symfony/flex/pull/845

Once https://github.com/symfony-tools/recipes-checker/pull/2 is merged, this can be merged.

### DIFF
--- a/.github/workflows/callable-flex-update.yml
+++ b/.github/workflows/callable-flex-update.yml
@@ -47,5 +47,8 @@ jobs:
                     git rm -q *.json
                     mv .github/flex-endpoint/*.json .
                     git add *.json
+                    # merge in new archived files
+                    for file in ../flex_endpoint/archived/*/*.json; do mv "$file" "archived/$(basename $(dirname "$file"))/$(basename "$file")"; done
+                    git add archived/*/*.json
                     git commit -m 'Update Flex endpoint' || true
                     git push origin -f flex/main
--- a/.github/workflows/callable-flex-update.yml
+++ b/.github/workflows/callable-flex-update.yml
@@ -48,7 +48,8 @@ jobs:
                     mv .github/flex-endpoint/*.json .
                     git add *.json
                     # merge in new archived files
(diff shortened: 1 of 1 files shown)


.github/workflows/callable-flex-update.yml
```diff
@@ -47,5 +47,8 @@ jobs:
                     git rm -q *.json
                     mv .github/flex-endpoint/*.json .
                     git add *.json
+                    # merge in new archived files
+                    for file in ../flex_endpoint/archived/*/*.json; do mv "$file" "archived/$(basename $(dirname "$file"))/$(basename "$file")"; done
```
nicolas-grekas, line 51: where does `flex_endpoint` come from?
shouldn't this be `mkdir -p archived; mv .github/flex-endpoint/archived/*.json archived/` instead?

weaverryan: Ha! Yes, sorry - you're right - this was a local testing path.

So I've added the `mkdir -p archived` (that was missing)Ha! Yes, sorry - you're right - this was a local testing path.

I've added the `mkdir -p archived` (that was missing). I've also fixed the path.

However, I don't think the `mv` you suggested works because the `.json` files are in subdirectories - e.g. they are `archived/symfony.twig-bundle.3.3/e13a79427495e670ef7a19f0e25fd13fbdd909d4.json`. It IS very possible that my mv statement can be simplified, let me know if you have an idea :)

.github/workflows/callable-flex-update.yml
```diff
@@ -47,5 +47,9 @@ jobs:
                     git rm -q *.json
                     mv .github/flex-endpoint/*.json .
                     git add *.json
```
nicolas-grekas, line 49: just after:
```
cp -a .github/flex-endpoint/archived .
git add archived/
```
and remove the rest?

weaverryan: Yup, you nailed it - that worked. Thank you for pointing that out :)

PR #1038, opened 2021-12-17
Add callable action to repopulate the recipe archives

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | None

Relates to: https://github.com/symfony-tools/recipes-checker/pull/3

On push to `main` - which is rare - we would run the new command to completely repopulate the `archived` directory on the `flex/main` branch. This really only needs to be done once, because it's kept up to date after automatically. But we could choose to run it on push just in case.

There IS one problem: how to replicate this for `recipes-contrib`?

### DIFF
--- a/.github/workflows/callable-flex-update-archived.yml
+++ b/.github/workflows/callable-flex-update-archived.yml
@@ -4,7 +4,7 @@ on:
     workflow_call:
         inputs:
             branch:
-                default: main
+                default: master
                 required: false
                 type: string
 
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Let's use a manual trigger instead?
https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

nicolas-grekas:
This would mean calling that workshop `callable-*` and referencing it from recipe-contrib's master branch

weaverryan:
GREAT idea. WDYT about this attempt? I can't test, so I'm shooting a bit blindly here :)

PR #1040, opened 2021-12-18
Proposal to include stacktraces in logs

l-vo:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | 

I didn't find any PR talking about this subject, thus I give it a try. Is there any reason to not have stacktraces in logs in the default configuration ? It's mainly useful for prod environment, but may be handy in other environments too.


nicolas-grekas:
This might flood the logs. I feel like sentry is a better way to collect traces when needed.

fabpot:
I would also recommend against using that by default.

PR #1041, opened 2021-12-18
Use new scheb/2fa package (replaces scheb/2fa-bundle)

wouterj:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

As of version 5.0, scheb/2fa-bundle is now named scheb/2fa. This moves the recipes/alias to the new package name.

I also added config for the 6.x-dev version. I'm not sure how we manage recipes for unstable versions, but it allows to remove some config for guards. I'm happy to remove the commit if we don't want unstable recipes in this repo.

cc @scheb

### DIFF
--- a/scheb/2fa/5.0/config/packages/scheb_2fa.yaml
+++ b/scheb/2fa/5.0/config/packages/scheb_2fa.yaml
@@ -1,0 +1,8 @@
+# See the configuration reference at https://github.com/scheb/2fa/blob/master/doc/configuration.md
+scheb_two_factor:
+    security_tokens:
+        - Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken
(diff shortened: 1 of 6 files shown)


scheb:
I wouldn't want people to install the `scheb/2fa` package, as it is an all-in-one package, which is very likely pulling in extra dependencies that people don't need in their application. `scheb/2fa-bundle` + extra packages based on individual needs is what should be used.

The `6.x` branch is still in heavy development, so I wouldn't include that in the recipes (yet).

wouterj:
Oh, alright. Fine by me 🙂

I was trying to use the 6.x branch (as a test) in a Symfony 6 app and couldn't find it on packagist. So I thought the package was deprecated.

scheb:
I'll open up a PR once the 6.x is ready, got that on my to-do list 👍

PR #1044, opened 2021-12-23
Consolidating env-specific files into main config file for simplicity

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | I'm working on the Encore docs right now. TODO Doctrine docs

Strictly speaking, we can't know for sure that the user is using 5.3 and higher for these packages. However, only 4.4 is still maintained. To be pragmatic, I think we should remove them now. We're already doing this for various other recipes, like [symfony/routing](https://github.com/symfony/recipes/blob/master/symfony/routing/5.3/config/packages/routing.yaml) and [twig.yaml](https://github.com/symfony/recipes/blob/master/symfony/twig-bundle/5.3/config/packages/twig.yaml).

I've made this change only in the most recent versions. Also, I've left environment-specific files still for Monolog and things like web profiler, where they are only enabled in certain environments.

Cheers!


weaverryan:
The CI failure... may be false? 

> The doctrine/orm package is required when the doctrine.orm config is set.

This is true... but the CI only runs:

> composer require -W --ansi 'doctrine/doctrine-bundle:^2.4' 'symfony/webpack-encore-bundle:^1.9'

Iirc, this is a known "limitation" we live with: we assume you will install and use the ORM if you install DoctrinBundle, even though that is not strictly a requirement.

fabpot (approved):
Great move!

symfony/webpack-encore-bundle/1.9/config/packages/webpack_encore.yaml
```diff
@@ -33,3 +33,17 @@ webpack_encore:
...
+
+framework:
+    assets:
+        json_manifest_path: '%kernel.project_dir%/public/build/manifest.json'
+
+#when@prod:
+    #webpack_encore:
+        # Cache the entrypoints.json (rebuild Symfony's cache when entrypoints.json changes)
+        # Available in version 1.2
+        #cache: true
```
nicolas-grekas, line 45: what about putting the `#` at the beginingof the line for all `when@` blocks?

fabpot:
I would recommend doing it for everything. See #1055 for instance, it is almost always better to be able to see all the configurations in one file to get the big picture.

PR #1004, opened 2021-09-22
pass static functions to runtime (clearer intent and improves performance)

kaznovac:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | symfony/symfony-docs#...

before:
```
% autocannon -c 10 -a 100 http://127.0.0.1:8080 
Running 100 requests test @ http://127.0.0.1:8080
10 connections

┌─────────┬───────┬────────┬────────┬────────┬───────────┬──────────┬────────┐
│ Stat    │ 2.5%  │ 50%    │ 97.5%  │ 99%    │ Avg       │ Stdev    │ Max    │
├─────────┼───────┼────────┼────────┼────────┼───────────┼──────────┼────────┤
│ Latency │ 94 ms │ 442 ms │ 584 ms │ 587 ms │ 441.91 ms │ 96.86 ms │ 630 ms │
└─────────┴───────┴────────┴────────┴────────┴───────────┴──────────┴────────┘
┌───────────┬────────┬────────┬────────┬─────────┬─────────┬────────┬────────┐
│ Stat      │ 1%     │ 2.5%   │ 50%    │ 97.5%   │ Avg     │ Stdev  │ Min    │
├───────────┼────────┼────────┼────────┼─────────┼─────────┼────────┼────────┤
│ Req/Sec   │ 14     │ 14     │ 21     │ 22      │ 20      │ 3.04   │ 14     │
├───────────┼────────┼────────┼────────┼─────────┼─────────┼────────┼────────┤
│ Bytes/Sec │ 736 kB │ 736 kB │ 1.1 MB │ 1.16 MB │ 1.05 MB │ 159 kB │ 736 kB │
└───────────┴────────┴────────┴────────┴─────────┴─────────┴────────┴────────┘

Req/Bytes counts sampled once per second.

100 requests in 5.02s, 5.25 MB read
```

after:
```
% autocannon -c 10 -a 100 http://127.0.0.1:8080
Running 100 requests test @ http://127.0.0.1:8080
10 connections

┌─────────┬────────┬────────┬────────┬────────┬──────────┬──────────┬────────┐
│ Stat    │ 2.5%   │ 50%    │ 97.5%  │ 99%    │ Avg      │ Stdev    │ Max    │
├─────────┼────────┼────────┼────────┼────────┼──────────┼──────────┼────────┤
│ Latency │ 106 ms │ 401 ms │ 472 ms │ 472 ms │ 388.1 ms │ 70.83 ms │ 473 ms │
[...]

PR #1046, opened 2021-12-31
improve index.php for sf6

garak:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | none

Improve the content of front controller, by exploiting latest PHP features

### DIFF
--- a/symfony/framework-bundle/6.0/config/packages/cache.yaml
+++ b/symfony/framework-bundle/6.0/config/packages/cache.yaml
@@ -1,0 +1,19 @@
+framework:
+    cache:
+        # Unique name of your app: used to compute stable namespaces for cache keys.
+        #prefix_seed: your_vendor_name/app_name
+
+        # The "app" cache stores to the filesystem by default.
+        # The data in this cache should persist between deploys.
+        # Other options include:
+
+        # Redis
+        #app: cache.adapter.redis
+        #default_redis_provider: redis://localhost
+
+        # APCu (not recommended with heavy random-write workloads as memory fragmentation can cause perf issues)
+        #app: cache.adapter.apcu
+
+        # Namespaced pools use the above "app" backend by default
+        #pools:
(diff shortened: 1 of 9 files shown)


wouterj:
Hi! Thanks for the proposal.

However, I'm not sure about the added value of this change. As discussed in #1004, `static` doesn't bring anything for global functions.
The added value of arrow functions is their automatic binding to the current scope (i.e. access to variables from the scope of the function). This however is in contrast with the purpose of this closure in the Runtime component: to decouple the inside of the closure from its runtime context.

Btw, if other people do agree with this change, it should not be done for 6.0 but only for 6.1 ([see the update policy in the README](https://github.com/symfony/recipes#updating-recipes)).

kbond:
I agree with @wouterj.

garak:
> Btw, if other people do agree with this change, it should not be done for 6.0 but only for 6.1 ([see the update policy in the README](https://github.com/symfony/recipes#updating-recipes)).

The README mentions "the next version". The current version (the higher available) is 5.4, so 6.0 in the next.

derrabus:
> The current version (the higher available) is 5.4, so 6.0 in the next.

No, 6.0 is the current stable version, see https://symfony.com/releases

garak:
> > The current version (the higher available) is 5.4, so 6.0 in the next.
> 
> No, 6.0 is the current stable version, see https://symfony.com/releases

We're talking about **recipe** version, right? Not about Symfony version.

fabpot:
The recipe version is the same as the package version.

garak:
> The recipe version is the same as the package version.

So why there's no 6.0 in recipe?

stof:
@garak because we want to apply the same recipe for 5.4 and 6.0, given that 6.0 is the same than 5.4 without deprecations (and the recipe does not rely on deprecated features). We avoid copying the recipe when there is no difference.

PR #1047, opened 2022-01-03
Update controller syntax in scheb/2fa-bundle

scheb:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

The controller syntax needs to be different for Symfony 6

### DIFF
--- a/scheb/2fa-bundle/5.0/config/routes/scheb_2fa.yaml
+++ b/scheb/2fa-bundle/5.0/config/routes/scheb_2fa.yaml
@@ -1,7 +1,7 @@
 2fa_login:
     path: /2fa
     defaults:
-        _controller: "scheb_two_factor.form_controller:form"
+        _controller: "scheb_two_factor.form_controller::form"
 
 2fa_login_check:
     path: /2fa_check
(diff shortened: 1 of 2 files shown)


stof (approved):

stof:
Maybe this should also be updated in the recipe for the 5.0 version of the bundle, given that the new syntax works in Symfony 4.1+ and the old syntax (with a single `:`) is deprecated in 5.1+

scheb:
Oh, didn't know Symfony 4.1+ is fine with that syntax. Thanks for letting me know! I've added a commit to change it on 5.x as well.

chalasr (approved):

stof:
@scheb yeah, the deprecated was initially added in 4.1 as well, planning to drop it in 5.0, but it was too painful to write bundles supporting both the 3.4 LTS and 4.1+ without deprecation triggered by their routing files, so we delayed the deprecation by a major version (the extra code to handle the old syntax was just 3 LOC so the maintenance cost was acceptable)

stof:
@fabpot @nicolas-grekas the CI fails because it tries to run the recipe using PHP 7.4 while the bundle requires PHP 8+. What is the policy for this CI ?

nicolas-grekas:
Ignoring the CI :)

stof:
Maybe we should migrate the CI to use PHP 8. I suspect more and more packages will start using it as their min version (we did it in Symfony 6)

yceruto (approved):

stof:
Merging this, as the bot won't ignore the CI

PR #1051, opened 2022-01-14
[FrameworkBundle] Remove src/Tests from list of excluded folders

HypeMC:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Since Symfony 4 & Flex the default directory structure doesn't have a `src/Tests` folder making the exclude redundant.

### DIFF
--- a/symfony/framework-bundle/4.4/config/services.yaml
+++ b/symfony/framework-bundle/4.4/config/services.yaml
@@ -19,7 +19,6 @@ services:
             - '../src/DependencyInjection/'
             - '../src/Entity/'
             - '../src/Kernel.php'
-            - '../src/Tests/'
 
     # controllers are imported separately to make sure services can be injected
     # as action arguments even if you don't extend any base controller class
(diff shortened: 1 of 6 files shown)


symfony/framework-bundle/4.4/config/services.yaml
```diff
@@ -19,7 +19,6 @@ services:
             - '../src/DependencyInjection/'
```
yceruto, line 19: I'd also remove `DependencyInjection` as there is no convention for it at the project level (it's the case for bundles). Furthermore, we can currently use the same Kernel class to define/handle a DI extension.

derrabus: I think it's sensible to leave `DependencyInjection` in here. It's still the place I'd put my compiler passes in and ignoring the directory here makes it clear that those passes should not be registered as services.

derrabus (approved):

OskarStark (approved):

stof (approved):

PR #1057, opened 2022-02-12
phpunit: let APP_DEBUG enable debug

ro0NL:
| Q             | A
| ------------- | ---
| License       | MIT

So that deprecations can be logged in APP_ENV=test with APP_DEBUG=1

```
# monolog.yaml
monolog:
  channels: [deprecation]
  handlers:
    deprecations:
      type: stream
      path: "%kernel.logs_dir%/%kernel.environment%.deprecations.log"
      channels: [ "deprecation" ]
```

### DIFF
--- a/phpunit/phpunit/9.3/tests/bootstrap.php
+++ b/phpunit/phpunit/9.3/tests/bootstrap.php
@@ -1,6 +1,7 @@
 <?php
 
 use Symfony\Component\Dotenv\Dotenv;
+use Symfony\Component\ErrorHandler\Debug;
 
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Which part of what `Debug::enable()` does is required? e.g DebugClassLoader is already enabled by the phpunit-bridge.

ro0NL:
Im invoking `phpunit` :)

ro0NL:
Right, didn't realize ... but there are 2 files: https://github.com/symfony/recipes/blob/master/symfony/phpunit-bridge/5.3/tests/bootstrap.php

PR #1059, opened 2022-02-16
Specify charset when using MySQL

greg0ire:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc PR  | https://github.com/symfony/symfony-docs/pull/16515

Because of backward-compatibility reasons, not specifying the charset
will result in "utf8" being used, which is a deprecated charset.
Also, when the charset for creating tables is not specified, the one
specified in the DSN is reused if specified.

In the future, DBAL is going to drop default values, which means we will
have to specify a charset anyway.

See:
- https://github.com/doctrine/dbal/pull/5277 (about the reuse of connection charset for table charset)
- https://github.com/doctrine/dbal/pull/4644


derrabus (approved):

nicolas-grekas:
The recipe for 2.3 should also be updated, so that projects that start on 4.4 also get the change.

PR #1060, opened 2022-02-17
Skip recipe for doctrine/annotations when symfony/routing 6 is installed

### LINKED ISSUE #1030: duplicate recipes for routing and annotations
the `doctrine/annotations` recipe creates a `config/routes/annotations.yaml` that contains 

```yaml
controllers:
    resource: ../../src/Controller/
    type: annotation

kernel:
    resource: ../../src/Kernel.php
    type: annotation
```

https://github.com/symfony/recipes/blob/master/doctrine/annotations/1.0/config/routes/annotations.yaml

&&

the `symfony/router` recipe creates `config/routes.yaml` that contains 

```yaml
controllers:
    resource: ../src/Controller/
    type: annotation

kernel:
    resource: ../src/Kernel.php
    type: annotation
```

https://github.com/symfony/recipes/blob/master/symfony/routing/6.0/config/routes.yaml

Which recipe can we remove?

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Fix #1030

### DIFF
--- a/doctrine/annotations/1.10/manifest.json
+++ b/doctrine/annotations/1.10/manifest.json
@@ -1,0 +1,5 @@
+{
+    "conflict": {
+        "symfony/routing": "<6.0"
+    }
+}


doctrine/annotations/1.10/manifest.json
```diff
@@ -0,0 +1,5 @@
+{
+    "conflict": {
+        "symfony/routing": "<6.0"
```
alexislefebvre, line 3: Is this rule contradicting the title of this PR?

nicolas-grekas: this recipe is empty, but only when used with >=6.0, that's the point

alexislefebvre: Right, thanks for your answer.

PR #1066, opened 2022-03-08
Simplify env processing

94noni:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | https://github.com/symfony/symfony-docs/pull/16575

Friendly ping @javiereguiluz 
Here is the same proposition as the doc PR related to removing this `resolve` env processor as it is not needed


nicolas-grekas:
It is needed when using sqlite.

javiereguiluz:
Let's close this one for the reason given by Nicolas. Instead, we'll change the docs a bit. Thanks!

PR #1071, opened 2022-03-30
[symfony/mailer] use `null://null` for `MAILER_DSN`

kbond:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

This makes it easier to run tests that send emails "out of the box".


Kocal:
We could _simply_ update the `.env.test` and add `MAILER_DSN=null://null` line, no?

kbond:
I don't believe the recipe can _update_ `.env.test`, can it? Or are you suggesting adding to the phpunit recipe?

kbond:
> Yeah it looks like it can

I meant add a single line to the `.env.test` file.

> But that was just a suggestion

Both achieve the same thing, doesn't matter to me. Adding the the phpunit/phpunit-bridge recipes seems simpler.

kbond:
I modified this PR to just always use `null://null` instead of `smtp://localhost`. I feel it rather unlikely someone has a local mail server. The docker integration is unaffected by this.

fabpot (approved):
I think it makes sense.

PR #1073, opened 2022-04-05
Update package.json

94noni:
| Q             | A
| ------------- | ---
| License       | MIT

default mandatory props


stof:
`version` is not actually mandatory when you don't plan to publish a package in the registry.

94noni:
hum indeed @stof 
should I close or only remove the `version` part?

stof:
IIRC, the name is also optional when you don't publish it. It is only required when you want to submit to the registry. So I would close this PR.

94noni:
@stof understood, thx for your time :)

PR #1075, opened 2022-04-20
[routing] Remove useless config in routes.yaml

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

Follows https://github.com/symfony/symfony/pull/46115

### DIFF
--- a/symfony/routing/6.1/manifest.json
+++ b/symfony/routing/6.1/manifest.json
@@ -1,0 +1,9 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "aliases": ["router"],
+    "conflict": {
+        "symfony/framework-bundle": "<5.3"
+    }
+}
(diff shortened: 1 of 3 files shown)


symfony/routing/6.1/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "aliases": ["router"],
+    "conflict": {
+        "symfony/framework-bundle": "<5.3"
```
nicolas-grekas, line 7: 6.1 isn't it, to get support for `type: attribute`

dunglas: It should be a conflict with `symfony/routing` isn't it? Or should I just keep `annotation` for now?

nicolas-grekas: my bad, nothing to do here, routing is already >= 6.1 by the name of the recipe and this is unrelated to fwb

PR #1078, opened 2022-04-26
Adding a new callable recipes readme GH action

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | None

Relates to https://github.com/symfony-tools/recipes-checker/pull/9. That should be merged first, then I can test this on a fork.

PR #1079, opened 2022-04-28
Partner GH action workflow for updating RECIPES.md

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | none

To be merged after #1078.

### DIFF
--- a/.github/workflows/flex-update.yml
+++ b/.github/workflows/flex-update.yml
@@ -14,3 +14,9 @@ jobs:
         uses: symfony/recipes/.github/workflows/callable-flex-update.yml@main
         with:
             versions_json: .github/versions.json
+
+    call-recipe-readme-update:
+        needs: 'call-flex-update'
+        uses: symfony/recipes/.github/workflows/callable-recipe-readme-update.yml@main
+        with:
+            branch: main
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
[Oops](https://github.com/symfony/recipes/runs/6241616892):

```
remote: error: GH006: Protected branch update failed for refs/heads/main.        
remote: error: Required status check "call-qa / Run checks" is expected. At least 2 approving reviews are required by reviewers with write access.        
To https://github.com/symfony/recipes
 ! [remote rejected] main -> main (protected branch hook declined)
```

nicolas-grekas:
There's something strange also: this workflow will run on "push" to "main", and it will push to main. There's a circular loop here. It will likely stop by itself since the RECIPES.md file won't have any changes to be committed, but then, I'm wondering if the action won't fail because either "commit" or "push" might return a non-zero status code because of this empty diff.

nicolas-grekas:
We should maybe push to flex/main instead of main. That'd fix both issues.

weaverryan:
I’ll work on that :). I thought about the circular loop, but it didn’t actually happen in real life - my theory is that a push from GH actions doesn’t trigger another action. But, pushing to flex/main seems totally fine.

PR #1087, opened 2022-05-18
[webpack-encore-bundle] disable strict_mode by default

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Now that webapp-pack [contains webpack-encore-bundle](https://github.com/symfony/webapp-pack/blob/main/composer.json#L31), extending [`base.html.twig`](https://github.com/symfony/recipes/blob/main/symfony/twig-bundle/5.4/templates/base.html.twig) throws an exception unless one runs `yarn` beforehand.

This is bad DX for newcomers.


weaverryan (approved):
Ok, I'm fine with this

derrabus (approved):

Issue #1089, opened 2022-05-25
Symfony webapps and their hard reliance on Stimulus


NMe84:
This has been bugging me for a while and I never really knew where to bring this issue up. I ran into it again today and decided that instead of just staying quiet I should just decide on a place to bring this up and start a discussion, so here we are.

I'll just get right to the point: I don't understand why, when creating a new webapp project, Symfony is opinionated on which Javascript libraries to use. Nowadays when you create a new webapp, Symfony includes Stimulus and the Stimulus-bridge (which is fine) but then after that the default recipe also creates the setup for Stimulus in Webpack's config and actually includes the folder and file structure for it.

I think this is a fundamentally bad idea. I don't have anything against Stimulus and I understand that the option should be there, but it should definitely not be on by default. If I want to use React, Angular, Vue, Svelte or any other option I first have to remove all the boilerplate that the recipe added and in recent times that's literally what my process looks like whenever I start a new project: I install Symfony, then go back manually and remove anything Stimulus related. Ever single time. It's annoying and unnecessary and I can't for the life of me understand why any of it is enabled by default in a framework that otherwise says it's as unopinionated as it can be. And I can't be the only person who has this issue, the other frameworks I mentioned are by far more popular than Stimulus has ever been and most likely more popular than it ever will be.

Can someone enlighten me and explain what I'm missing? And would you be willing to reconsider this default behaviour?

derrabus:
Thank you for opening this discussion.

To be clear, you are getting all that boilterplate you complain about when you use the `--webapp` switch for the `symfony new` command:

```sh
symfony new --webapp my-new-app
```

What exactly is a webapp? Or rather, what is a good skeleton for a webapp? You probably realize that if you ask ten developers this question, you will get eleven answers. At least. That webapp switch is opinionated by design and if we changed what it does, we would just implement someone else's opinion.

If you omit that switch, you get a minimal project to which you can add only what you need. As long as you don't install any Stimulus-related packages, you should not get Stimulus-related boilerplate.

I myself hardly use `--webapp` because I do have strong opinions on how my app should be laid out and which libraries it should use for what purpose. I don't expect a scaffolding tool like Symfony Flex to meet those opinions, so starting with the minimal app is the best way for me.

What's stopping you from starting with the minimal app like I do?

NMe84:
> That webapp switch is opinionated by design and if we changed what it does, we would just implement someone else's opinion.

[...]

PR #1090, opened 2022-05-27
[RCF][WIP] Proposed symfony/ux-meta Recipe

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | Later if accepted

This is a proposal to split Stimulus/Symfony UX out of the `symfony/webpack-encore-bundle`. 2 reasons:

1) Some people (e.g. #1089) don't want Stimulus. Alone, I don't personally think this is a good enough reason. You have to optimize one path, it won't work for everyone.

2) Separating it MAY make the purpose and functionality of Symfony UX more clear. Currently, to install Symfony UX, you run `composer require symfony/webpack-encore-bundle`... which is a little weird. After, you would run `composer require symfony/ux`.

## How it would work:

* A) We create a new `symfony/ux-package` which requires `symfony/webpack-encore-bundle` and `symfony/ux-meta`

* B) We create a new `symfony/ux-meta` package. This will hold the "full UX setup", which is equivalent to the current `symfony/webpack-encore-bundle` recipe.

That's it! When you run `composer require symfony/ux`, it will install WebpackEncoreBundle, but the`symfony/ux-meta` recipe will take precedence.

## Complications

* i) In the future, when people run `composer recipes:update symfony/webpack-encore-bundle`, it will try to delete some files like `assets/controllers.json`, `assets/bootstrap.js` and will try to remove the `import './bootstrap'` line from `assets/app.js`. A blog post about what's happening could alleviate this.
* ii) If you install `symfony/webpack-encore-bundle` FIRST and THEN `symfony/ux`, you will NOT have a complete setup. You will be missing the `import './bootstrap'` line from `assets/app.js`, 2 packages from `package.json` and 1 line from `webpack.config.js`.

To fix (ii), we would need (A) a simple "patch" system in Flex to basically "insert" some a line into the middle of an existing file (e.g. `webpack.config.js` or `assets/app.js` and (B) a Flex configurator for `package.json` (we have such a thing, but we would need to extend it to read some packages from `manifest.json`.

Relates to #1089


symfony/ux-meta/1.0/config/packages/webpack_encore.yaml
```diff
@@ -0,0 +1,49 @@
+webpack_encore:
```
stof, line 1: This file should not be in the ux-meta recipe. It should stay managed fully by the recipe of the WebpackEncoreBundle IMO.

symfony/ux-meta/1.0/manifest.json
```diff
@@ -0,0 +1,21 @@
+{
+    "bundles": {
+        "Symfony\\WebpackEncoreBundle\\WebpackEncoreBundle": ["all"]
```
stof, line 3: ux-meta should not register WebpackEncoreBundle (especially not if `symfony/ux-meta` does not directly require the bundle)

symfony/ux-meta/1.0/manifest.json
```diff
@@ -0,0 +1,21 @@
...
+    },
+    "copy-from-recipe": {
+        "assets/": "assets/",
+        "config/": "%CONFIG_DIR%/",
+        "package.json": "package.json",
+        "webpack.config.js": "webpack.config.js"
+    },
+    "aliases": ["ux", "encore", "webpack", "webpack-encore"],
```
stof, line 11: The webpack and encore aliases should not be added on this ux-meta package (and `ux` should be removed from the WebpackEncoreBundle recipe)

stof (changes requested):

symfony/ux-meta/1.0/config/packages/webpack_encore.yaml
weaverryan: You’re right! Only files that need to be different for UX should be here.

symfony/ux-meta/1.0/manifest.json
```diff
@@ -0,0 +1,21 @@
...
+    "gitignore": [
+        "/node_modules/",
+        "/%PUBLIC_DIR%/build/",
+        "npm-debug.log",
+        "yarn-error.log"
+    ],
+    "conflict": {
+        "symfony/framework-bundle": "<5.4"
```
stof, line 19: this conflict is probably not necessary anymore once you remove the webpack_encore.yaml file btw

fabpot:
@weaverryan What do we need to move forward here?

weaverryan:
I can close this for now. I DO think we should split the Stimulus stuff out of WebpackEncoreBundle into its own package. When we do that, some recipe changes will be needed and we can revisit.

PR #1093, opened 2022-06-03
Move recipes from metas to packs

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Since https://github.com/symfony/flex/pull/923, recipes can be bound to packs and still be managed by `recipes:update`.
Aka packs don't need a sibling `*-meta` repo anymore.


kbond (approved):

dunglas (approved):

stof:
indeed, this should not have been done before releasing the Flex version supporting the pre-requisite, because recipe changes go live on merge, while Flex changes don't propagate in the ecosystem until the release is done (and even then, it is safer to give a bit of time)

nicolas-grekas:
All tags pushed. This should all work nicely now.

PR #1094, opened 2022-06-06
[symfony/ux-autocomplete] Adding UX autocomplete recipe

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | https://github.com/symfony/ux/pull/330

Hi!

This adds a recipe for the proposed new ux-autocomplete package... the only needed thing is the route import.

Cheers!


weaverryan:
This is ready. The QA fails because the CI ALWAYS tests with PHP 7.4... and this is a PHP 8 package only. I'm not sure what we should do about that.

kbond (approved):

fabpot (approved):

PR #1095, opened 2022-06-09
[web-profiler-bundle] Enable serializer data collector on 6.1

chalasr:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

See https://github.com/symfony/symfony/pull/46625 for background.

### DIFF
--- a/symfony/web-profiler-bundle/6.1/manifest.json
+++ b/symfony/web-profiler-bundle/6.1/manifest.json
@@ -1,0 +1,11 @@
+{
+    "bundles": {
+        "Symfony\\Bundle\\WebProfilerBundle\\WebProfilerBundle": ["dev", "test"]
+    },
+    "copy-from-recipe": {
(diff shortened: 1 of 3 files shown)


symfony/web-profiler-bundle/6.1/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+    "bundles": {
+        "Symfony\\Bundle\\WebProfilerBundle\\WebProfilerBundle": ["dev", "test"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "conflict": {
+        "symfony/framework-bundle": "<5.3"
```
nicolas-grekas, line 9: should be 6.1, isn't it?

chalasr: probably yes since the option is from framework-bundle, what made you rollback? ^^

PR #1119, opened 2022-08-17
Add example how to bind arguments to the services.yaml

alexander-schranz:
| Q             | A
| ------------- | ---
| License       | MIT

I always must google if it is `services.bind`, `services._bind` or `services._defaults.bind` or even `services._defaults._bind`. Think it would be great to have this example in it.

### DIFF
--- a/symfony/framework-bundle/6.2/config/services.yaml
+++ b/symfony/framework-bundle/6.2/config/services.yaml
@@ -10,6 +10,8 @@ services:
     _defaults:
         autowire: true      # Automatically injects dependencies in your services.
         autoconfigure: true # Automatically registers your services as commands, event subscribers, etc.
+        # Bind arguments globally
+        # https://symfonycasts.com/screencast/symfony-fundamentals/global-bind
         #bind:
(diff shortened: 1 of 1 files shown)


symfony/framework-bundle/6.2/config/services.yaml
```diff
@@ -10,6 +10,11 @@ services:
     _defaults:
         autowire: true      # Automatically injects dependencies in your services.
         autoconfigure: true # Automatically registers your services as commands, event subscribers, etc.
+        # Bind arguments globally
+        # https://symfonycasts.com/screencast/symfony-fundamentals/global-bind
```
fabpot, line 14: I'd prefer a reference to the Symfony docs.

alexander-schranz: link updated

fabpot (approved):

nicolas-grekas:
Shouldn't we revert this PR? In 6.2, people are going to use `#[Autowire]` most of the time instead.

nicolas-grekas:
See #1119

PR #1122, opened 2022-08-29
Doctrine Migration profiler disabled by default

94noni:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | https://github.com/symfony/symfony/issues/47275

imho we should not activate this by default, but rather let ppl discover this feature as [documented here](https://github.com/doctrine/DoctrineMigrationsBundle/pull/471)

### DIFF
--- a/doctrine/doctrine-migrations-bundle/3.1/config/packages/doctrine_migrations.yaml
+++ b/doctrine/doctrine-migrations-bundle/3.1/config/packages/doctrine_migrations.yaml
@@ -3,4 +3,4 @@ doctrine_migrations:
         # namespace is arbitrary but should be different from App\Migrations
         # as migrations classes should NOT be autoloaded
         'DoctrineMigrations': '%kernel.project_dir%/migrations'
-    enable_profiler: '%kernel.debug%'
+    enable_profiler: false # '%kernel.debug%' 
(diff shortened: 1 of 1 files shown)


stof:
I agree about that. This profiler has a performance impact while showing something not really related to the request. And it adds extra doctrine queries that make using the Doctrine panel less useful.

doctrine/doctrine-migrations-bundle/3.1/config/packages/doctrine_migrations.yaml
```diff
@@ -3,4 +3,4 @@ doctrine_migrations:
         # namespace is arbitrary but should be different from App\Migrations
         # as migrations classes should NOT be autoloaded
         'DoctrineMigrations': '%kernel.project_dir%/migrations'
-    enable_profiler: '%kernel.debug%'
+    enable_profiler: false # '%kernel.debug%'
```
stof, line 6: The comment should be removd entirely IMO.

stof (changes requested):

doctrine/doctrine-migrations-bundle/3.1/config/packages/doctrine_migrations.yaml
fabpot: Agreed

kbond (approved):

weaverryan (approved):

OskarStark (approved):

PR #1124, opened 2022-09-12
Increase length of JWT secret to avoid errors in lcobucci/jwt ^4.2

silverbackdan:
I'm not sure if I should have added this in a new version or included previous config directories.

But the length of the JWT secret needs to be longer by default
See: https://github.com/symfony/symfony-docs/pull/17259/files
See: https://github.com/symfony/mercure/issues/89
See: https://github.com/lcobucci/jwt/releases/tag/4.2.0


stof (approved):

chalasr:
> I'm not sure if I should have added this in a new version or included previous config directories.

The version of the recipe must match an existing release of the target package, so given this is kind of a bugfix I think we should update all existing versions instead

silverbackdan:
Sure, 0.1 and 0.2 define the actual token and it is signed with the short key - do I need to update those?

silverbackdan:
Not quite sure what I need to change in the PR to avoid this error - sorry.

chalasr:
The verification is handled by the mercure hub which is written in Go and uses github.com/golang-jwt/jwt/v4 which does not have such validation, so this is not an issue for now but might be one day. However mercure-bundle 0.1 and 0.2 didn't use lcobucci/jwt so I'd say we don't care at all about these versions.

chalasr (dismissed):

chalasr:
> Not quite sure what I need to change in the PR to avoid this error - sorry.

No worry, I fear there's nothing we can do, all good for this PR. /cc @fabpot

symfony/mercure-bundle/0.3/manifest.json
```diff
@@ -12,7 +12,7 @@
         "#3": "The public URL of the Mercure hub, used by the browser to connect",
         "MERCURE_PUBLIC_URL": "https://example.com/.well-known/mercure",
         "#4": "The secret used to sign the JWTs",
-        "MERCURE_JWT_SECRET": "!ChangeMe!"
+        "MERCURE_JWT_SECRET": "!MUST_CHANGE_THIS_SECRET_JWT_KEY!"
```
dunglas, line 15:
```suggestion
        "MERCURE_JWT_SECRET": "!ChangeThisMercureHubJWTSecretKey!"
```

symfony/mercure-bundle/0.3/manifest.json
```diff
@@ -22,8 +22,8 @@
                 "  restart: unless-stopped",
                 "  environment:",
                 "    SERVER_NAME: ':80'",
-                "    MERCURE_PUBLISHER_JWT_KEY: '!ChangeMe!'",
-                "    MERCURE_SUBSCRIBER_JWT_KEY: '!ChangeMe!'",
+                "    MERCURE_PUBLISHER_JWT_KEY: '!MUST_CHANGE_THIS_SECRET_JWT_KEY!'",
```
dunglas, line 25:
```suggestion
                "    MERCURE_PUBLISHER_JWT_KEY: '!ChangeThisMercureHubJWTSecretKey!",
```

symfony/mercure-bundle/0.3/manifest.json
```diff
@@ -22,8 +22,8 @@
...
+                "    MERCURE_SUBSCRIBER_JWT_KEY: '!MUST_CHANGE_THIS_SECRET_JWT_KEY!'",
```
dunglas, line 26:
```suggestion
                "    MERCURE_SUBSCRIBER_JWT_KEY: '!ChangeThisMercureHubJWTSecretKey!'",
```

dunglas (changes requested):
I just released a patch version of Mercure changing the default key used in dev mode: https://github.com/dunglas/mercure/releases/tag/v0.14.1

Can you please use the same value, so the recipe will work out of the box in dev mode.

silverbackdan:
Sure thing @dunglas - updated the PR now.

dunglas (dismissed):

chalasr (dismissed):

chalasr:
@silverbackdan The CI was complaining about the missing table in the PR description (including the MIT license). I took the liberty to add it myself, hope it's ok for you. The check is still red though, can you please try to force-push to retrigger it?

silverbackdan:
Ah perfect - sorry I missed that.
@chalasr and @dunglas - nothing has changed in the PR just the commit message with the MIT licence in there too.

stof (approved):

dunglas (approved):

silverbackdan:
Merci

PR #1125, opened 2022-09-13
[api-platform/core] add 3.0 default configuration

soyuka:
| Q             | A
| ------------- | ---
| License       | MIT

Preparing v3 release see https://github.com/api-platform/api-platform/pull/2249

ping @dunglas

### DIFF
--- a/api-platform/core/3.0/config/routes/api_platform.yaml
+++ b/api-platform/core/3.0/config/routes/api_platform.yaml
@@ -1,0 +1,4 @@
+api_platform:
+    resource: .
+    type: api_platform
+    prefix: /api
--- a/api-platform/core/3.0/manifest.json
+++ b/api-platform/core/3.0/manifest.json
@@ -1,0 +1,9 @@
+{
+    "bundles": {
+        "ApiPlatform\\Symfony\\Bundle\\ApiPlatformBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/",
+        "src/": "%SRC_DIR%/"
+    }
+}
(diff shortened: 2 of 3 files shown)


api-platform/core/3.0/config/packages/api_platform.yaml
```diff
@@ -0,0 +1,11 @@
+api_platform:
+    title: Hello API Platform
+    version: 1.0.0
+    # Mercure integration, remove if unwanted
+    mercure: ~
```
stof, line 5: Shouldn't this rely on a smart default based on the presence of mercure in the dependencies, as done by FrameworkBundle for many settings ? If yes, it would allow not having that key in the recipe

api-platform/core/3.0/config/packages/api_platform.yaml
```diff
@@ -0,0 +1,11 @@
...
+    # Good cache defaults for REST APIs
```
stof, line 6: If these are good defaults, why aren't they the default value in ApiPlatformBundle directly ?

soyuka:
your right @stof I removed the configuration file alltogether.

api-platform/core/3.0/config/packages/api_platform.yaml
dunglas: Actually, these are good defaults for any Symfony application... This should be in FrameworkBundle. I don't want API Platform to override Symfony's default silently, it's why it's in the recipe and not in the bundle.

stof: But those are an ApiPlatform configuration setting, not a FrameworkBundle one.

dunglas (approved):

stof (approved):

stof:
@CRC-Mismatch you have better chance to get an answer if you ask in the ApiPlatform discussion channels (so in the repository or in the dedicated channel on Slack).

PR #1129, opened 2022-09-29
add MAILER_DSN by default

shyim:
| Q             | A
| ------------- | ---
| License       | MIT

I would like to have `MAILER_DSN` by default set to something. When the user now uses the mailer before changing the `.env` the code breaks and throws an exception as the env is not set

### DIFF
--- a/symfony/mailer/4.3/config/packages/mailer.yaml
+++ b/symfony/mailer/4.3/config/packages/mailer.yaml
@@ -1,3 +1,6 @@
+parameters:
+    env(MAILER_DSN): 'null://null'
+
 framework:
     mailer:
         dsn: '%env(MAILER_DSN)%'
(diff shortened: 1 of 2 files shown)


shyim:
A second idea we add a fallback for MAILER_DSN to `null://null` using parameters

fabpot:
I think the current setup makes more sense. You need to configure something. Either because you install a provider, or because you have Symfony CLI with a docker-compose that has a mail catcher. If not I think it makes sense to have an explicit configuration.

shyim:
I am currently adding flex into an own template and our application injects somewhere the Mailer and the entire Website is broken as it cannot find the `MAILER_DSN` by default.

But I think it's better for Symfony starter to have an exception that the env is missing, so you think about what the mailer should be.

So I fixed it for our own template and I guess that's enough https://github.com/shopware/recipes/commit/c24db539c12bc3359141bfdcecd1905fe70ce025

PR #1130, opened 2022-10-05
[league/oauth2-server-bundle] Add the recipe

chalasr:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

This bundle is maintained by myself and is an alternative to the no-longer-maintained FOSOAuthServerBundle.

### DIFF
--- a/league/oauth2-server-bundle/0.4/manifest.json
+++ b/league/oauth2-server-bundle/0.4/manifest.json
@@ -1,0 +1,14 @@
+{
+    "bundles": {
+        "League\\Bundle\\OAuth2ServerBundle\\LeagueOAuth2ServerBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "OAUTH_PRIVATE_KEY": "%kernel.project_dir%/%CONFIG_DIR%/jwt/private.pem",
+        "OAUTH_PUBLIC_KEY": "%kernel.project_dir%/%CONFIG_DIR%/jwt/public.pem",
+        "OAUTH_PASSPHRASE": "%generate(secret)%",
+        "OAUTH_ENCRYPTION_KEY": "%generate(secret)%"
+    }
+}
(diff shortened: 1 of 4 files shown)


derrabus (dismissed):

league/oauth2-server-bundle/0.4/manifest.json
```diff
@@ -0,0 +1,14 @@
+{
+    "bundles": {
+        "League\\Bundle\\OAuth2ServerBundle\\LeagueOAuth2ServerBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "OAUTH_PRIVATE_KEY": "%kernel.project_dir%/%CONFIG_DIR%/jwt/private.pem",
+        "OAUTH_PUBLIC_KEY": "%kernel.project_dir%/%CONFIG_DIR%/jwt/public.pem",
```
weaverryan, line 10: These are paths that will sometimes be different from one machine to another (hence using env vars)?

league/oauth2-server-bundle/0.4/post-install.txt
```diff
@@ -0,0 +1,10 @@
+  1. <bg=magenta;fg=white> Provide a key pair </>
+    * Generate a private/public key pair (with passphrase) at the locations defined in your .env file: <comment>https://oauth2.thephpleague.com/installation/#generating-public-and-private-keys</>
+    * Configure the <info>private_key</> and <info>public_key</> with the respective key locations
+    * Set the <info>private_key_passphrase</> to the private key password set in the previous step
```
weaverryan, line 4: In practice, they should set the `OAUTH_PASSPHRASE` env var, right? If so, does it make sense to "generate" one for them automatically in the recipe?

chalasr: Actually it's already the case, the post-install suggestion was wrong here. Updated

derrabus (approved):

PR #1131, opened 2022-10-07
switch schickling/mailcatcher to jeanberu/mailcatcher

ivoba:
| Q             | A
| ------------- | ---
| License       | MIT

Replace `schickling/mailcatcher` Docker image with `jeanberu/mailcatcher` because it is smaller with less MB.
Functionality should be the same.

`schickling/mailcatcher`: Compressed size: 25.44 MB
https://github.com/schickling/dockerfiles/blob/master/mailcatcher/Dockerfile

`jeanberu/mailcatcher`: Compressed size: 15.27 MB
https://github.com/Jean-Beru/docker-mailcatcher/blob/master/Dockerfile

This should save some MBs on the internets.

### DIFF
--- a/symfony/mailer/4.3/manifest.json
+++ b/symfony/mailer/4.3/manifest.json
@@ -9,7 +9,7 @@
         "docker-compose.override.yml": {
             "services": [
                 "mailer:",
-                "  image: schickling/mailcatcher",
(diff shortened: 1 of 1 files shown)


fabpot:
Closing in favor of #1281

PR #1134, opened 2022-10-17
mailer: missing xsl extension

COil:
| Q             | A
| ------------- | ---
| License       | MIT

For the full explanation, check out this ticket please (or i'll copy/paste)

https://github.com/dunglas/symfony-docker/pull/308

### DIFF
--- a/symfony/mailer/4.3/manifest.json
+++ b/symfony/mailer/4.3/manifest.json
@@ -14,5 +14,11 @@
             ]
         }
     },
+    "dockerfile": [
+        "RUN set -eux; \\",
+        "    install-php-extensions \\",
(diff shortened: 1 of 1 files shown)


maxhelias:
Shouldn't it be more on the package `twig/inky-extra` ?

stof:
or even better: on a recipe for the Inky package

fabpot:
It's already the case: see https://github.com/lorenzo/pinky/pull/10/files

maxhelias:
@fabpot This is more about installing the extension in the dockerfile with a recipe :)

fabpot:
But then, we need a standardized way of doing it.

PR #1139, opened 2022-10-24
Update DSNs for DBAL 3.5

derrabus:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | N/A

I've created a new recipe for DoctrineBundle and made the DSN examples ready for DBAL 4:

* Use the actual driver names as scheme instead of deprecated aliases.
* Added an example for MariaDB because sepcifying the server version correctly is a bit non-obvious here.
* Use full versions instead of short ones.
* Bumped Postgres to 15.0.

### DIFF
--- a/doctrine/doctrine-bundle/2.7/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/2.7/config/packages/doctrine.yaml
@@ -4,7 +4,7 @@ doctrine:
 
         # IMPORTANT: You MUST configure your server version,
         # either here or in the DATABASE_URL env var (see .env file)
-        #server_version: '13'
+        #server_version: '15'
     orm:
         auto_generate_proxy_classes: true
         naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
--- a/doctrine/doctrine-bundle/2.7/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/2.7/config/packages/doctrine.yaml
@@ -1,0 +1,42 @@
+doctrine:
+    dbal:
+        url: '%env(resolve:DATABASE_URL)%'
+
+        # IMPORTANT: You MUST configure your server version,
+        # either here or in the DATABASE_URL env var (see .env file)
+        #server_version: '13'
+    orm:
+        auto_generate_proxy_classes: true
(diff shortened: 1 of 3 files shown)


stof:
The `postgresql:` sheme is a standard set by Postgresql itself (the official `psql` CLI tool uses it as well, and PaaS platform are exposing credential URLs using that scheme). Is DBAL really deprecating that ? This does not make sense as it would forbid using PaaS env variables directly.

derrabus:
see doctrine/dbal#5697

stof:
@derrabus see my comments there, to see whether we still support standard database URLs when specifying a driver option separately or whether we are forced to migrate to DBAL-specific schemes (in which case this recipe update is not enough to keep compat with PaaS hosting providers as they won't change their scheme)

doctrine/doctrine-bundle/2.7/post-install.txt
```diff
@@ -0,0 +1,4 @@
+  * Modify your DATABASE_URL config in <fg=green>.env</>
+
+  * Configure the <fg=green>driver</> (postgresql) and
+    <fg=green>server_version</> (13) in <fg=green>config/packages/doctrine.yaml</>
```
fbourigault, line 4:
```suggestion
    <fg=green>server_version</> (15) in <fg=green>config/packages/doctrine.yaml</>
```

doctrine/doctrine-bundle/2.7/config/packages/doctrine.yaml
```diff
@@ -0,0 +1,42 @@
+doctrine:
+    dbal:
+        url: '%env(resolve:DATABASE_URL)%'
+
+        # IMPORTANT: You MUST configure your server version,
+        # either here or in the DATABASE_URL env var (see .env file)
+        #server_version: '13'
```
fbourigault, line 7:
```suggestion
        #server_version: '15'
```

fbourigault (commented):
I added a few suggestions to have a consistent database version in `doctrine.yaml`, `docker-compose.yml` and `post-install.txt`.

derrabus:
Good catch, thanks!

derrabus:
Closing in favor of doctrine/dbal#5843.

PR #1145, opened 2022-11-19
Use UUID v7 as default for new projects

ihmels:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

Symfony 6.2 introduces support for UUID v7, which is to be preferred over UUID v6 [according to the draft RFC][1]:

> Implementations SHOULD utilize UUID version 7 over UUID version 1 and
> 6 if possible.

In order not to change the UUID generation in existing projects, UUID v7 could be configured as default for new projects via this recipe.

[1]: https://www.ietf.org/archive/id/draft-peabody-dispatch-new-uuid-format-04.html#section-5.2-2

### DIFF
--- a/symfony/framework-bundle/6.2/config/packages/uid.yaml
+++ b/symfony/framework-bundle/6.2/config/packages/uid.yaml
@@ -1,0 +1,4 @@
+framework:
+    uid:
+        default_uuid_version: 7
+        time_based_uuid_version: 7


wouterj:
I'm not sure if we want to recommend a standard that is still in draft.

ihmels:
@wouterj Isn't UUID v6 also just being introduced in this RFC?

nicolas-grekas (approved):
Makes sense to me. v6 is the current default and is also a draft RFC, so moving to v7 would make sense (and v7 is really better).

OskarStark (approved):

derrabus:
This change breaks `symfony new --version=next myapp` because the UID component is not installed by default.

```
Symfony operations: 4 recipes (6e8494c3270e0afba2d0dbb3defc3de6)
  - Configuring symfony/flex (>=1.0): From github.com/symfony/recipes:main
  - Configuring symfony/framework-bundle (>=6.2): From github.com/symfony/recipes:main
  - Configuring symfony/console (>=5.3): From github.com/symfony/recipes:main
  - Configuring symfony/routing (>=6.2): From github.com/symfony/recipes:main
Executing script cache:clear [KO]
 [KO]
Script cache:clear returned with error code 1
!!
!!  In FrameworkExtension.php line 448:
!!
!!    Uid support cannot be enabled as the Uid component is not installed. Try ru
!!    nning "composer require symfony/uid".
!!
!!
!!
Script @auto-scripts was called via post-update-cmd
```

OskarStark:
Can we test this somehow to avoid such regressions in the future?

derrabus:
We can, but the corresponding CI workflow is broken at the moment because it tries to run with the v5 skeleton on PHP 7.4 first. See the CI on #1149.

fabpot:
This recipe must be moved to `symfony/uid` instead of `symfony/framework-bundle`, right?

stof:
@fabpot yes, and that was done in #1149

PR #1146, opened 2022-11-19
Add lock as an alias for symfony/lock

pierredup:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | N/A

Add `lock` as an alias for `symfony/lock`, so that you can use `composer req lock`.

### DIFF
--- a/symfony/lock/4.4/manifest.json
+++ b/symfony/lock/4.4/manifest.json
@@ -6,5 +6,6 @@
         "#1": "Choose one of the stores below",
         "#2": "LOCK_DSN=redis://localhost",
         "LOCK_DSN": "flock"
-    }
+    },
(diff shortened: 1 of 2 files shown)


chalasr (approved):
I thought `symfony/*` packages were automatically aliased 🤷‍♂️

kbond (approved):

stof:
This alias already exists thanks to the automatic aliasing: https://github.com/symfony/recipes/blob/7e2f8be50205ce92893bec7fa65479f370b6de9a/index.json#L149

pierredup:
@stof Then there is something else wrong, because I can't use `lock` as an alias

Screenshot 2022-11-21 at 13 05 17

PR #1168, opened 2023-01-22
Remove api-platform/core recipes

fck-ms:
A "composer require" of a library should never lead to changes in the configuration of an application. This is only acceptable behavior for bundles and packs. As api-platform/core is a library (according to the composer.json file) there should be no recipes.

### DIFF
--- a/api-platform/core/2.1/config/packages/api_platform.yaml
+++ b/api-platform/core/2.1/config/packages/api_platform.yaml
@@ -1,3 +1,0 @@
-api_platform:
-    mapping:
-        paths: ['%kernel.project_dir%/src/Entity']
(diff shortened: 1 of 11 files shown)


chalasr:
Actually, the recipe wires the Symfony bundle provided by (and embedded in) the api-platform/core package. Cllosing for this reason. Thanks

fck-ms:
Then the bundle should be a separate package and recipes should be added for the bundle package. At this moment the api-platform/core package is published as a library and there shouldn't be any recipes for library packages.

chalasr:
Possibly yes, please consider discussing this on the api-platform/core repository.

fck-ms:
https://github.com/api-platform/core/issues/5357

PR #1171, opened 2023-01-25
[symfony/ux-live-component] Update config file

AntoineAresu:
| Q             | A
| ------------- | ---
| License       | MIT

Fix an error I have experienced : An exception has been thrown during the rendering of a template ("Unable to generate a URL for the named route "ux_live_component" as such route does not exist.").

symfony 6.1
symfony/ux-live-component 2.6.1
@symfony/stimulus-bridge 3.2.0

### DIFF
--- a/symfony/ux-live-component/2.6/config/routes/ux_live_component.yaml
+++ b/symfony/ux-live-component/2.6/config/routes/ux_live_component.yaml
@@ -1,4 +1,5 @@
 live_component:
+    type: php
     resource: '@LiveComponentBundle/config/routes.php'
     prefix: '/_components'
     # adjust prefix to add localization to your components


weaverryan (approved):
I'm not sure why this would be needed in some cases (I've never needed it), but it's also harmless :)

fabpot:
Let's understand why it's needed under some circumstances. If we can fix the route cause, that's better.

weaverryan:
@AntoineAresu Can you give some info about when this happens? What Symfony version you have? What packages you have installed?

AntoineAresu:
Sorry this is my first PR on an official repository.

It happens after upgrading `[symfony/ux-live-component]` from 2.2 to 2.5
I tried to upgrade from 2.5 to 2.6 to see if it could resolve the situation, but only the update of this config file fixed it.

Like I said, the project run under Symfony 6.1.
symfony/ux-live-component 2.6
@symfony/stimulus-bridge 3.2.0

But i can try to create a bug reproducer if it is needed

weaverryan:
Try also doing a `composer recipes:update symfony/ux-live-component`. In 2.6, we changed the location of the routing file - you can see how the new route import should look here: https://github.com/symfony/recipes/blob/main/symfony/ux-live-component/2.6/config/routes/ux_live_component.yaml

> Sorry this is my first PR on an official repository.

No worries - happy to have you here :)

AntoineAresu:
I did update de recipe before updating the config file with `type: php`.
The location of the routing file was ok, and the file existed in this location, but it wasn't read.

Unfortunately, i didn't succeed in creating a reproducer so far.

weaverryan:
I'll close for now. If someone else hits this, they'll hopefully ind this issue and can add their situation.

Cheers!

PR #1174, opened 2023-02-06
[phpunit] limited support for PHPUnit 10

jrushlow:
| Q             | A
| ------------- | ---
| License       | MIT


symfony/phpunit-bridge/5.3/bin/phpunit
```diff
@@ -6,9 +6,15 @@ if (!ini_get('date.timezone')) {
...
-    require PHPUNIT_COMPOSER_INSTALL;
-    PHPUnit\TextUI\Command::main();
+    if (class_exists(PHPUnit\TextUI\Command::class)) {
+        // PHPUnit < 9.x
+        define('PHPUNIT_COMPOSER_INSTALL', dirname(__DIR__).'/vendor/autoload.php');
+        require PHPUNIT_COMPOSER_INSTALL;
+        PHPUnit\TextUI\Command::main();
+    } else {
+        // PHPUnit >= 10.0
+        require dirname(__DIR__).'/vendor/phpunit/phpunit/phpunit';
```
nicolas-grekas, line 16: the bridge is not compatible with phpunit 10.
I'd suggest removing that part for now at least

weaverryan: I'm wondering about a 3rd option :).

Is there a reason why we called `PHPUnit\TextUI\Command::main()` for PHPUnit 9.5 vs requiring the executable? Could this new code be used in ALL cases?

> the bridge is not compatible with phpunit 10.

I understand this... but since you CAN easily install `phpunit-bridge` and PHPUnit 10, if there is a straightforward way to get this file working today, why wait until tomorrow to do it? :)

curry684: > the bridge is not compatible with phpunit 10.

I would argue that at all times the `phpunit-bridge` should preventively declare a conflict with future major versions of PHPunit until tested/adapted, it already declares some:
```
    "conflict": {
        "phpunit/phpunit": "<7.5|9.1.2"
    },
```
I'd suggest adding `|>=11` once v10 support is implemented (won't make any difference now as Composer will just install the older version not declaring the conflict).

Right now running `symfony new test` followed by `composer req symfony/phpunit-bridge phpunit/phpunit` on a new project will create a broken project.

CRC-Mismatch:
@nicolas-grekas @weaverryan @curry684 I'm joining late, and uncalled, but... I'm pretty sure now something has already changed related to PHPUnit 10.x (at least in regards to Symfony 5.4 and 6.2, unsure about 6.3 for the moment...)

nicolas-grekas:
let me close as stalled, also because a lot changed on the topic recently so this may be outdated

PR #1184, opened 2023-03-16
phpunit-bridge[phpunit.xml.dist]: use dynamic xsd managed by phpunit-bridge

kaznovac:
phpunit-bridge manages phpunit.xsd in the folder `vendor/bin/.phpunit/phpunit.xsd` this allows config to be verified against schema corresponding to the `SYMFONY_PHPUNIT_VERSION` version specified


maxhelias:
This is not compatible with the default behavior of test-pack: https://packagist.org/packages/symfony/test-pack, which installs phpunit directly, so the xsd is available.

PR #1189, opened 2023-04-03
[zenstruck/messenger-test] add recipe

kbond:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

I think this package is well-used enough to justify the recipe here.


OskarStark:
I think it's better to just register the bundle in `test` and not `dev`, thanks @kbond 👍

fabpot (approved):

Issue #1173, opened 2023-02-06
Class "PHPUnit\TextUI\Command" not found for phpunit/phpunit:10


valodya-street:
https://github.com/symfony/recipes/blob/b1176e36c42b0738c1475960202d7e074863ca3c/symfony/phpunit-bridge/5.3/bin/phpunit#L11

weaverryan:
Yup, I'm guessing that file needs an update for phpunit 10

stof:
The Symfony phpunit bridge does not support PHPUnit 10 yet. This is known and tracked at https://github.com/symfony/symfony/issues/49069

nicolas-grekas:
Fixed by https://github.com/symfony/test-pack/pull/19 in practice.

manuelcanga:
This can be mixed if you replace `PHPUnit\TextUI\Command::main()' to:

```php
(new \PHPUnit\TextUI\Application())->run($GLOBALS['argv']);
```

More details: https://dev.to/manuelcanga/fix-symfony-6x-tests-with-phpunit-10-18a1

melroy89:
Yes so actually it's with `exit()`:

```php
exit((new \PHPUnit\TextUI\Application())->run($GLOBALS['argv']));
```

derrabus:
See #1249

melroy89:
Interesting my composer `phpunit/phpunit` recipe was up to date. And this code you showed me is not working with PHPUnit v10.. 

` PHPUnit\TextUI\Command` does not exists anymore in PHPUnit v10.

You will get:

```
$ ./bin/phpunit tests/Unit
PHP Fatal error:  Uncaught Error: Class "PHPUnit\TextUI\Command" not found in /blabla/balbalba/project/bin/phpunit:11
Stack trace:
#0 {main}
  thrown in /blab/blbla/project/bin/phpunit on line 11
  ```

vinceAmstoutz:
> @derrabus Interesting my composer `phpunit/phpunit` recipe was up to date. And this code you showed me is not working with PHPUnit v10..
> 
> ` PHPUnit\TextUI\Command` does not exists anymore in PHPUnit v10.
> 
> You will get:
> 
> ```
> $ ./bin/phpunit tests/Unit
> PHP Fatal error:  Uncaught Error: Class "PHPUnit\TextUI\Command" not found in /blabla/balbalba/project/bin/phpunit:11
> Stack trace:
> #0 {main}
[...]

PR #1195, opened 2023-05-03
[symfony/phpunit-bridge] run PHPUnit by it's official entrypoint

kaznovac:
Run the PHPUnit script by it's own entrypoint - so this script does not depend on the internal class

| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | N/A

Resolves https://github.com/symfony/recipes/issues/1173


maxhelias:
This is not the right solution. See the ticket mentioned #1173 and https://github.com/symfony/symfony/pull/50371

kaznovac:
@maxhelias see https://github.com/symfony/recipes/pull/1249

PR #1196, opened 2023-05-03
[symfony/ux-translator] TranslatorBundle class now prefixed by Ux

r-huyvaert:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | N/A

The class TranslatorBundle of the Symfony UX Translator component has recently been renamed to UxTranslatorBundle (Commit [bc96739](https://github.com/symfony/ux-translator/commit/bc96739e03cc6312f7b49cab4cb7c7060c3e4999)).

Currenty, the following line added to the bundles.php file is as the following :

```Symfony\UX\Translator\TranslatorBundle::class => ['all' => true],```

It needs to be renamed here too so it won't lead to the following error after the install :

image

### DIFF
--- a/symfony/ux-translator/2.7/manifest.json
+++ b/symfony/ux-translator/2.7/manifest.json
@@ -1,6 +1,6 @@
 {
     "bundles": {
-        "Symfony\\UX\\Translator\\TranslatorBundle": ["all"]
+        "Symfony\\UX\\Translator\\UxTranslatorBundle": ["all"]
     },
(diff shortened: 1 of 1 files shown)


Kocal:
Thanks!

This was modified to prevent [bundle loading issue with the Symfony Translator](https://github.com/symfony/ux/pull/819), but I forgot to modify it in #1185 which was merged with a broken CI.

stof (approved):

PR #1198, opened 2023-05-03
Prefer php-http/discovery over nyholm/psr7 when installed

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Note that php-http/discovery itself defaults to nyholm/psr7 so in practice this shouldn't change much, except for people that want to rely on another PSR-17 implementation.

### DIFF
--- a/php-http/discovery/1.18/config/packages/http_discovery.yaml
+++ b/php-http/discovery/1.18/config/packages/http_discovery.yaml
@@ -1,0 +1,10 @@
+services:
+    Psr\Http\Message\RequestFactoryInterface: '@http_discovery.psr17_factory'
+    Psr\Http\Message\ResponseFactoryInterface: '@http_discovery.psr17_factory'
+    Psr\Http\Message\ServerRequestFactoryInterface: '@http_discovery.psr17_factory'
+    Psr\Http\Message\StreamFactoryInterface: '@http_discovery.psr17_factory'
+    Psr\Http\Message\UploadedFileFactoryInterface: '@http_discovery.psr17_factory'
+    Psr\Http\Message\UriFactoryInterface: '@http_discovery.psr17_factory'
+
+    http_discovery.psr17_factory:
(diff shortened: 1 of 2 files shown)


stof (approved):

fabpot (approved):

tacman:
The recipe throws a warning about not being installable.

> Skipping recipe for nyholm/psr7: all versions of the recipe conflict with your package versions.

```bash
~/g/museado/museado$ composer why nyholm/psr7
php-http/discovery 1.19.4 conflicts nyholm/psr7 (<1.0)   
toshy/bunnynet-php 4.5.0  requires  nyholm/psr7 (^1.6.1)
 
~/g/museado/museado$ composer why php-http/discovery
knplabs/github-api                v3.14.1 requires  php-http/discovery (^1.12) 
meilisearch/meilisearch-php       v1.10.0 requires  php-http/discovery (^1.7)  
php-http/multipart-stream-builder 1.4.2   requires  php-http/discovery (^1.15) 
symfony/http-client               v7.1.4  conflicts php-http/discovery (<1.15) 
symfony/psr-http-message-bridge   v7.1.4  conflicts php-http/discovery (<1.15) 
```

I got an explanation here: https://github.com/ToshY/BunnyNet-PHP/discussions/141

but is there a way to remove the warning?  It's not causing any harm, of course.

derrabus:
Is this a warning? I read this as: FYI, I didn't install a recipe for this package and here's the reason why.

tacman:
Yes, it's not a warning, but it's a message that says "I didn't do something because there was a conflict". So I'd love to avoid seeing that message, I think it happens on every deployment.

nicolas-grekas:
I don't have a solution for skipping this warning. You can just ignore it for now.
We don't want this recipe installed when php-http/discovery is also installed, and that's the way to do it...

tacman:
ah, that makes more sense.  Thanks.

PR #1199, opened 2023-05-09
Fix port matching on postgresql docker

Nek-:
| Q             | A
| ------------- | ---
| License       | MIT

By running the default configuration (tested under macos, m1 pro) it binds to the wrong port (60057->5432/tcp) I have no idea why, but this fixes the issue and is compatible with any other installation.

_side note: the issue seems to be the exact same with mailcatcher_


fabpot:
The current code is correct. It means we want to expose the psql port from the container to a random port on the host. That allows you to have many projects using the same services simultaneously as they are using different ports. If you are using the Symfony CLI, it automatically exposes the correct env vars. See https://symfony.com/doc/current/setup/symfony_server.html#symfony-server-docker

Nek-:
@fabpot I'm hitting this issue for every installations so here I am pushing it. My point is that it also makes it hard to connect a client to the running database. (while occupied port is a trivial fix, random port is a bit wild)

stof:
@Nek- when you use `symfony run`, the Symfony CLI automatically injects the appropriate env variables (including the env variables recognized by standard tools like the `pgsql` CLI client)

PR #1206, opened 2023-05-30
[doctrine-bundle] leverage install-php-extensions

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

As recommended in the README of the official Docker image for PHP, [Symfony Docker](https://github.com/dunglas/symfony-docker) as well as [API Platform](https://api-platform.com) now come with [`mlocati/docker-php-extension-installer`](https://github.com/mlocati/docker-php-extension-installer) preinstalled.

This allows us to simplify and make more reliable our Doctrine Bundle recipe.

### DIFF
--- a/doctrine/doctrine-bundle/1.12/manifest.json
+++ b/doctrine/doctrine-bundle/1.12/manifest.json
@@ -15,7 +15,10 @@
         "DATABASE_URL": "postgresql://app:!ChangeMe!@127.0.0.1:5432/db_name?serverVersion=15&charset=utf8"
     },
     "dockerfile": [
-        "RUN install-php-extensions pdo_pgsql"
+        "RUN apk add --no-cache --virtual .pgsql-deps postgresql-dev && \\",
+        "\tdocker-php-ext-install -j$(nproc) pdo_pgsql && \\",
+        "\tapk add --no-cache --virtual .pgsql-rundeps so:libpq.so.5 && \\",
+        "\tapk del .pgsql-deps"
     ],
     "docker-compose": {
         "docker-compose.yml": {
--- a/doctrine/doctrine-bundle/1.12/manifest.json
+++ b/doctrine/doctrine-bundle/1.12/manifest.json
@@ -15,10 +15,7 @@
         "DATABASE_URL": "postgresql://app:!ChangeMe!@127.0.0.1:5432/db_name?serverVersion=15&charset=utf8"
     },
     "dockerfile": [
-        "RUN apk add --no-cache --virtual .pgsql-deps postgresql-dev && \\",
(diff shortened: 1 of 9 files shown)


weaverryan:
Do we have basically assume that users will be starting with https://github.com/dunglas/symfony-docker? I'm trying to think of a case where a user is spinning up their own Dockerfile. But that may simply be a case we are *not* covering on purpose.

What about a recipe update? Will this update their Dockerfile... but perhaps the new `mlocati/docker-php-extension-installer` won't be in there (since that didn't come from a recipe originally iirc)?

I'm for this change - just thinking about the rough cases.

dunglas:
> Do we have basically assume that users will be starting with https://github.com/dunglas/symfony-docker? I'm trying to think of a case where a user is spinning up their own Dockerfile. But that may simply be a case we are not covering on purpose.

Not really. The Docker configurator has always made the assumption that the `Dockerfile` comes from Symfony Docker, API Platform or at least something quite similar.

Especially, the markers required for this configurator to be executed must be present and at an appropriate position in the `Dockerfile` for the recipe to be executed properly, and AFAIK only Symfony Docker and its derivatives provide theses markers.

However, in this specific case, the recipe will work for any `Dockerfile` having `mlocati/docker-php-extension-installer` installed, which is very common and popular (this script is even recommended in the README of the official PHP Docker image).

Actually, this patch makes the Doctrine recipe **less** dependent on Symfony Docker because `mlocati/docker-php-extension-installer` also supports Debian-based images (Symfony Docker only uses Alpine, and our current script only supports Alpine too). 

> What about a recipe update?

Indeed, that will cause issues when upgrading the recipe. The user will have to add `mlocati/docker-php-extension-installer` (which is a single line fix).

This script has been added a while ago in the `Dockerfile`s I maintain, so the number of impacted users impacted should be low, but we should document this change because this issue will definitely occurs.

On the other hand, the benefit is that it will be a BC break once and for all because the script is well maintained, while our current shell code can break at any moment if Alpine changes the way it packages libpq, if a new major version of libpq is released, etc etc

WDYT?

weaverryan:
Well explained. It's good with me, but could we only change it in the latest version of the recipe? Or, better, in a new 2.9 version? I think, in general, we've tried to adopt a policy of not updating old recipes versions if we can avoid it.

dunglas:
@weaverryan good idea. Updated to use the script only for 2.9.

dunglas:
Can we merge this?

nicolas-grekas (approved):

maxhelias (approved):

fabpot (approved):

PR #1227, opened 2023-08-04
[api-platform/core] prevent a deprecation warning and sync with upstream

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

Enabling the standard behavior for PUT requests is necessary to prevent a deprecation warning.
I also imported the defaults [we set in the API Platform distribution](https://github.com/api-platform/api-platform/blob/main/api/config/packages/api_platform.yaml).

### DIFF
--- a/api-platform/core/3.1/config/packages/api_platform.yaml
+++ b/api-platform/core/3.1/config/packages/api_platform.yaml
@@ -1,0 +1,10 @@
+api_platform:
+    title: Hello API Platform
+    version: 1.0.0
+    # Good defaults for REST APIs
+    defaults:
+        stateless: true
+        cache_headers:
+            vary: ['Content-Type', 'Authorization', 'Origin']
+        extra_properties:
+            standard_put: true
--- a/api-platform/core/2.1/post-install.txt
+++ b/api-platform/core/2.1/post-install.txt
@@ -1,8 +1,10 @@
   * Your API is almost ready:
-    1. Create your first API resource in <info>src/Entity</info>;
+    1. Create your first API resource in <info>src/ApiResource</info>;
     2. Go to <info>/api</info> to browse your API
+
+  * Using MakerBundle? Try <info>php bin/console make:entity --api-resource</info> 
 
   * To enable the GraphQL support, run <comment>composer require webonyx/graphql-php</>,
     then browse <info>/api/graphql</info>.
 
-  * <fg=blue>Read</> the documentation at <comment>https://api-platform.com/docs</>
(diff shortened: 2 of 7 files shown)


api-platform/core/3.1/config/packages/api_platform.yaml
```diff
@@ -0,0 +1,10 @@
+api_platform:
+    title: Hello API Platform
+    version: 1.0.0
+    # Good defaults for REST APIs
+    defaults:
+        stateless: true
```
weaverryan, line 6: This might be too opinionated? If you create a frontend on the same domain that logs in via cookies, then authentication wouldn't work, correct?

dunglas: Cookies can be used, but the session cannot. It's one of [the mandatory constraints of the REST architecture](https://www.ics.uci.edu/%7Efielding/pubs/dissertation/rest_arch_style.htm#sec_5_1_3).

IMHO it's not opinionated, it's just encouraging the correct way of creating REST APIs (and web applications in general).

Better let the user remove this line only if they know what they are doing (and by doing that they are probably doing a mistake) than having them creating non-stateless "REST" APIs.

dunglas: By the way since 6.3 stateless routes automatically enable stateless firewalls, so enabling this feature shouldn't be an issue even when using the default config of firewalls: https://symfony.com/doc/current/reference/configuration/security.html#stateless

stof: but if this is the good thing to do, why isn't it the default in ApiPlatform then ?

dunglas: @stof it is: https://github.com/api-platform/api-platform/blob/main/api/config/packages/api_platform.yaml

stof: I'm talking about the default in the Configuration class of the bundle (i.e. the actual default value of the setting), not the config provided in the API platform distribution.

nicolas-grekas (approved):

api-platform/core/3.1/config/packages/api_platform.yaml
dunglas: Because this would be a BC break. We'll maybe do that for API Platform 4 next year.

weaverryan (approved):

PR #1229, opened 2023-08-08
Recommend NPM instead of Yarn

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

/cc @weaverryan @dunglas

### DIFF
--- a/api-platform/admin-pack/1.0/post-install.txt
+++ b/api-platform/admin-pack/1.0/post-install.txt
@@ -1,5 +1,5 @@
   * Your admin is almost ready:
-    1. Install the JavaScript dependencies by running <comment>yarn add @api-platform/admin @babel/preset-react</comment>
+    1. Install the JavaScript dependencies by running <comment>npm add @api-platform/admin @babel/preset-react</comment>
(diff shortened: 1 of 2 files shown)


wouterj (approved):

dunglas (approved):

melroy89:
Just for info. I would like to see the demo also move to NPM (package-lock.json): https://github.com/symfony/demo

stollr:
I'm curious as to why the recommendation was changed from Yarn to NPM. Are there any information?

stof:
Back when we started this recommendation (when introducing webpack-encore), yarn was strictly superior to npm, mostly thanks to having a proper lockfile (at that time, npm did not have package-lock.json but only the shrinkwrap mechanism) and being faster.
Since then, npm has caught up on those fronts. And it has the big benefit of being installed by default when you install node, which requires 1 less step.
Note however that webpack-encore is not dropping support for yarn (it tests compatibility with several package managers) and so you can keep using yarn if you prefer (or pnpm for that matter)

stollr:
This makes sense. Thanks for your explanation.

PR #1232, opened 2023-08-11
[FrameworkBundle] Remove router and translator default values

Jean-Beru:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  |

This PR removes the `framework.default_locale`,  `framework.translator.fallback` and  `framework.router.utf8` options since it's already their default value.

### DIFF
--- a/symfony/routing/6.0/config/packages/routing.yaml
+++ b/symfony/routing/6.0/config/packages/routing.yaml
@@ -1,7 +1,5 @@
 framework:
     router:
-        utf8: true
-
         # Configure how to generate URLs in non-HTTP contexts, such as CLI commands.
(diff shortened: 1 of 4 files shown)


fabpot:
I would be very careful when removing entries that only define default values.
These entries are very useful as they help developers change the defaults without looking at the docs.
@nicolas-grekas

nicolas-grekas:
I'd suggest removing only the UTF-8 one indeed

nicolas-grekas (changes requested):
please revert *all* changes and create a new recipe for 7.0 from the 6.2 one that removes only the `utf8` option. Recipes for older versions must have the `utf8` option explicitly.

PR #1234, opened 2023-08-14
[SecurityBundle] Remove config for password_hashers

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Needs https://github.com/symfony/symfony/pull/51380


chalasr:
As discussed in https://github.com/symfony/symfony/pull/51380 and privately I'm rather 👎 for this PR as it hides an important part of Symfony's security. The syntax and meaning of the `{ type : alg }` mapping are hard to guess and, in addition to the doc link that is removed by this patch, it provides a lot of information about how password hashing works in Symfony.

PR #1240, opened 2023-09-15
[api-platform/core] sane defaults for API Platform 3.2

soyuka:
| Q             | A
| ------------- | ---
| License       | MIT

In API Platform 4 `json` will be disabled by default so it's now recommended to explicit formats.
For `event_listeners_backward_compatibility_layer` see https://github.com/api-platform/core/pull/5657
We're using the Symfony Inflector now but this will probably break current code so we need a flag to `keep_legacy_inflector`
The `rfc_7807_compliant_errors` is to make Hydra errors compatible with the [JSON Problem](https://datatracker.ietf.org/doc/rfc7807/) specification. This may lead to unexpected error fields and therefore we need a compatibility flag.

### DIFF
--- a/api-platform/core/3.2/config/packages/api_platform.yaml
+++ b/api-platform/core/3.2/config/packages/api_platform.yaml
@@ -1,0 +1,18 @@
+api_platform:
+    title: Hello API Platform
+    version: 1.0.0
+    formats:
+        jsonld: ['application/ld+json']
+    doc_formats:
+        jsonld: ['application/ld+json']
+        jsonopenapi: ['application/vnd.openapi+json']
+        html: ['text/html']
+    defaults:
+        stateless: true
+        cache_headers:
+            vary: ['Content-Type', 'Authorization', 'Origin']
+        extra_properties:
+            standard_put: true
+            rfc_7807_compliant_errors: true
+    event_listeners_backward_compatibility_layer: false
+    keep_legacy_inflector: false
--- a/api-platform/core/3.2/config/routes/api_platform.yaml
+++ b/api-platform/core/3.2/config/routes/api_platform.yaml
@@ -1,0 +1,4 @@
+api_platform:
+    resource: .
+    type: api_platform
+    prefix: /api
--- a/api-platform/core/3.2/manifest.json
+++ b/api-platform/core/3.2/manifest.json
@@ -1,0 +1,9 @@
+{
+    "bundles": {
+        "ApiPlatform\\Symfony\\Bundle\\ApiPlatformBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/",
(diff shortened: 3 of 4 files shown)


api-platform/core/3.2/config/packages/api_platform.yaml
```diff
@@ -0,0 +1,26 @@
...
+        stateless: true
+        cache_headers:
+            vary:
+                - Content-Type
+                - Authorization
+                - Origin
+        extra_properties:
+            standard_put: true
+            rfc_7807_compliant_errors: true
+
```
dunglas, line 26:
```suggestion
```

api-platform/core/3.2/config/packages/api_platform.yaml
```diff
@@ -0,0 +1,26 @@
...
+        - text/html
+    event_listeners_backward_compatibility_layer: false
+    keep_legacy_inflector: false
+    defaults:
+        stateless: true
+        cache_headers:
+            vary:
+                - Content-Type
+                - Authorization
+                - Origin
```
dunglas, line 22:
```suggestion
    formats:
        jsonld: ['application/ld+json']
    doc_formats:
        jsonld: ['application/ld+json']
        html: ['text/html']
    event_listeners_backward_compatibility_layer: false
    keep_legacy_inflector: false
    defaults:
        stateless: true
        cache_headers:
            vary: ['Content-Type', 'Authorization', 'Origin']
```

(JSON:API docs format shouldn't be enabled by default)

dunglas (changes requested):

api-platform/core/3.2/config/packages/api_platform.yaml
soyuka: and openapi in its JSON format?

dunglas: Sorry my bad, indeed `application/vnd.openapi+json` should be enabled. I read too quickly and confused it with JSON:API.

api-platform/core/3.2/config/packages/api_platform.yaml
```diff
@@ -0,0 +1,18 @@
...
+    title: Hello API Platform
+    version: 1.0.0
+    formats:
+        jsonld: ['application/ld+json']
+    doc_formats:
+        jsonld: ['application/ld+json']
+        html: ['text/html']
+        jsonopenapi: ['application/vnd.openapi+json']
+    event_listeners_backward_compatibility_layer: false
+    keep_legacy_inflector: false
```
dunglas, line 11: Maybe could we but these two lines at the end, as they will disappear in v4 and are less relevant for end users?

dunglas (dismissed):

nicolas-grekas (approved):

dunglas (approved):

chalasr (approved):

PR #1246, opened 2023-10-11
[UX Vue] Add lines for components lazy-loading

Kocal:
| Q             | A
| ------------- | ---
| License       | MIT

Hi, this PR follows https://github.com/symfony/ux/issues/1177#issuecomment-1757659579. 

The added lines were removed from the documentation, but I really think we should still document the lazy-loading behavior of Vue controller components (thanks to `'lazy'` parameter).

cc @weaverryan

### DIFF
--- a/symfony/ux-vue/2.9/manifest.json
+++ b/symfony/ux-vue/2.9/manifest.json
@@ -31,6 +31,27 @@
         },
         {
             "file": "assets/app.js",
+            "content": "// If you prefer to lazy-load your Vue.js controller components, in order to reduce the JavaScript bundle the smallest as possible,",
+            "position": "bottom",
+            "warn_if_missing": true,
+            "requires": "symfony/webpack-encore-bundle"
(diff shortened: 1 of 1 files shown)


symfony/ux-vue/2.9/manifest.json
```diff
@@ -29,6 +29,27 @@
...
+            "warn_if_missing": true,
+            "requires": "symfony/webpack-encore-bundle"
+        },
+        {
+            "file": "assets/app.js",
+            "content": "registerVueControllerComponents(require.context('./vue/controllers', true, /\\.vue$/, 'lazy'));",
+            "position": "bottom",
+            "warn_if_missing": true,
+            "requires": "symfony/webpack-encore-bundle"
+        },
```
weaverryan, line 52: Thanks for opening - but I think I'd rather mention this in the docs - vs adding 4 lines to the user's app.js

PR #1249, opened 2023-11-05
Fix bin/phpunit for PHPUnit 10

### LINKED ISSUE #1248: Class "PHPUnit\TextUI\Command" not found
### Symfony version(s) affected

v6.3

### Description

When I execute `php bin/phpunit` I get the following exception:

```console
$ php bin/phpunit
PHP Fatal error:  Uncaught Error: Class "PHPUnit\TextUI\Command" not found in /home/xxxxx/xxxxx/xxxxx/bin/phpunit:11
Stack trace:
#0 {main}
  thrown in /home/xxxxx/xxxxx/xxxxx/bin/phpunit on line 11
```

### How to reproduce

Here's what I did: 

1. I created a new project with **Symfony** command: `symfony new my-project`.
2. After some days of work I installed **test-pack** with `composer require test`.
3. Then I followed terminal instructions and created a new test `bin/console make:test`.
4. Finally I tried to execute the test with `php bin/phpunit`. Fatal error raises.

### Possible Solution

Well, as the error suggests `PHPUnit\TextUI\Command` does not exists, so I checked `composer.json` and  **PHPUnit v10.4** was installed, so I downgraded to **PHPUnit v9** :

```console
$ composer require phpunit/phpunit:^9 --dev -W
```

And the error was fixed.

### Additional Context

In this specific project **PHPUnit v10** is always installed, I tried to reproduce the error in a fresh project and **PHPUnit v9** was installed.

Anyway, after a quick  revision, I noticed that **PHPUnit** version is never specified by Symfony, IMHO this should be done in `symfony/phpunit-bridge` to avoid installing **PHPUnit v10** since `PHPUnit\TextUI\Command` exists no more.

Thank you :)

derrabus:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | N/A

Fixes #1248.

Our `bin/phpunit` calls the PHPUnit from the vendor folder, if we use Composer to install PHPUnit. Instead of calling `require` on PHPUnit's entrypoint directly, we call internal PHPUnit classes to bootstrap the CLI runner.

The reason for that was that PHPUnit's bin script contains a shebang (`#!/usr/bin/env php`) which would be output on PHP 7. On PHP 8 however, that limitation is gone and with it the reason for our workaround. Calling `require` on PHPUnit's entrypoint makes the bin/phpunit script more resilient against future internal changes in PHPUnit and allows the script to function with PHPUnit 10.

### DIFF
--- a/symfony/phpunit-bridge/5.3/bin/phpunit
+++ b/symfony/phpunit-bridge/5.3/bin/phpunit
@@ -6,9 +6,13 @@ if (!ini_get('date.timezone')) {
 }
 
 if (is_file(dirname(__DIR__).'/vendor/phpunit/phpunit/phpunit')) {
-    define('PHPUNIT_COMPOSER_INSTALL', dirname(__DIR__).'/vendor/autoload.php');
-    require PHPUNIT_COMPOSER_INSTALL;
-    PHPUnit\TextUI\Command::main();
+    if (PHP_VERSION_ID >= 80000) {
+        require dirname(__DIR__).'/vendor/phpunit/phpunit/phpunit';
+    } else {
+        define('PHPUNIT_COMPOSER_INSTALL', dirname(__DIR__).'/vendor/autoload.php');
+        require PHPUNIT_COMPOSER_INSTALL;
+        PHPUnit\TextUI\Command::main();
+    }
 } else {
     if (!is_file(dirname(__DIR__).'/vendor/symfony/phpunit-bridge/bin/simple-phpunit.php')) {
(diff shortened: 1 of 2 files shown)


stof (approved):

xabbuh (approved):

melroy89:
Despite this PR merged into the `main` branch. The changes seem to be lost: https://github.com/derrabus/recipes/blob/main/symfony/phpunit-bridge/6.3/bin/phpunit

PR #1253, opened 2023-11-09
[FrameworkBundle] Improve require to require_once for all maintained Symfony versions

AurelienPillevesse:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | None

Improving `require` to `require_once` for all maintained Symfony versions (5.4, 6.4 and 7.0).

### DIFF
--- a/symfony/framework-bundle/5.4/config/preload.php
+++ b/symfony/framework-bundle/5.4/config/preload.php
@@ -1,5 +1,5 @@
 <?php
 
 if (file_exists(dirname(__DIR__).'/var/cache/prod/App_KernelProdContainer.preload.php')) {
-    require dirname(__DIR__).'/var/cache/prod/App_KernelProdContainer.preload.php';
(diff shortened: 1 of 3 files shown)


bocharsky-bw (commented):
What's the improvement on this? Like, what benefits using `require_once` vs `require`?

I think we have `require` here on purpose.

AurelienPillevesse:
When I added Sonar to analyze my project. I looked without vendor folder restriction and it suggested to change it from `require` to `require_once` so I created a PR in case it was a mistake. If not, we can close it.

bocharsky-bw:
In theory that file should be included only once... if for some reason it's included twice - you will see a clear error because that's an architectural error. In such cases `require` make life simpler, unless we're talking about requiring your custom classes when it may happen

stof:
Well, the question is what does it actually solve, not whether Sonar just reported that without any real-world impact.

weaverryan:
Thanks for opening this!

I agree - unless this is solving a real issue, there's not a compelling enough reason to make this change.

Cheers!

PR #1254, opened 2023-11-09
[CI] Update CI for 'Create-project with skeleton ^6' step

AurelienPillevesse:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | None

Update CI for 'Create-project with skeleton ^6' step.

### DIFF
--- a/.github/workflows/callable-qa.yml
+++ b/.github/workflows/callable-qa.yml
@@ -269,12 +269,12 @@ jobs:
                     exit $EXIT_CODE
 
             -
-                name: Setup PHP 8.1
+                name: Setup PHP 8.2
                 if: "always() && steps.config.outcome == 'success'"
                 uses: shivammathur/setup-php@v2
                 with:
                     coverage: "none"
-                    php-version: "8.1"
+                    php-version: "8.2"
 
             -
                 name: Create-project with skeleton ^6


AurelienPillevesse:
Don't really understand why CI is failing right now.
Wanted to fix one of the issues visible in this one : https://github.com/symfony/recipes/pull/1253 (caused by PHP version)

AurelienPillevesse:
image

About the other error, I can see that symfony/symfony ^5.4 is downloading symfony/framework-bundle ^7.0.
Do you know if it's normal or not, I don't know enough this part ? @nicolas-grekas @fabpot

AurelienPillevesse:
My bad, I saw that it's Symfony 7.0 which will need PHP 8.2. Not the version ^6.

weaverryan:
> About the other error in the CI, I can see that symfony/symfony ^5.4 is downloading symfony/framework-bundle ^7.0.
Do you know if it's normal or not, I don't know enough this part ?

I'm not sure. It comes from the `composer require -W --ansi $PACKAGES` line in the CI. That `$PACKAGES` I believe is an attempt to fetch the packages affected by the PR - https://github.com/symfony/recipes/pull/1254/files#diff-a63af90bcded38e234e48f5a253229ba26cfbcef5271976a3e39a8c8b3c86894R239 - and if the 7.0 version of `symfony/framework-bundle` is affected by the recipe, then it tries to install it. I'm not really sure how to fix that - we'd need to know which packages are / are not compatible with different Symfony versions. So, this is tricky - and maybe we are ok with it and just let the CI be a helpful guide, not something that always needs to pass.

PR #1256, opened 2023-11-16
Add TwigBundle 6.4 with `file_name_pattern: *.twig`

GromNaN:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

By default, Twig templates have the `.twig` extension. They can be collocated with other files (SVG, CSS, JS or even [PHP for Twig Components](https://twitter.com/weaverryan/status/1724827396279357498))

Adding the option `twig.file_name_pattern: '*.twig'` by default for new projects is necessary to avoid compiling non-twig files.

Related to symfony/symfony#45845

### DIFF
--- a/symfony/twig-bundle/6.4/config/packages/twig.yaml
+++ b/symfony/twig-bundle/6.4/config/packages/twig.yaml
@@ -1,5 +1,4 @@
 twig:
-    default_path: '%kernel.project_dir%/templates'
     file_name_pattern: '*.twig'
 
 when@test:
--- a/symfony/twig-bundle/6.4/config/packages/twig.yaml
+++ b/symfony/twig-bundle/6.4/config/packages/twig.yaml
@@ -1,0 +1,7 @@
+twig:
+    default_path: '%kernel.project_dir%/templates'
+    file_name_pattern: '*.twig'
+
+when@test:
+    twig:
+        strict_variables: true
--- a/symfony/twig-bundle/6.4/manifest.json
+++ b/symfony/twig-bundle/6.4/manifest.json
@@ -1,0 +1,12 @@
+{
+    "bundles": {
+        "Symfony\\Bundle\\TwigBundle\\TwigBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/",
+        "templates/": "templates/"
+    },
+    "conflict": {
+        "symfony/framework-bundle": "<5.3"
+    }
+}
--- a/symfony/twig-bundle/6.4/templates/base.html.twig
+++ b/symfony/twig-bundle/6.4/templates/base.html.twig
@@ -3,7 +3,7 @@
     <head>
         <meta charset="UTF-8">
         <title>{% block title %}Welcome!{% endblock %}</title>
-        <link rel="icon" href="data:image/svg+xml,<svg xmlns=%22http://www.w3.org/2000/svg%22 viewBox=%220 0 128 128%22><text y=%221.2em%22 font-size=%2296%22>⚫️</text></svg>">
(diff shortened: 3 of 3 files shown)


symfony/twig-bundle/6.4/templates/base.html.twig
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="UTF-8">
+        <title>{% block title %}Welcome!{% endblock %}</title>
+        <link rel="icon" href="data:image/svg+xml,<svg xmlns=%22http://www.w3.org/2000/svg%22 viewBox=%220 0 128 128%22><text y=%221.2em%22 font-size=%2296%22>⚫️</text></svg>">
```
GromNaN, line 6: Do you know why we use this sober emoji instead of something more fun like ⭐ or 🎯.

symfony/twig-bundle/6.4/config/packages/twig.yaml
```diff
@@ -0,0 +1,7 @@
+twig:
+    default_path: '%kernel.project_dir%/templates'
```
stof, line 2: isn't this the default already in TwigBundle ? If yes, I suggest removing the explicit config duplicating it, to simplify things.

yceruto: yes, it's already the case

yceruto: yes, it's already the case

GromNaN: I'll remove this config: https://github.com/symfony/symfony/blob/b0f5d7e7db2a0b1ff38fb2c8c933cb8edd05d07b/src/Symfony/Bundle/TwigBundle/DependencyInjection/Configuration.php#L145C33-L145C33

symfony/twig-bundle/6.4/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "bundles": {
+        "Symfony\\Bundle\\TwigBundle\\TwigBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/",
+        "templates/": "templates/"
+    },
+    "conflict": {
+        "symfony/framework-bundle": "<5.3"
```
GromNaN, line 10: Is it still necessary? The bundle has a [conflict rule](https://github.com/symfony/symfony/blob/v6.4.0-RC1/src/Symfony/Bundle/TwigBundle/composer.json#L41) with `symfony/framework-bundle <5.4`

OskarStark:
> Related to #45845

Is not clickable

symfony/twig-bundle/6.4/config/packages/twig.yaml
fabpot: We have a few settings that are just the defaults to show how to override things. This is one of them. Maybe not that frequent, I just wanted to highlight why it was there.

symfony/twig-bundle/6.4/templates/base.html.twig
fabpot: It looks like the sf logo... but without sf 😂

stof: also, projects must change their favicon to use their own anyway. Having a default one here avoids getting 404 errors when browsers try to load a default `favicon.ico` file otherwise.

GromNaN: Oh! Let's add the `sf` then.
```suggestion
        <link rel="icon" href="data:image/svg+xml,<svg xmlns=%22http://www.w3.org/2000/svg%22 viewBox=%220 0 128 128%22><text y=%221.2em%22 font-size=%2296%22>⚫️</text><text y=%221.3em%22 x=%220.2em%22 font-size=%2276%22 fill=%22%23fff%22>sf</text></svg>">
```
Chrome
image

Firefox
image

GromNaN: I like the idea of making a icon from an emoji. It make its very easy to anyone starting a project to use something from the existing panel.

fabpot (approved):

OskarStark (approved):

PR #1272, opened 2023-12-15
Removing framework.session.cookie_samesite < FrameworkBundle 4.2

stefantalen:
The `cookie_samesite` option is not available before Symfony 4.2, having this in the recipe results in an error:

> Unrecognized option "cookie_samesite" under "framework.session"

| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/framework-bundle/3.4/config/packages/framework.yaml
+++ b/symfony/framework-bundle/3.4/config/packages/framework.yaml
@@ -7,7 +7,6 @@ framework:
     # Remove or comment this section to explicitly disable session support.
     session:
         handler_id: null
-        cookie_samesite: lax
 
     #esi: true
     #fragments: true


chalasr:
Actually the option has been backported to 3.4 some time after it was introduced (see https://github.com/symfony/recipes/pull/738 & https://github.com/symfony/symfony/pull/35605). 
Given the situation into which the mentioned error can happen are edge cases (unmaintained versions that new projects should not rely on), I think there is nothing we should do here hence I'm going to close. 
Thanks for your understanding

stefantalen:
@chalasr Thanks for the extra info! I agree with your statement that new projects should not rely on unmaintained versions.

However, in our situation we are in the process of upgrading a Symfony 3.4 project to 6.x with intermediate steps at
- 3.4 -> 4.0 (for Flex)
- 4.0 -> 4.4
- 4.4 -> 5.4
- 5.4 -> 6.x

Having configuration issues because of recipes, adds some noice to the process.

Because this option has been backported, the least we could do (for DX)  is add the configuration option in comments for 4.0 or add a comment to the addition in 3.4 stating it was backported and doesn't work when upgrading 3.4 to 4.0.

PR #1280, opened 2024-01-11
[Symfony][Router] Set the default method for a route to "GET"

lyrixx:
| Q             | A
| ------------- | ---
| License       | MIT

* I always sspecify the method(s) in my route declaration and it's boring
* Allowing by default all methods does not make sens
* Allowing by default all methods can be a security issue, because
  someone can DDOS the application by sending POST request, to a
  resource that is usually cached (but only for GET)
* explicit is better than implicit

### DIFF
--- a/symfony/routing/6.2/config/routes.yaml
+++ b/symfony/routing/6.2/config/routes.yaml
@@ -3,3 +3,4 @@ controllers:
         path: ../src/Controller/
         namespace: App\Controller
     type: attribute
+    methods: [GET]


symfony/routing/6.2/config/routes.yaml
```diff
@@ -3,3 +3,4 @@ controllers:
         path: ../src/Controller/
         namespace: App\Controller
     type: attribute
+    methods: [GET]
```
Korbeil, line 6: Maybe we should put this only for 7.0+ ?

stof:
This is not setting the default. It is forcing all routes to use that method due to the fact that properties defined on a route import are not default values for route definitions but overrides.

PR #1288, opened 2024-01-31
Moving `{% block importmap %}` *above* `{% block javascripts %}`

ThomasLandauer:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | none

@weaverryan
Disclaimer: I'm not sure about this - especially I didn't check if `before_target` is actually doing what I want ;-)

Here's the scenario I'm talking about:
```twig
{# base.html.twig #}
{% block javascripts %}
    {% block importmap %}{{ importmap('app') }}{% endblock %}
    // Here's some important JavaScript stuff that's needed on *every* page
{% endblock %}
```
```twig
{# some_page.html.twig #}
{% block importmap %}{{ importmap(['app', 'some_page']) }}{% endblock %}
```
Now how can I get the other important JavaScript onto `some_page`? IMO only by doing:
```twig
{% block javascripts %}
    {{ parent() }}
{% endblock %}
```
But this will also render the importmap *twice*.
Right?

My solution: Move `{% block importmap %}` **outside** of `{% block javascripts %}`. What do you think?

If you merge this, the code block at https://symfony.com/doc/6.4/frontend/asset_mapper.html#installation would need to be updated again... :-)

### DIFF
--- a/symfony/asset-mapper/6.4/manifest.json
+++ b/symfony/asset-mapper/6.4/manifest.json
@@ -16,7 +16,7 @@
         {
             "file": "templates/base.html.twig",
             "content": "{% block importmap %}{{ importmap('app') }}{% endblock %}",
-            "position": "after_target",
+            "position": "before_target",
             "target": "{% block javascripts %}",
             "warn_if_missing": true
         }


symfony/asset-mapper/6.4/manifest.json
```diff
@@ -16,7 +16,7 @@
         {
             "file": "templates/base.html.twig",
             "content": "{% block importmap %}{{ importmap('app') }}{% endblock %}",
-            "position": "after_target",
+            "position": "before_target",
```
weaverryan, line 19: Interesting situation :). But (see error above), there is no `before_target` - there just wasn't a need for it. Either we need to think of some other solution or add that to Flex.

smnandre:
**TL;DR;** there is nothing to change 

(and, personal opinion, i think this use case is a bit too specific to change the recipe anyway)

### Base scenario 

```twig
{# foo.html.twig #}
{% block javascripts %}

{% block importmap %} -importmap- {% endblock %}

--important js--

{% endblock %}
```


```twig
{# bar.html.twig #}
{% extends 'foo.html.twig' %}

```

Result: 

```html
-import map 
--important js--
```


### Override importmap:

```twig
{# bar.html.twig #}
{% extends 'foo.html.twig' %}

{% block importmap %} -another import map {% endblock %}

```

Result: 

```html
  -another import map 
--important js--
```

### Override important JS:

```twig
{# bar.html.twig #}
{% extends 'foo.html.twig' %}

{% block javascripts %} --another important js-- {% endblock %}
```

Result: 

```html
--another important js-- 
```

### Override important JS + parent

```twig
{# bar.html.twig #}
{% extends 'foo.html.twig' %}

{% block javascripts %}
{{ parent() }}
 --another important js-- 
{% endblock %}
```

Result: 

```html
    -importmap- 
--important js--

    --another important js-- 
```



### Override both:

```twig
{# bar.html.twig #}
{% extends 'foo.html.twig' %}

{% block importmap %}
- another importmap
{% endblock %}

{% block javascripts %}
{{ parent() }}
 --another important js-- 
{% endblock %}

```

Result: 
```html
- another importmap

--important js--

    --another important js-- 
```


https://twigfiddle.com/ke86dz

ThomasLandauer:
Thanks, you're right! I thought it's required/recommended to override the blocks in the same nesting situation as they are defined:
```twig
{% block javascripts %}
{% block importmap %} - another importmap- {% endblock %}
{{ parent() }}
 --another important js-- 
{% endblock %}
```
Which then lead to my problem above.

So I'm closing here, since this is a general Twig misunderstanding, unrelated to this recipe.

Anyway, I still can't see any real reason why `{% block importmap %}` should live *inside* `{% block javascripts %}` - so I guess it's just a matter of personal taste.

PR #1289, opened 2024-02-01
[symfony/twig-bundle] Add name to `{% endblock %}`

ThomasLandauer:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | none

Reasons:
* It's good practice anyway IMO, especially when blocks are getting longer and longer...
* Would solve the problem at https://github.com/symfony/recipes/pull/1288#discussion_r1474889507 nicely :-)


nicolas-grekas:
Let me close - these block will always remain small since the file should be extended anyway.

PR #1294, opened 2024-02-16
Use "resolve" processor for MAILER_DSN env variable

leonboot:
| Q             | A
| ------------- | ---
| License       | MIT

For the Doctrine bundle, the `DATABASE_URL` variable is processed by the `resolve` env var processor. This is nice, since the value can be read from file by using the `file` processor when defining its value, for instance. It would be nice if this would be applied to the `MAILER_DSN` for the Symfony Mailer as well, for the same reasons.

### DIFF
--- a/symfony/mailer/4.3/config/packages/mailer.yaml
+++ b/symfony/mailer/4.3/config/packages/mailer.yaml
@@ -1,3 +1,3 @@
 framework:
     mailer:
-        dsn: '%env(MAILER_DSN)%'
+        dsn: '%env(resolve:MAILER_DSN)%'


leonboot:
It seems the "Run updated recipes" job fails because the package version to install is distilled from the modified files of the PR, but symfony/mailer:^4.3 can't be installed because of its dependency on symfony/event-dispatcher.

Offering an updated PR by creating a 5.0 version of the recipe will probably still break the v6 test. And the change provided in this PR will still work for Symfony 4 as far as I can tell, since the `resolve` processor is already available there.

nicolas-grekas:
Hi, I'm not sure I understand the motivation for this change. What's the relation to the `file` processor?

fabpot:
Closing as there is no more feedback.

leonboot:
Looking at the documentation again, this may have been a result of misinterpretation on my part. I believed the `resolve:` env var processor would allow for using other env var processors in the env var's value (like `file:DATABASE_URL_FILE`, for instance). But looking at the docs, the resolve processor only resolved `%variable%` notations.

PR #1296, opened 2024-02-22
chore: Always catch errors when calling `console`

llupa:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | [#53946](https://github.com/symfony/symfony/issues/53946)

Pinging @wouterj @greg0ire 

This is my first time doing "recipe" work and I am not sure if this is the correct place to open this PR. Let me know for any change needed! 

In short: when executing console commands in production mode/env commands **have to fail** with exit 1, and not render the `\Error` and exit 0.

### DIFF
--- a/symfony/console/6.4/bin/console
+++ b/symfony/console/6.4/bin/console
@@ -1,0 +1,20 @@
+#!/usr/bin/env php
+<?php
+
+use App\Kernel;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+
+if (!is_file(dirname(__DIR__).'/vendor/autoload_runtime.php')) {
+    throw new LogicException('Symfony Runtime is missing. Try running "composer require symfony/runtime".');
+}
+
+require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
+
+return function (array $context) {
+    $kernel = new Kernel($context['APP_ENV'], (bool) $context['APP_DEBUG']);
+
+    $app = new Application($kernel);
+    $app->setCatchErrors('prod' === $context['APP_ENV']);
+
+    return $app;
+};
--- a/symfony/console/6.4/bin/console
+++ b/symfony/console/6.4/bin/console
@@ -14,7 +14,7 @@ return function (array $context) {
     $kernel = new Kernel($context['APP_ENV'], (bool) $context['APP_DEBUG']);
 
     $app = new Application($kernel);
-    $app->setCatchErrors('prod' === $context['APP_ENV']);
+    $app->setCatchErrors(true);
 
     return $app;
 };
--- a/symfony/console/6.4/manifest.json
+++ b/symfony/console/6.4/manifest.json
@@ -1,0 +1,6 @@
+{
+    "copy-from-recipe": {
+        "bin/": "%BIN_DIR%/"
+    },
+    "aliases": ["cli"]
+}


symfony/console/7.0/bin/console
```diff
@@ -0,0 +1,20 @@
...
+    throw new LogicException('Symfony Runtime is missing. Try running "composer require symfony/runtime".');
+}
+
+require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
+
+return function (array $context) {
+    $kernel = new Kernel($context['APP_ENV'], (bool) $context['APP_DEBUG']);
+
+    $app = new Application($kernel);
+    $app->setCatchErrors('prod' === $context['APP_ENV']);
```
greg0ire, line 17: Why would you want to ever not catch errors? :thinking:

symfony/console/7.0/bin/console
greg0ire, line 1: All Symfony versions that can be installed on PHP 8.3 should probably be changed.

symfony/console/7.0/bin/console
llupa: I see it the same way as you. But I tried to keep it in the _spirit_ of comments at this [pull-request](https://github.com/symfony/symfony/pull/50420)

While developing or testing, you might not want it to happen? But in `prod` definitely catch it.

greg0ire: Sorry, which comment(s) are you referring to?

llupa: https://github.com/symfony/symfony/pull/50420#issuecomment-1576688896, https://github.com/symfony/symfony/pull/50420#issuecomment-1576712904 and https://github.com/symfony/symfony/pull/50420#issuecomment-1576849725

> What about adding a flag to choose if we should handle throwable or not? 

> Sounds like a good idea 👍

> ...
> And if the developer wants more output, they can run the same command with -v flags

Maybe I am misunderstanding, but it was my impression that the ability to catch was intended to be optional.

symfony/console/7.0/bin/console
llupa: I could add something for 5.4, 6.4 too but two things come to mind, 

a) #1293 this PR might be relevant as I noticed the tests go only up to 6.4 (I might have misunderstood this) 
b) this catch-errors feature was only introduced in 6.4 so a backport / fix might be needed to 5.4? 

I only mention these because I kind of get for the most part the contribution process, but I am a bit uncertain at times.

symfony/console/7.0/bin/console
greg0ire: The ability to catch is optional/configurable, but:

- given how bad Symfony behaves when it's off
- given the default value for errors is inconsistent with the default value for exceptions
- given inconsistencies between dev and prod might lead to WTF

I'd recommend to aim for something simple: the same experience everywhere.

symfony/console/7.0/bin/console
llupa: So I tried with skeleton web-apps of Symfony and just did a _close to real life_ scenario. Symfony 5.4 is OK because both Error and Exception are handled in the same place. A thrown error will exit with 1. 

I moved the PR down to 6.4 so it always catches Errors too. But I am not sure how to fix the tests :-/

nicolas-grekas:
This would be a work around. Let's fix the root issue instead, see https://github.com/symfony/symfony/pull/54189

wouterj:
@nicolas-grekas I think aside of this issue that triggered this PR, we might want to still do this.

The console app now defaults to `catchExceptions: true` and `catchErrors: false`. I think it makes sense to make both `true` in Symfony 8, meaning the console terminate event is dispatched for both exceptions and errors.
Thus we can do the traditional BC cycle here: deprecate not setting catchErrors, change recipe to always use new default (this PR), and in 8.0, swap default around, undeprecate not setting catchErrors and update recipe to not override the default.
What do you think?

(btw, thanks for providing a patch in the error handler!)

llupa:
Not as detailed as Wouter's comment, but I also will have to set it to true for my projects because I cannot really think of a use case (DX) where I want Exceptions to be caught and rendered one way, but Errors not caught (probably will be after your patch) but rendered a different way. 

Maybe it boils down to personal preference, but it is odd that it works differently for one from the other.

nicolas-grekas:
I think we should minimize those flip-flap with deprecations so I'd better not change anything personally.
Also the current logic is perfectly fine to me. Error are split from Exception because they're not the same beast.

PR #1299, opened 2024-03-14
[doctrine/doctrine-bundle] Disable controller resolver auto mapping by default

bobvandevijver:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | https://github.com/doctrine/DoctrineBundle/pull/1762 & https://github.com/symfony/symfony-docs/pull/19671

Reflects the new default for doctrine.orm.controller_resolver.auto_mapping to ensure new users will not be faced with a deprecation when starting with doctrine bundle 2.12 or higher.

### DIFF
--- a/doctrine/doctrine-bundle/2.12/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/2.12/config/packages/doctrine.yaml
@@ -1,0 +1,52 @@
+doctrine:
+    dbal:
+        url: '%env(resolve:DATABASE_URL)%'
+
+        # IMPORTANT: You MUST configure your server version,
+        # either here or in the DATABASE_URL env var (see .env file)
+        #server_version: '16'
+
+        profiling_collect_backtrace: '%kernel.debug%'
+        use_savepoints: true
+    orm:
+        auto_generate_proxy_classes: true
+        enable_lazy_ghost_objects: true
+        report_fields_where_declared: true
+        validate_xml_mapping: true
+        naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
+        auto_mapping: true
+        mappings:
+            App:
+                type: attribute
+                is_bundle: false
+                dir: '%kernel.project_dir%/src/Entity'
+                prefix: 'App\Entity'
+                alias: App
+        controller_resolver:
+            auto_mapping: false
+
+when@test:
+    doctrine:
+        dbal:
+            # "TEST_TOKEN" is typically set by ParaTest
+            dbname_suffix: '_test%env(default::TEST_TOKEN)%'
+
+when@prod:
+    doctrine:
+        orm:
+            auto_generate_proxy_classes: false
+            proxy_dir: '%kernel.build_dir%/doctrine/orm/Proxies'
+            query_cache_driver:
+                type: pool
+                pool: doctrine.system_cache_pool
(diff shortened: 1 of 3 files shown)


bobvandevijver:
I do not think the action failure is something I can solve, or is it?

derrabus (approved):

stof (approved):

seb-jean:
Hi,
Since the default value of setting `doctrine.orm.controller_resolver.auto_mapping` is `false`, I get the following error:

`Cannot autowire argument $user of "App\Controller\ProfileController::show()": it needs an instance of "App\Entity\User" but this type has been excluded in "config/services.yaml".`

My controller:
```php
// src/Controller/ProfileController.php
namespace App\Controller;

use App\Entity\User;
use Symfony\Component\HttpFoundation\Request;
use Symfony\Component\HttpFoundation\Response;
use Symfony\Component\Routing\Attribute\Route;

class ProfileController
{
    #[Route('/profile/{ulid}')]
    public function show(Request $request, User $user): Response
    {
        // ...
    }
}
```

kbond:
Does changing `{ulid}` to `{user}` solve (or changing this config value to true)?

seb-jean:
With `{user}`, I have error:
```
An exception occurred while executing a query: SQLSTATE[22P02]: Invalid text representation: 7 ERROR: invalid input syntax for type integer: "1C9UESfu5Jk1kuhoF1vUh8"
CONTEXT: unnamed portal parameter $1 = '...'
```

With `doctrine.orm.controller_resolver.auto_mapping` to true, it works but it is not recommended.

kbond:
I think in your case, you'll want to explicitly use the MapEntity attribute.

ebitkov:
For further reading:


- https://github.com/doctrine/DoctrineBundle/pull/1762
- https://github.com/symfony/symfony/issues/50739

seb-jean:
It works with `#[MapEntity]`

```php
// src/Controller/ProfileController.php
namespace App\Controller;

use App\Entity\User;
use Symfony\Component\HttpFoundation\Request;
use Symfony\Component\HttpFoundation\Response;
use Symfony\Component\Routing\Attribute\Route;

class ProfileController
{
    #[Route('/profile/{ulid}')]
    public function show(
        Request $request,
        #[MapEntity(mapping: ['ulid' => 'ulid'])] User $user
    ): Response {
        // ...
    }
}
```

fabpot:
We have the same issue as described above in the Symfony book.

We now need to use `#[MapEntity(mapping: ['slug' => 'slug'])]` which seems unnecessary.

I think we should revert this change.

stof:
The auto-mapping using **all** fields causes lots of weird cases though (especially when you can have 2 entities in your signature. Having this feature off by default avoids lots of weird issues (which might be _silent_ bugs)

stof:
Note that we only changed the recipes to change the value for _new_ projects.

nicolas-grekas:
I'm proposing to revert this change in #1302: this breaks the DX, we need to find a better way.

nicolas-grekas:
FYI, I'm proposing https://github.com/symfony/symfony/pull/54455 for a possible solution to these ambiguities.

nicolas-grekas:
Resubmitted as https://github.com/symfony/recipes/pull/1316

PR #1302, opened 2024-03-27
Enable entity automapping for controllers again

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Reverts the setting set in #1299

I'm not sure this is the solution we want to merge, but I'm sure the recent changes have not been made with DX in mind, and they're making the experience worse (see comments in https://github.com/symfony/recipes/pull/1299)

I'm kindly asking to redesign this change with DX in mind. :pray: If we don't have any better idea yet, I suggest reverting the change on DoctrineBundle until we have a DX-friendly solution.

### DIFF
--- a/doctrine/doctrine-bundle/2.12/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/2.12/config/packages/doctrine.yaml
@@ -23,7 +23,7 @@ doctrine:
                 prefix: 'App\Entity'
                 alias: App
         controller_resolver:
-            auto_mapping: false
+            auto_mapping: true
 
 when@test:
     doctrine:


bobvandevijver:
I am strongly against this: the recipe currently follows the default of the doctrine bundle, and was adjusted to not cause deprecation notices on new installations. This change should be put on hold, and only be merged if the default in the doctrine bundle changes.

stof:
@nicolas-grekas the auto-mapping feature causes WTF behavior when you have multiple entities involved in the signature of the controller, especially because it can silently load an unwanted entity. And this is not only theoretical: I've had silent bugs in production loading an unwanted entity because of this behavior.
The auto-mapping indeed works nice for the very simple case (the one in the demo for instance), but its actual impact for more complex cases is totally bad DX.

nicolas-grekas:
Sure, I'm not negating all this.

What I'm saying is that the current changes are not enough. It's not enough to break DX for the simple cases because DX for complex cases. That's a bad compromise if I can say it this way.

We should find a way to improve DX in both simple and complex situation (or improve DX in complex situations without breaking DX in simple ones).

The current solution is just a shift from one evil to another. That's not what I call an improvement. We should seek for something better.

ebitkov:
So to clarify:

- Auto-wiring an entity by the primary key is still working
- Auto-mapping doesn't reliably work with multiple autowired entities
- Auto-mapping doesn't reliably work with multiple, in part optional parameters

Are there any cases the previous auto-mapping breaks? 
Are there any possible ideas to re-able single-parameter auto-mapping (e.g. `#[Route("/article/{slug}"]`), because I feel like this is the main concern.

Maybe this helps to move this discussion to a solution.

stof:
@ebitkov the issue is how to guess the fact that you want to use the `slug` request parameter to create a mapping to fill a `product` argument based on its `slug` property ?
Some cases being _broken_ by the auto-mapping end up being broken exactly because of a guessed mapping that look like that.

ebitkov:
@stof I'm not a Symfony Contributor with little to no knowledge about the inner magic of this framework, so bare with me. 

In my mind, something like 

```php
class ArticleController 
{
    #[Route("/article/{slug}")
    public function show(Article $article): Response
    {
        # ...
    }
}
```

and `Article` being a doctrine-managed entity with a property `$slug`, looks to me like a clear intend of "map that single parameter to the single entity I configured", but I guess that's the same thought, that lead to the current problem. 

Maybe we could create a "simple auto-mapping", that only supports single parameter to entity mapping and throw an exception when ever you try to do more. Maybe with a notice like "can't auto-map entity, use the attribute". 

That would remove the need to map something like this manually:

```php
class ArticleController 
{
    #[Route("/article/{slug}")
    public function show(
        #[MapEntity(mapping: ["slug" => "slug"])] Article $article
    ): Response
    {
        // ...
    }
}
```

---

I guess cases like these are the problem why auto-mapping might be difficult: 

```php
class ArticleController 
{
    #[Route("/article/{id}/{slug}")
    public function show(Article $article): Response
    {
        // ...
    }
}
```

or

```php
class ArticleController 
{
    #[Route("/article/{slug}/comment/{id}")
    public function show(Article $article, Comment $comment): Response
    {
        // ...
    }
}
```
---

The more I thing about it the more I feel like it actually does more sense to force developers map there entities themselves then have a half-cooked auto-mapper workaround just to save a single attribute.

On paper it sounds great to just put your entities into the arguments of your action method, but I rarely use this features, because I often need to check other fields (e.g. status, publishing date) before showing the entity and I prefer to do this via DQL/SQL then preemptively loading the entity and then check, if the field are correct.   

Sooo... I don't know. I seems to be more work then use to me at this point. 

What you think @nicolas-grekas?

derrabus (approved):
I agree that we should merge this PR, although I still believe that DoctrineBundle is on the right track.

The auto mapping feature is background magic that can have surprising side effects as others have elaborated already. Turning the feature off by default (which is what doctrine/DoctrineBundle#1762 is prepared) is the safe option.

By enabling the feature in the recipe, we improve the developer experience for new projects. At the same time, we give developers a very obvious way to disable the feature, should it ever cause trouble for them.

bobvandevijver (changes requested):
If the recipe enables the feature by default, no developer will ever see that its default in the doctrine bundle is disabled, untill it hits the unwanted behaviour already described causing them to look for this. Having this option enabled in the recipe is therefore bad DX, no question.

If you do read the documentation (as you should), and you want to use this while being informed as to how it works (which the documentation in its current state does not), you can enable it. 

At the very least this will need a comment with a warning regarding the bad behaviour of the auto mapping feature, leading the developer to the documentation update that I have prepared for this: https://github.com/symfony/symfony-docs/pull/19671.

derrabus:
> If the recipe enables the feature by default, no developer will ever see that its default in the doctrine bundle is disabled,

And they don't have to care what the default is. The recipe makes the option discoverable.

> no question

That's kind of not how I personally exchange arguments, but you do you.

> At the very least this will need a comment with a warning

No strong opinion here. We can certainly add a clarifying comment with a link to the docs.

derrabus:
> a value that is working correctly for all use cases

If that one-size-fits-all setting would exist, we wouldn't make it configurable.

> "broken" auto mapping

This is probably the point where we disagree. The auto mapping feature is not "broken" per se. It's a DX feature that can be turned off if you need more control over how requests are mapped to entity identifiers.

fabpot (approved):

PR #1307, opened 2024-04-18
fix healthcheck postgres user

DrWarpMan:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

By default, root user is used to connect to the postgres database.
If the database does not have a "root" role, the following error is logged everytime the health check happens:
`FATAL:  role "root" does not exist`

It does not seem to affect the 'healthy' status of the container (the command still returns zero exit code), but it spams the log:
```
database-1  | 2024-04-18 06:26:08.016 UTC [67] FATAL:  role "root" does not exist
database-1  | 2024-04-18 06:27:08.192 UTC [69] FATAL:  role "root" does not exist
database-1  | 2024-04-18 06:28:08.352 UTC [71] FATAL:  role "root" does not exist
```

### DIFF
--- a/doctrine/doctrine-bundle/2.12/manifest.json
+++ b/doctrine/doctrine-bundle/2.12/manifest.json
@@ -29,7 +29,7 @@
                 "    POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-!ChangeMe!}",
                 "    POSTGRES_USER: ${POSTGRES_USER:-app}",
                 "  healthcheck:",
-                "    test: [\"CMD\", \"pg_isready\"]",
+                "    test: [\"CMD\", \"pg_isready -U ${POSTGRES_USER:-app}\"]",
                 "    timeout: 5s",
                 "    retries: 5",
                 "    start_period: 60s",


fabpot (approved):

gabrielpenide:
I am having problems with this healthcheck change, as database containers state they are unhealthy. Turns out that making the `test` line look like
```yaml
test: ["CMD", "pg_isready", "-U", "${POSTGRES_USER:-app}"]
```
partially solves the problem (containers are now healthy) but logs are populated with the following message:
```
FATAL:  database "<redacted - shows $POSTGRES_USER variable value>" does not exist
```
Further modification of the line solves the problem for me:
```yaml
test: ["CMD", "pg_isready", "-d", "${POSTGRES_DB:-app}", "-U", "${POSTGRES_USER:-app}"]
```

Not sure, but it seems something related with using `CMD` vs. `CMD-SHELL` and/or incorrectly escaping variables. Can anyone share some thoughts about this?

gabrielpenide:
Yes. They both are different from the default value `app` and they are both different from each other.

DrWarpMan:
The change you made should solve the issue, would you like to make the pull request yourself?

```yaml
test: ["CMD", "pg_isready -d ${POSTGRES_DB:-app} -U ${POSTGRES_USER:-app}"]
```

gabrielpenide:
I can make the pull request, of course. But first I want to notice that I wasn't able to make this work with your exact suggestion. I mean that I need to use the comma-separated style for the healthcheck to pass:
```yaml
test: ["CMD", "pg_isready", "-d", "${POSTGRES_DB:-app}", "-U", "${POSTGRES_USER:-app}"]
```
I'm very far from a Docker expert. That's why I pointed out this on my original message:
> Not sure, but it seems something related with using `CMD` vs. `CMD-SHELL` and/or incorrectly escaping variables.

I would appreciate if you or anyone else elaborates on this before I make the pull request, because it slightly changes your original choice.

DrWarpMan:
Yes, you are right. In this case your provided solution is better.
Apparently the two type of syntaxes are called "shell" and "exec" forms.

This answer explains it more: https://stackoverflow.com/a/47940538
It also references to the following Docker documentation, which provides even more information:
https://docs.docker.com/reference/dockerfile/#shell-and-exec-form

Does your database name contain any special characters, spaces or something?

gabrielpenide:
It doesn't. Just lowercase [a-z] letters, not even numbers or special characters.

DrWarpMan:
From what I tested, it seems there are only two possible combinations, either you use:
```yaml
test: ["CMD-SHELL", "pg_isready -d ${POSTGRES_DB:-app} -U ${POSTGRES_USER:-app} || exit 1"]`
```
or
```yaml
test: ["CMD", "pg_isready", "-d", "${POSTGRES_DB:-app}", "-U", "${POSTGRES_USER:-app}"]
```
(Apparently when using CMD, `|| exit 1` is appended automatically.)

Can you try both syntaxes and see if both of them work @gabrielpenide ?

gabrielpenide:
Both work for me. Personally, I would go with the `CMD` option, but I guess that's a matter of taste.

PR #1312, opened 2024-05-03
api-platform 3.3 configuration

soyuka:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/api-platform/core/3.3/config/packages/api_platform.yaml
+++ b/api-platform/core/3.3/config/packages/api_platform.yaml
@@ -1,0 +1,18 @@
+api_platform:
+    title: Hello API Platform
+    version: 1.0.0
+    formats:
+        jsonld: ['application/ld+json']
+    docs_formats:
+        jsonld: ['application/ld+json']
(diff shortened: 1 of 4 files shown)


api-platform/core/3.3/config/packages/api_platform.yaml
```diff
@@ -0,0 +1,18 @@
...
+        html: ['text/html']
+    defaults:
+        stateless: true
+        cache_headers:
+            vary: ['Content-Type', 'Authorization', 'Origin']
+        extra_properties:
+            standard_put: true
+            rfc_7807_compliant_errors: true
+    keep_legacy_inflector: false
+    use_symfony_listeners: true
```
soyuka, line 18: @dunglas should we keep this to true? If we continue to recommend using controllers we should keep this to true, if not maybe that we should recommend using false instead?

dunglas: I would set it to false by default, but I've no strong opinion about this.

dunglas (approved):

fabpot (approved):

PR #1314, opened 2024-05-15
[framework-bundle] Prefix generated APP_SECRET by NotSecure-

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

While it's ok to commit the `APP_SECRET` used for dev and test environments, in production, the `APP_SECRET` environment variable must be overridden by a local value that is not committed (using a secret vault, using Symfony Secrets, creating a local environment variable, or creating an unversioned `.env.prod` file are acceptable options).

With this patch, Flex will prefix the generated value with `NotSecure-` to make it clear that the committed value must not be used in production.

Brought to our attention by @ramsey's https://phpc.social/@ramsey/112437517679689885.
See also https://github.com/symfony/symfony/issues/38021.

We'll also have to document that `APP_SECRET` is used as the value for https://symfony.com/doc/current/reference/configuration/framework.html#secret

### DIFF
--- a/symfony/framework-bundle/6.4/manifest.json
+++ b/symfony/framework-bundle/6.4/manifest.json
@@ -13,7 +13,7 @@
     },
     "env": {
         "APP_ENV": "dev",
-        "APP_SECRET": "%generate(secret)%"
+        "APP_SECRET": "NotSecure-%generate(secret)%"
     },
     "gitignore": [
         "/.env.local",
--- a/symfony/framework-bundle/7.0/manifest.json
+++ b/symfony/framework-bundle/7.0/manifest.json
@@ -13,7 +13,7 @@
     },
     "env": {
         "APP_ENV": "dev",
-        "APP_SECRET": "%generate(secret)%"
+        "APP_SECRET": "NotSecure-%generate(secret)%"
     },
     "gitignore": [
         "/.env.local",


javiereguiluz:
Thanks Kévin, but I'm not sure this is the best solution for the issue mentioned:

* Now, a new Symfony project will show `NotSecure-...` as the value of the secret; some people will think that Symfony is not safe by default;
* Others will just remove the `NotSecure-` prefix or will replace this value entirely but will keep committing it to the repo.

In my opinion, we can't properly solve this without a good comment below `APP_SECRET` explaining the situation and giving clear indications about what to do.

dunglas:
@javiereguiluz by definition, `.env` contains placeholder environment variables that serve as examples of how to configure the app on production. To me, having a `NotSecure-` prefix in one of these values doesn't say that Symfony isn't secure but on the contrary, that Symfony takes security seriously and makes it crystal clear that this value must not be used as-is in production.

API Platform has always used a similar placeholder for `APP_SECRET` (which, IMHO, is even more clear), and as far as I know, nobody says that API Platform isn't secure: https://github.com/api-platform/api-platform/blob/main/api/.env#L23

nicolas-grekas:
>API Platform has always used a similar placeholder for APP_SECRET (which, IMHO, is even more clear), and as far as I know, nobody says that API Platform isn't secure

Then I'm going to be the one saying this is highly insecure. There are scanners out there that do test for vulnerable Symfony apps that didn't change their APP_SECRET and that's a successful approach to hack many webapps apparently. "not is us" is not a satisfactory answer IMHO. This is related to https://github.com/symfony/symfony/pull/52471

I'm :-1: with this PR because I think we should do better on the topic.

Here are some thoughts:

Ideally, we would remove this APP_SECRET entirely for skeleton apps. And for apps that do need it, we already have another stable secret we can use: SYMFONY_DECRYPTION_SECRET. It'd be great if we could derivate kernel.secret from that var when available.

I also had a look at services that do need this secret, here they are:

- `uri_signer` - this one is used to generate signed URLs when using ESI fragments. After https://github.com/symfony/symfony/pull/52471, can we consider removing this requirement and encrypt only opportunistically when kernel.secret is populated?
- `security.authenticator.remember_me*` - here, we could use tokens stored in the user DB. Would that make sense?
- `security.authenticator.abstract_login_link_signature_hasher` - for this one, tokens could be used instead?
- `security.login_throttling.%firewallName%.limiter` and here, is the secret really needed when computing the hash? What if we didn't do it?

If we think we *do* need kernel.secret, then an idea could be to turn this into a service instead of a parameter. That service would have one responsibility: throw a nice and actionable error message when no secret is configured, with a __toString method to pass the secret to consumers when its set.

dunglas:
@nicolas-grekas I agree with your long-term approach. However, this patch has the benefit of improving the security and the DX right now, and can easily be reverted when your approach is implemented.

Big +1 for the service, so we could also improve the support for Docker and Kubernetes secrets, HashiCorp Vault, etc. by providing ad-hoc implementations.

> Then I'm going to be the one saying this is highly insecure. There are scanners out there that do test for vulnerable Symfony apps that didn't change their APP_SECRET and that's a successful approach to hack many webapps apparently. 

Most apps also need API tokens, SMTP passwords etc. We need to find a way to make it clear that values in `.env` are only placeholders that sometimes need to be changed in production.
Maybe could we do something like throwing if we detect at runtime that an env var suffixed by `_SECRET` has not been overrode?

OskarStark:
What about renaming .env to .env.placeholders or .env.template or .env.whatever ?

I like your idea @nicolas-grekas 👍🏻

dunglas:
@OskarStark We follow the same convention as most ecosystems (Angular, Ruby etc) and it's widely accepted that `.env` is the place for placeholders. Let's not bring our own convention while there is already a one that is popular.

ebitkov:
> Note that this threat model analysis makes me realize we don't have any way to rotate the secret, so that people leaving the company who had access to the secret couldn't get long term access to such capabilities. This is most concerning to me: not having a way to properly guard against a threat you want to alleviate.

I noticed the same problem when facing this topic. These are the ways of generating a new `APP_SECRET` value I found so far:

- Look up an online hex number generator and create a new value there.
- Re-install all recipes by running `rm symfony.lock; composer recipes:install --force`. This generates a new value for the `APP_SECRET` variable, but also resets all your configuration to the default.
- Create a new blank project and take the generated value from there
- Use `bin/console secrets:set APP_SECRET --random` to generate a new random value in the vault. This generates a shorter value (16 chars) and not a hex number, but a random string (dunno, if that actually matters)

The last one is the most sensible, but for me as a developer, it's confusing, that `secret:set --random` creates a structurally different value then the one generated in the `.env` file (it's not a hex string and way shorter), so I leave the value for `APP_SECRET` in `.env` as it is, since I don't want to break anything...

So yes, an easy way to generate a new value for `APP_SECRET`, that is well documented, would already help. Maybe extending `secret:set --random` would be the way to go, since secrets belong into the vault anyway.

nicolas-grekas:
Closing in favor of https://github.com/symfony/recipes/pull/1317 :crossed_fingers: 
Thanks for shaking this topic :)

PR #1316, opened 2024-05-21
[doctrine/doctrine-bundle] Disable automapping when using FWB >= 7.1

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Follows #1299 and #1302

Possible after https://github.com/symfony/symfony/pull/54720

Use explicit mappings instead, e.g.:
`#[Route('/conference/{id:conference}/{slug:conference}')]`

/cc @bobvandevijver we could now deprecate *enabling* automapping in doctrine-bundle. Up to send a PR doing so?


bobvandevijver (approved):
I will look into the mentioned PR for the doctrine bundle 👍🏻

yceruto (approved):

fabpot (approved):

PR #1324, opened 2024-07-03
Check for existence of vendor directory for better error message

94noni:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | .

replicates https://github.com/symfony/recipes/pull/1301 for the web entry point
I came across this on a recent project I boot up, inside `bin/console` but not `public/index`

please feel free to close if it is not relevant

### DIFF
--- a/symfony/framework-bundle/7.2/public/index.php
+++ b/symfony/framework-bundle/7.2/public/index.php
@@ -10,7 +10,9 @@ if (!is_file(dirname(__DIR__).'/vendor/autoload_runtime.php')) {
     throw new LogicException('Symfony Runtime is missing. Try running "composer require symfony/runtime".');
 }
 
-require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
+if (false === include dirname(__DIR__).'/vendor/autoload_runtime.php') {
+    throw new LogicException('Symfony Runtime is missing. Try running "composer require symfony/runtime".');
+}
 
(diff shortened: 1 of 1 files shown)


symfony/framework-bundle/7.2/public/index.php
```diff
@@ -2,6 +2,14 @@
...
 
+if (!is_dir(dirname(__DIR__).'/vendor')) {
+    throw new LogicException('Dependencies are missing. Try running "composer install".');
+}
+
+if (!is_file(dirname(__DIR__).'/vendor/autoload_runtime.php')) {
+    throw new LogicException('Symfony Runtime is missing. Try running "composer require symfony/runtime".');
+}
+
 require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
```
nicolas-grekas, line 13: I'd rather go with less calls on the hotpath, maybe something to sync with the console script?
```suggestion
if (false === include dirname(__DIR__).'/vendor/autoload_runtime.php') {
    throw new LogicException('Symfony Runtime is missing. Try running "composer require symfony/runtime".');
}
```

94noni: I applied your suggestion here

but I am unsure to understand your second part of the comment : `maybe something to sync with the console script?`
may I ask you to explicit a bit more please ?

PR #1326, opened 2024-07-08
[Notifier] Enable auto-configuration for the `GoIP` bridge

ahmedghanem00:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  |

This has been forgotten to be set/enabled when the recipe was introduced for the first time.

### DIFF
--- a/symfony/go-ip-notifier/6.4/manifest.json
+++ b/symfony/go-ip-notifier/6.4/manifest.json
@@ -1,4 +1,13 @@
 {
+    "add-lines": [
+        {
+            "file": "config/packages/notifier.yaml",
+            "position": "after_target",
+            "warn_if_missing": true,
+            "target": "        texter_transports:",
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

ahmedghanem00:
@OskarStark Thanks for the review. It's notable also that since this package was renamed last year from the old name `goip-notifier` to the new name `go-ip-notifier`, the old name is still present on Packagist [here](https://packagist.org/packages/symfony/goip-notifier) ​​and the package can only be downloaded/installed via this old name.

This also makes the package recipe here not work correctly because Flex assumes the new name 😞. So, someone with proper access to Packagist may do the renaming there as well? (abandon the package with the old name and then re-publish it again under the new name)

OskarStark:
friendly ping @fabpot

fabpot (approved):

PR #1331, opened 2024-07-30
Introduce APP_CACHE_POOLS_DIR to ease splitting app from system pools

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

So that splitting app from system pools becomes just a matter of setting an env var.


nicolas-grekas:
I forgot that we already split var/cache/prod/pools/system from var/cache/prod/pools/app so that it's just a mounting option away.

PR #1333, opened 2024-08-19
[symfony/framework-bundle/7.0 & 7.2] Add imports for services.yaml wi…

Arkemlar:
…th some comments for clarification

| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | .

This is tiny change for `config/services.yaml` file. I lost hours trying to understand how to use multiple service files correctly and everything I needed to know is the import order. Now it sounds obvious for me, but lots of stackoverflow questions shows that I am not the only one stuck on that. So I added some comments for imports statement to services.yaml. Hope it helps others to save time.

### DIFF
--- a/symfony/framework-bundle/7.0/config/services.yaml
+++ b/symfony/framework-bundle/7.0/config/services.yaml
@@ -1,6 +1,12 @@
 # This file is the entry point to configure your own services.
 # Files in the packages/ subdirectory configure your dependencies.
 
+# There comes imported service files, if you would like to split config to multiple files.
+# Symfony loads first the imported files and then it processes the parameters and services defined in this file itself.
+# Duplicating services will be overridden and last config wins.
+# https://symfony.com/doc/current/service_container/import.html
+imports:
+
 # Put parameters here that don't need to change on each machine where the app is deployed
 # https://symfony.com/doc/current/best_practices.html#use-parameters-for-application-configuration
 parameters:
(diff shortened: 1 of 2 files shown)


Arkemlar:
Hmm, can someone review and approve please?

94noni:
what about add this in the doc instead ? could be more valuable (if not already done, or written differently )

Arkemlar:
@94noni I think, there are lower chances to miss it if we place it in file.

Arkemlar:
@nicolas-grekas @stof guys, could you please review this PR and approve if it's fine?

symfony/framework-bundle/7.0/config/services.yaml
```diff
@@ -1,6 +1,12 @@
 # This file is the entry point to configure your own services.
 # Files in the packages/ subdirectory configure your dependencies.
 
+# There comes imported service files, if you would like to split config to multiple files.
```
fabpot, line 4:
```suggestion
# Import additional files that should be processed before this file.
```

symfony/framework-bundle/7.0/config/services.yaml
```diff
@@ -1,6 +1,12 @@
...
+# Symfony loads first the imported files and then it processes the parameters and services defined in this file itself.
+# Duplicating services will be overridden and last config wins.
```
fabpot, line 6: Let's remove these two lines as one can read the docs at this point.

fabpot (changes requested):

PR #1337, opened 2024-09-02
Enable stateless headers/cookies-based CSRF protection

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Relies on https://github.com/symfony/symfony/pull/58095

### DIFF
--- a/symfony/framework-bundle/7.2/config/packages/framework.yaml
+++ b/symfony/framework-bundle/7.2/config/packages/framework.yaml
@@ -8,6 +8,11 @@ framework:
     #esi: true
     #fragments: true
 
+    # Enable stateless CSRF protection for forms and logins/logouts
+    form: { csrf_protection: { token_id: submit } }
+    csrf_protection:
+        stateless_token_ids: [submit, authenticate, logout]
+
 when@test:
     framework:
         test: true
--- a/symfony/stimulus-bundle/2.20/assets/controllers/csrf_protection_controller.js
+++ b/symfony/stimulus-bundle/2.20/assets/controllers/csrf_protection_controller.js
@@ -1,0 +1,60 @@
+var nameCheck = /^[-_a-zA-Z0-9]{4,22}$/;
+var tokenCheck = /^[-_/+a-zA-Z0-9]{24,}$/;
+
+// Generate and double-submit a CSRF token in a form field and a cookie, as defined by Symfony's SameOriginCsrfTokenManager
+document.addEventListener('submit', function (event) {
+    var csrfField = event.target.querySelector('input[data-controller="csrf-protection"]');
+
+    if (!csrfField) {
+        return;
+    }
+
+    var csrfCookie = csrfField.getAttribute('data-csrf-protection-cookie-value');
+    var csrfToken = csrfField.value;
+
+    if (!csrfCookie && nameCheck.test(csrfToken)) {
+        csrfField.setAttribute('data-csrf-protection-cookie-value', csrfCookie = csrfToken);
+        csrfField.value = csrfToken = btoa(String.fromCharCode.apply(null, (window.crypto || window.msCrypto).getRandomValues(new Uint8Array(18))));
+    }
+
+    if (csrfCookie && tokenCheck.test(csrfToken)) {
+        var cookie = csrfCookie + '_' + csrfToken + '=' + csrfCookie + '; path=/; samesite=strict';
+        document.cookie = window.location.protocol === 'https:' ? '__Host-' + cookie + '; secure' : cookie;
+    }
+});
+
+// When @hotwired/turbo handles form submissions, send the CSRF token in a header in addition to a cookie
+// The `framework.csrf_protection.check_header` config option needs to be enabled for the header to be checked
+document.addEventListener('turbo:submit-start', function (event) {
+    var csrfField = event.detail.formSubmission.formElement.querySelector('input[data-controller="csrf-protection"]');
+
+    if (!csrfField) {
+        return;
+    }
+
+    var csrfCookie = csrfField.getAttribute('data-csrf-protection-cookie-value');
+
+    if (tokenCheck.test(csrfField.value) && nameCheck.test(csrfCookie)) {
+        event.detail.formSubmission.fetchRequest.headers[csrfCookie] = csrfField.value;
+    }
+});
+
+// When @hotwired/turbo handles form submissions, remove the CSRF cookie once a form has been submitted
+document.addEventListener('turbo:submit-end', function (event) {
+    var csrfField = event.detail.formSubmission.formElement.querySelector('input[data-controller="csrf-protection"]');
+
+    if (!csrfField) {
+        return;
+    }
+
+    var csrfCookie = csrfField.getAttribute('data-csrf-protection-cookie-value');
+
+    if (tokenCheck.test(csrfField.value) && nameCheck.test(csrfCookie)) {
+        var cookie = csrfCookie + '_' + csrfField.value + '=0; path=/; samesite=strict; max-age=0';
+
+        document.cookie = window.location.protocol === 'https:' ? '__Host-' + cookie + '; secure' : cookie;
+    }
+});
+
+/* stimulusFetch: 'lazy' */
+export default 'csrf-protection-controller';
--- a/symfony/stimulus-bundle/2.20/assets/bootstrap.js
+++ b/symfony/stimulus-bundle/2.20/assets/bootstrap.js
@@ -1,0 +1,2 @@
+// register any custom, 3rd party controllers here
+// app.register('some_controller_name', SomeImportedController);
--- a/symfony/stimulus-bundle/2.20/assets/controllers.json
+++ b/symfony/stimulus-bundle/2.20/assets/controllers.json
@@ -1,0 +1,4 @@
+{
+    "controllers": [],
+    "entrypoints": []
+}
--- a/symfony/stimulus-bundle/2.20/assets/controllers/hello_controller.js
+++ b/symfony/stimulus-bundle/2.20/assets/controllers/hello_controller.js
@@ -1,0 +1,16 @@
+import { Controller } from '@hotwired/stimulus';
+
+/*
+ * This is an example Stimulus controller!
+ *
+ * Any element with a data-controller="hello" attribute will cause
+ * this controller to be executed. The name "hello" comes from the filename:
+ * hello_controller.js -> "hello"
+ *
+ * Delete this file or adapt it for your use!
+ */
+export default class extends Controller {
+    connect() {
+        this.element.textContent = 'Hello Stimulus! Edit me in assets/controllers/hello_controller.js';
+    }
+}
--- a/symfony/stimulus-bundle/2.20/manifest.json
+++ b/symfony/stimulus-bundle/2.20/manifest.json
@@ -1,0 +1,41 @@
+{
+    "bundles": {
+        "Symfony\\UX\\StimulusBundle\\StimulusBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "assets/": "assets/"
+    },
+    "aliases": ["stimulus", "stimulus-bundle"],
+    "conflict": {
+        "symfony/framework-bundle": "<7.2",
+        "symfony/security-csrf": "<7.2",
+        "symfony/webpack-encore-bundle": "<2.0",
+        "symfony/flex": "<1.20.0 || >=2.0.0,<2.3.0"
+    },
+    "add-lines": [
+        {
+            "file": "webpack.config.js",
+            "content": "\n    // enables the Symfony UX Stimulus bridge (used in assets/bootstrap.js)\n    .enableStimulusBridge('./assets/controllers.json')",
+            "position": "after_target",
+            "target": ".splitEntryChunks()"
+        },
+        {
+            "file": "assets/app.js",
+            "content": "import './bootstrap.js';",
+            "position": "top",
+            "warn_if_missing": true
+        },
+        {
+            "file": "assets/bootstrap.js",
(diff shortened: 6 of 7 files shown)


symfony/stimulus-bundle/2.19/manifest.json
```diff
@@ -0,0 +1,41 @@
...
+    "bundles": {
+        "Symfony\\UX\\StimulusBundle\\StimulusBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "assets/": "assets/"
+    },
+    "aliases": ["stimulus", "stimulus-bundle"],
+    "conflict": {
+        "symfony/framework-bundle": "<7.2",
+        "symfony/security-csrf": "<7.2",
```
stof, line 11: wouldn't this prevent applying the whole stimulus-bundle recipe when installing StimulusBundle 2.19 with FrameworkBundle 6.4 ?

stof (commented):
should we enable this double-submit validation by default ? For sites that rely on stateful authentication anyway, the session-based CSRF token manager has the benefit of being compatible with form submissions done without JS (either because JS is disabled or because it failed to execute)

symfony/stimulus-bundle/2.19/assets/controllers/csrf_protection_controller.js
```diff
@@ -0,0 +1,35 @@
+document.addEventListener('submit', function (event) {
```
stof, line 1: IMO, it would make sense to ship this controller in a symfony/ux package instead of a recipe, so that it can be properly maintained (for instance, there is no good way to manage security advisories for code provided by recipes, because the advisories would have to apply to projects that were created based on the recipe, not on the recipe itself, as the recipe is not a dependency of the project).

Then, we could have the recipe of that bundle turning on the double-submit CSRF protection.

symfony/stimulus-bundle/2.19/manifest.json
nicolas-grekas: flex skips conflicting recipes by version, so it'll install the version before without any pb

nicolas-grekas:
> should we enable this double-submit validation by default ? For sites that rely on stateful authentication anyway, the session-based CSRF token manager has the benefit of being compatible with form submissions done without JS (either because JS is disabled or because it failed to execute)

We should yes IMHO. Submissions done without JS are also compatible thanks to the `Origin` check, which is enough to accept the submission. The cookie submission becomes required only if a previous submission successfully used it (which means JS was enabled before). The only thing this would break is support for legacy browsers. Given we're talking about new projects here, that's acceptable (and it's still possible to reconfigure so everybody is covered, advanced users should do some more work as expected, but the DX is best by default.)

symfony/stimulus-bundle/2.19/assets/controllers/csrf_protection_controller.js
nicolas-grekas: I'm not sure we should take control of that snippet. By putting it in client's app, we make it easy to tweak. But moving it to vendor dep would require making it parametrized, and this is a path we'd better not take IMHO.
If we have to change the script for security reasons, we can always reference use another data-contoller attribute by default so that older versions become dead-code.

symfony/ux-turbo/2.19/assets/vue/controllers/Hello.vue
```diff
@@ -0,0 +1,9 @@
+<template>
```
smnandre, line 1: Not sure we should add a .vue file here

symfony/ux-turbo/2.19/manifest.json
```diff
@@ -0,0 +1,14 @@
...
+    "conflict": {
+        "symfony/framework-bundle": "<7.2",
+        "symfony/security-csrf": "<7.2"
+    },
+    "add-lines": [
+        {
+            "file": "config/packages/framework.yaml",
+            "position": "after_target",
+            "target": "        csrf_protection:",
+            "content": "            check_header: true"
```
smnandre, line 11: Do i understand this wrong or ... will it fail for every form where Turbo is not enabled ? 

Because this is a very common scenario

symfony/stimulus-bundle/2.19/assets/controllers/csrf_protection_controller.js
```diff
@@ -0,0 +1,35 @@
...
+    if (!csrfCookie) {
+        csrfField.setAttribute('data-csrf-protection-cookie', csrfCookie = csrfField.value);
+    }
+
+
+    if (!/^[-a-zA-Z0-9_]+$/.test(csrfCookie)) {
+        return
+    }
+
+    let csrfToken = btoa(String.fromCharCode.apply(null, (window.crypto || window.msCrypto).getRandomValues(new Uint8Array(18))));
```
smnandre, line 19: As this file will only be loaded when stimulus is used, and stimulus is [not compatible](https://stimulus.hotwired.dev/handbook/installing#browser-support) with IE11 (Turbo [neither](https://github.com/hotwired/hotwire-rails/issues/32#issuecomment-860091010))... 

.. i really think we can drop legacy code :) 


Question: if this recipe is released, is it possible to update it in the following weeks..  or will we need to release a new stimilus-bundle version ?

nicolas-grekas: Last time I checked, it worked on IE10. Since then, I moved this as a lazy stimulus controller.
It's all about the cost to support, and here, if the cost can be low (DX-wise), I'd rather favor a solution that includes as many users of the internet as possible.

smnandre: Ok i won't ask a thrid time 😅  thx

symfony/stimulus-bundle/2.19/assets/controllers/csrf_protection_controller.js
```diff
@@ -0,0 +1,45 @@
...
+
+document.addEventListener('turbo:submit-start', function (event) {
+    let csrfField = event.detail.formSubmission.formElement.querySelector('input[data-controller="csrf-protection"]');
+
+    if (csrfField) {
+        event.detail.formSubmission.fetchRequest.headers[csrfField.getAttribute('data-csrf-protection-cookie')] = csrfField.value;
+    }
+});
+
+document.addEventListener('turbo:submit-end', function (event) {
```
smnandre, line 33: Smart!

symfony/ux-turbo/2.19/assets/vue/controllers/Hello.vue
nicolas-grekas: oops :)

symfony/ux-turbo/2.19/manifest.json
nicolas-grekas: you've got it right - new projects start with turbo enable for every forms

smnandre: Ok.. bold but we'll see :) And worst case a PR is quickly done to add an option :)

smnandre:
⚠️  reminder to not push before 7.2 (i know it's obvious.. but we will release a 2.20 soon so i rather say it ^^)

symfony/stimulus-bundle/2.19/assets/controllers/csrf_protection_controller.js
kbond: We're trying to slim down the number of ux packages to only include ones that have a decent amount of both PHP and JS. Our long(ish)-term plan is to create a sort of recipe system on ux.symfony.com where user's could copy paste code (stimulus controllers, twig components, etc) into their apps. Several of our existing ux packages would then be deprecated in favour of a recipe.

Because the snippets are copy/pasted we don't have to worry about breaking apps when we improve the recipes.

That being said, the security situation isn't something I've thought of with regards to this system.

stof: The issue with copy-paste and that the project in which you paste the code has no relation to the place where you copied it, and so security advisory systems cannot copy the advisories they have for the source to all projects in which you pasted the code (not even considering the fact that advisories are about _affected versions_ of each project).

When the code is just glue code, this is fine as this is generally not the source of a security issues (if some particular usages of a library would be insecure, the library itself would have an advisory). But this case is not about glue code but implementing the (security-related) feature itself.

nicolas-grekas: > Several of our existing ux packages would then be deprecated in favour of a recipe.

That plan is not needed: kepe the packages but turn them into symfony-packs instead, that'll provide the best DX.

About putting that snippet for double-submit, I have the same approach: I don't want to make this a vendor package. I prefer putting people in charge.

The security aspect of this is very limited. The server-side is in control.

symfony/framework-bundle/7.2/config/packages/framework.yaml
```diff
@@ -2,6 +2,10 @@
 framework:
     secret: '%env(APP_SECRET)%'
 
+    form: { csrf_protection: { token_id: submit } }
+    csrf_protection:
+        double_submit_token_ids: [submit, authenticate, logout]
```
Tobion, line 7: I think we need to link to a documentation page here what this means with pros and cons. It's not very self-explaning.

symfony/framework-bundle/7.2/config/packages/framework.yaml
```diff
@@ -2,6 +2,10 @@
 framework:
     secret: '%env(APP_SECRET)%'
 
+    form: { csrf_protection: { token_id: submit } }
```
Tobion, line 5: This makes every form use the same token ID from what I understand. But this is in constrast to the recommendation at https://symfony.com/doc/current/security/csrf.html#csrf-protection-in-symfony-forms

> // an arbitrary string used to generate the value of the token
            // using a different string for each form improves its security
            'csrf_token_id'   => 'task_item',

If that is not true, it should be changed. What's the point of the token ID then at all? Just a way to distinguish different forms?

nicolas-grekas: Thanks for the comments. I agree docs on the topic need to be changed after this PR.
(Yes, token id is to differentiate by use case for each tokens, which means by forms most of the time, but this is not required for the CSRF protection to be effective)

nicolas-grekas:
PR ready for approval.

symfony/stimulus-bundle/2.20/assets/controllers/csrf_protection_controller.js
Kocal, line 1: Regexes `/^[-_/+a-zA-Z0-9]{24,}$` and `/^[-_a-zA-Z0-9]{4,22}$/` are both mentioned three times in the file, maybe we can declare them at the top of the file?

Kocal, line 1: Can we have a comment telling users what this file does? and add a link to the Symfony documentation about `framework.csrf_protection.check_header`?

Kocal (commented):
Minor comments

nicolas-grekas:
Thanks for the review @Kocal, comments resolved.

Kocal (approved):
LGTM

chalasr (dismissed):
Let's see how this goes

symfony/stimulus-bundle/2.20/assets/controllers/csrf_protection_controller.js
```diff
@@ -0,0 +1,60 @@
+nameCheck = /^[-_a-zA-Z0-9]{4,22}$/;
+tokenCheck = /^[-_/+a-zA-Z0-9]{24,}$/;
```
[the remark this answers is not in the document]
Kocal: Good catch, but more `const` or `let` to prevent global access to them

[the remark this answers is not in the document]
jdreesen: Since no `const` or `let` has been used in this file so far, I have also used `var` here ;)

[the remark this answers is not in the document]
stof: I suggest migrating to `const` everywhere in the file, as that's the modern way to write JS code (and Stimulus 3+ does not support old browsers that don't support `const` anyway)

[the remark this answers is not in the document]
nicolas-grekas: I added `var` but I don't think it's needed since we're at the root of the scope
I'm keeping `var` just because it makes the code compatible with even more old browser (eases copy/pastes if one what to put the code outside of a stimulus-loaded module.

symfony/framework-bundle/7.2/config/packages/framework.yaml
```diff
@@ -8,6 +8,11 @@ framework:
     #esi: true
     #fragments: true
 
+    # Enable stateless CSRF protection for forms and logins/logouts
+    form: { csrf_protection: { token_id: submit } }
+    csrf_protection:
+        stateless_token_ids: [submit, authenticate, logout]
```
chalasr, line 14: It might be nice to either have some explanations about what those values are about and where they come from so that one knows what to do or not with this option, or linking to a docs page telling that (once we have one).

chalasr (approved):

symfony/stimulus-bundle/2.20/assets/controllers/csrf_protection_controller.js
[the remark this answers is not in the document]
Kocal: > I added `var` but I don't think it's needed since we're at the root of the scope 

It's needed, otherwise the variable in globally accessible and mutable by other scripts 

Thanks!

symfony/framework-bundle/7.2/config/packages/framework.yaml
nicolas-grekas: Isn't it what L11 is about? Getting into more details might be lengthy and not a fit for this file?

nicolas-grekas:
Friendly up @symfony/mergers

fabpot (approved):

PR #1339, opened 2024-09-23
[api-platform/core] Support for v3.4 and v4

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

* Support for the subtree split (the bundle is now available in a standalone package in addition to as part of core: `api-platform/symfony`
* Simplified configuration 

Closes https://github.com/api-platform/core/issues/6631


stof:
Why is the PR title saying it is about support v4 when the recipe is for 3.4 ?

jrushlow (approved):
Looks good to me - merging will allow us to revert https://github.com/symfony/maker-bundle/pull/1598 over in MakerBundle

chalasr (approved):

chalasr:
Ah, not sure about the failing test https://github.com/symfony/recipes/actions/runs/11029950174/job/30633515716?pr=1339#step:6:275

api-platform/symfony/4.0/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
```
maxhelias, line 1: Shouldn't it conflict with api-platform/core < 4.0 ?

dunglas: WDYT @soyuka?

dunglas:
@chalasr This is because API Platform 4 requires PHP 8.2+. We forgot to bump the minimum version in `composer.json` in 4.0.0 before tagging (but the code was broken with php <8.2) and now Composer always tries to install this broken version with PHP 8.1. Maybe should we delete the 4.0.0 tag, but it will cause other kinds of issues.

For now I suggest to just merge as-is.

dunglas:
@chalasr we finally fixed it by rewriting the tags.

chalasr (approved):

nicolas-grekas (approved):

PR #1341, opened 2024-09-26
[WebpackEncoreBundle] Update recipe for Webpack Encore 5

Kocal:
| Q             | A
| ------------- | ---
| License       | MIT

I'm about to release [Webpack Encore 5](https://github.com/symfony/webpack-encore/milestone/4) soon, so we must update the recipe.
**EDIT:** released!

### DIFF
--- a/symfony/webpack-encore-bundle/2.0/post-install.txt
+++ b/symfony/webpack-encore-bundle/2.0/post-install.txt
@@ -1,5 +1,9 @@
   * Install NPM and run <fg=green>npm install</>
 
-  * Compile your assets: <fg=green>npm run dev</>
-
-  * Or start the development server: <fg=green>npm run watch</>
+  * Compile your assets for development: <fg=green>npm run dev</>
+  
+  * Compile your assets for development and watch for any modifications: <fg=green>npm run watch</>
+  
+  * Or start the development server: <fg=green>npm run dev-server</>
+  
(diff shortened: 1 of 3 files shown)


symfony/webpack-encore-bundle/2.0/post-install.txt
```diff
@@ -2,4 +2,4 @@
 
   * Compile your assets: <fg=green>npm run dev</>
 
-  * Or start the development server: <fg=green>npm run watch</>
+  * Or start the development server: <fg=green>npm run dev-server</>
```
stof, line 5: I would vote for keeping `watch` as the default suggestion as it works out of the box while `dev-server` requires installing an additional dependency.

Kocal: Why not

Kocal: in fact, I've also added `npm run build`, there is no reason to only show `npm run dev` and `npm run watch` (or `npm run dev-server`), since there is 4 scripts in the `package.json`.

Kocal:
The RP can be reviewed now, since Webpack Encore 5 has been released.

chalasr (approved):

Kocal:
Checks are failing because WebpackEncoreBundle does not support PHP 7.4

fabpot (approved):

PR #1351, opened 2024-10-30
Add recipe for `zenstruck/foundry` 2.0

kbond:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

Removes a deprecated config option.

### DIFF
--- a/zenstruck/foundry/2.0/config/packages/zenstruck_foundry.yaml
+++ b/zenstruck/foundry/2.0/config/packages/zenstruck_foundry.yaml
@@ -1,0 +1,5 @@
+when@dev: &dev
+    # See full configuration: https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#full-default-bundle-configuration
+    zenstruck_foundry:
+
+when@test: *dev
--- a/zenstruck/foundry/2.0/manifest.json
+++ b/zenstruck/foundry/2.0/manifest.json
@@ -1,0 +1,12 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "bundles": {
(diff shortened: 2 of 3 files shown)


nikophil (approved):

nicolas-grekas (commented):
Is configuring anything mandatory? Because if the default config file is empty, shouldn't we drop it?

kbond:
There is nothing mandatory to configure, no. I wanted to still show the dev/test YAML anchors to help the user out if they do need to configure something. What about adding a commented out config for common ones?

```yaml
when@dev: &dev
    # See full configuration: https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#full-default-bundle-configuration
    zenstruck_foundry:
        ## configure Faker's locale
        # faker:
                # locale: fr_FR

        ## Always force properties
        # instantiator:
            # always_force_properties: true

        ## Define "global state" which will be available for all your tests
        # global_state:
            # - \App\Tests\Factory\SomeGlobalFactory

        ## Change the default namespace where factories will be created by maker
        # make_factory:
            # default_namespace:    Factory

when@test: *dev
```

nicolas-grekas:
I understand this wish. I'm wondering about our policies for recipes: we try to have bundles that come with no config at all. Of course, we have reasons to configure some things, but when we can, we prefer putting recipes on a diet.

kbond:
> I'm wondering about our policies for recipes: we try to have bundles that come with no config at all.

I 100% agree with this normally but in the case of Foundry, you almost always *need* your dev/test config to match. We do explain this "trick" [in the docs](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#bundle-configuration) but thinking it could be easily missed.

fabpot (approved):

nicolas-grekas (approved):

PR #1353, opened 2024-11-07
Commented for config and removed dependency in package

jennevdmeer:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | 

As suggested by Ryan, I removed these lines to set the build notifications to be disabled by default.

Fixes https://github.com/symfony/webpack-encore/issues/1161

### DIFF
--- a/symfony/webpack-encore-bundle/2.0/webpack.config.js
+++ b/symfony/webpack-encore-bundle/2.0/webpack.config.js
@@ -37,7 +37,8 @@ Encore
      * https://symfony.com/doc/current/frontend.html#adding-more-features
      */
     .cleanupOutputBeforeBuild()
-    .enableBuildNotifications()
+    // Displays build status system notifications to the user
+    // .enableBuildNotifications()
     .enableSourceMaps(!Encore.isProduction())
     // enables hashed filenames (e.g. app.abc123.css)
     .enableVersioning(Encore.isProduction())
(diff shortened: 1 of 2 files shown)


symfony/webpack-encore-bundle/2.0/webpack.config.js
```diff
@@ -37,7 +37,8 @@ Encore
      * https://symfony.com/doc/current/frontend.html#adding-more-features
      */
     .cleanupOutputBeforeBuild()
-    .enableBuildNotifications()
+    // Displays build status system notifications to the user
+    // .enableBuildNotifications()
```
Kocal, line 41: Just two newlines for visual separations with the rest, and that's good to me :)

```suggestion

    // Displays build status system notifications to the user
    // .enableBuildNotifications()

```

Kocal (approved):

Kocal:
Cc @symfony/mergers 🙏🏻

symfony/webpack-encore-bundle/2.0/webpack.config.js
chalasr: Should the comment suggest installing the package this PR removes from the recipe dependencies?

Kocal: Yes, it's already the case, see https://github.com/symfony/webpack-encore/blob/main/lib/plugins/notifier.js#L30 and https://github.com/symfony/webpack-encore/blob/main/lib/features.js#L126-L132

stof:
@jennevdmeer could you do the change requested by @Kocal ?

Nyholm:
I like this idea. I always disable the notifier the first thing I do.

jennevdmeer:
@stof oh thought that was already done, but yes now it is.

lyrixx (approved):

Nyholm:
Thank you

PR #1360, opened 2024-11-20
[WIP] Enable CSRF protection only when security-csrf is installed

nicolas-grekas:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -


chalasr:
Should we have a `csrf` recipe?

nicolas-grekas:
Not sure it's possible because CSRF relies on two independent packages being enabled.
See https://github.com/symfony/symfony/pull/58937 instead.

PR #1364, opened 2024-11-28
Update manifest.json

flovntp:
If you don't set the "entry" port, then Docker would generate a random one and then, when executing Symfony Cli command, it fails. This would let the user to just start your docker container and then update doctrine schema (or run migration) as `5432` is the port configured in the default `.env`

And so, user would do: 
```
symfony server:start -d
docker compose up -d 
symfony console d:s:u --dump-sql
```

| Q             | A
| ------------- | ---
| License       | MIT


fabpot:
We do want a random port to allow people to work on several projects at the same time.

Issue #1365, opened 2024-12-03
Docker compose/Doctrine error with skeleton


chadwcarlson:
Creating a new project from the skeleton, there is an error when running `docker compose up -d` for the first time to make migrations. 

Steps to reproduce:

```bash
# Create a new project 
symfony new symfony_project --upsun
cd symfony_project

# Add bundles relevant to the demo we're producing
symfony composer require doctrine/annotations \
  doctrine/doctrine-bundle \
  doctrine/doctrine-migrations-bundle \
  doctrine/orm nelmio/cors-bundle \
  symfony/doctrine-bridge \
  symfony/html-sanitizer \
  symfony/http-client \
  symfony/intl symfony/monolog-bundle \
  symfony/security-bundle \
  symfony/serializer \
  symfony/twig-bundle \
  symfony/asset-mapper \
  symfony/asset \
  symfony/twig-pack

symfony composer require --dev doctrine/doctrine-fixtures-bundle symfony/maker-bundle

# Create a new entity
symfony console make:entity
```

Then,

```bash
$ docker compose up -d
yaml: line 5: did not find expected tag URI
```

A [quick Google search](https://stackoverflow.com/questions/68938697/docker-compose-yaml-line-9-did-not-find-expected-tag-uri) showed this was in the `!ChangeMe!` value for `POSTGRES_PASSWORD` generated. 

Updating the generated `compose.yaml` file to include quotes on `POSTGRES_PASSWORD` fixes this:

```yaml

services:
###> doctrine/doctrine-bundle ###
  database:
    image: postgres:${POSTGRES_VERSION:-16}-alpine
    environment:
      POSTGRES_DB: ${POSTGRES_DB:-app}
      # You should definitely change the password in production
      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-"!ChangeMe!"}
      POSTGRES_USER: ${POSTGRES_USER:-app}
    healthcheck:
      test: ["CMD", "pg_isready", "-d", "${POSTGRES_DB:-app}", "-U", "${POSTGRES_USER:-app}"]
      timeout: 5s
      retries: 5
      start_period: 60s
    volumes:
      - database_data:/var/lib/postgresql/data:rw
      # You may use a bind-mounted host directory instead, so that it is harder to accidentally remove the volume and lose all your data!
      # - ./docker/db/data:/var/lib/postgresql/data:rw
###< doctrine/doctrine-bundle ###

volumes:
###> doctrine/doctrine-bundle ###
[...]

PR #1366, opened 2024-12-03
Update `docker-compose` for doctrine/doctrine-bundle@2.13

chadwcarlson:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | https://github.com/symfony/recipes/issues/1365

### DIFF
--- a/doctrine/doctrine-bundle/2.13/manifest.json
+++ b/doctrine/doctrine-bundle/2.13/manifest.json
@@ -26,7 +26,7 @@
                 "  environment:",
                 "    POSTGRES_DB: ${POSTGRES_DB:-app}",
                 "    # You should definitely change the password in production",
-                "    POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-!ChangeMe!}",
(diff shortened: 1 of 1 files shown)


doctrine/doctrine-bundle/2.13/manifest.json
```diff
@@ -26,7 +26,7 @@
                 "  environment:",
                 "    POSTGRES_DB: ${POSTGRES_DB:-app}",
                 "    # You should definitely change the password in production",
-                "    POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-!ChangeMe!}",
+                "    POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-\"!ChangeMe!\"}",
```
fabpot, line 29: I suppose that the problem comes from the `!` char? Weirdly, we never had this problem before.

Anyway, what about making it simpler like (cc @dunglas):

```suggestion
                "    POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-ChangeMe}",
```

nicolas-grekas:
Let's assume this is fixed by upgrading docker.

Issue #1370, opened 2024-12-24
CSRF Protection with Fetch Requests


hlecorche:
First of all, thank you for the great work on the CSRF protection functionality in the `csrf_protection_controller.js` script. It’s a fantastic addition, and it works really well for handling traditional form submissions.

However, I’ve encountered an issue when using this CSRF protection in the context of fetch requests (without Turbo). Specifically, I found that the CSRF token is not always applied correctly when sending fetch requests, as the event listener for the form submission may be executed before the CSRF logic in `csrf_protection_controller.js` is applied.

In the current implementation, the CSRF token is injected into the form submission when the submit event is captured :

```javascript
// csrf_protection_controller.js
document.addEventListener('submit', function (event) {
    // CSRF protection logic
});
```

This works fine when the form is submitted in the traditional way. However, when using fetch (with custom AJAX logic), I encountered an issue where the CSRF protection logic is executed after the fetch request is sent, which results in the CSRF token not being updated in time.

Here’s an example that illustrates the problem :

```javascript
document.addEventListener('submit', function (event) {
	if (event.target.matches('form[data-toggle="ajax-form"]')) {
	  event.preventDefault()
	  // Send request with fetch
	}
})
```

Depending on the order of event listener registration, the event that triggers the fetch request can be executed before the csrf_protection_controller.js event listener. This is exactly what happens in my case, as my event listener is registered during the DOMContentLoaded event, which causes it to run before the CSRF protection logic is applied.

To fix this issue, I used the capture phase (true as the third argument in addEventListener) to ensure that the CSRF protection logic is executed before the fetch request is sent.

```javascript
// csrf_protection_controller.js
document.addEventListener('submit', function (event) {
    // CSRF protection logic
}, true);  // Using the capture phase to ensure this runs first
```

Why is the capture phase not used by default in the script? I believe this would help ensure the CSRF token is updated before the form is submitted, especially in cases where fetch is used.

Additionally, I noticed that the CSRF logic in csrf_protection_controller.js is tightly coupled with form submissions with Turbo. However, when using fetch (without Turbo), the same functionality is required — updating the CSRF token — but the existing solution works specifically for Turbo requests. This creates the need for duplication of code to handle CSRF protection in the same way for fetch requests.

[...]

PR #1375, opened 2025-01-08
[stimulus-bundle] Export CSRF protection helpers

hlecorche:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Here is my first proposal addressing #1370:

* Fixes the issue where the CSRF token is updated after the form is submitted (e.g., when using fetch requests) ;
* Avoids code duplication when reusing this logic in multiple places ;
* Replaces `var` by `const` or `let` for better modern JavaScript practices.

This code can be used without Turbo. Here's an example:

```js
import { generateCsrfHeaders, removeCsrfToken } from './csrf_protection.js';

// ...

document.addEventListener('before-event-with-jquery-or-custom-fetch-event', function (event) {
	Object.entries(generateCsrfHeaders(event.detail.form)).forEach(([name, value]) => {
	    // add header
	});
});

document.addEventListener('after-event-with-jquery-or-custom-fetch-event', function (event) {
	removeCsrfToken(event.detail.form);
});
```

It can also be used without relying on the form's submit event :

```js
import { generateCsrfToken, generateCsrfHeaders, removeCsrfToken } from './csrf_protection.js';

// ...

link.addEventListener('click', function (event) {
	const form = document.getElementById('my-form');
	generateCsrfToken(form);
	// Send form with fetch (or jquery) with generateCsrfHeaders and removeCsrfToken functions
});
```

I have made these changes in version 2.20 (the existing version). However, since the behavior of the files is modified, I assume it is necessary to create a new version.

### DIFF
--- a/symfony/stimulus-bundle/2.20/assets/controllers/csrf_protection_controller.js
+++ b/symfony/stimulus-bundle/2.20/assets/controllers/csrf_protection_controller.js
@@ -1,61 +1,79 @@
-var nameCheck = /^[-_a-zA-Z0-9]{4,22}$/;
-var tokenCheck = /^[-_/+a-zA-Z0-9]{24,}$/;
+const nameCheck = /^[-_a-zA-Z0-9]{4,22}$/;
+const tokenCheck = /^[-_\/+a-zA-Z0-9]{24,}$/;
 
 // Generate and double-submit a CSRF token in a form field and a cookie, as defined by Symfony's SameOriginCsrfTokenManager
 document.addEventListener('submit', function (event) {
-    var csrfField = event.target.querySelector('input[data-controller="csrf-protection"], input[name="_csrf_token"]');
-
-    if (!csrfField) {
-        return;
-    }
-
-    var csrfCookie = csrfField.getAttribute('data-csrf-protection-cookie-value');
-    var csrfToken = csrfField.value;
-
-    if (!csrfCookie && nameCheck.test(csrfToken)) {
-        csrfField.setAttribute('data-csrf-protection-cookie-value', csrfCookie = csrfToken);
-        csrfField.defaultValue = csrfToken = btoa(String.fromCharCode.apply(null, (window.crypto || window.msCrypto).getRandomValues(new Uint8Array(18))));
-        csrfField.dispatchEvent(new Event('change', {bubbles: true}));
-    }
-
-    if (csrfCookie && tokenCheck.test(csrfToken)) {
-        var cookie = csrfCookie + '_' + csrfToken + '=' + csrfCookie + '; path=/; samesite=strict';
-        document.cookie = window.location.protocol === 'https:' ? '__Host-' + cookie + '; secure' : cookie;
-    }
-});
+    generateCsrfToken(event.target);
+}, true);
 
 // When @hotwired/turbo handles form submissions, send the CSRF token in a header in addition to a cookie
 // The `framework.csrf_protection.check_header` config option needs to be enabled for the header to be checked
 document.addEventListener('turbo:submit-start', function (event) {
-    var csrfField = event.detail.formSubmission.formElement.querySelector('input[data-controller="csrf-protection"], input[name="_csrf_token"]');
-
-    if (!csrfField) {
-        return;
-    }
-
-    var csrfCookie = csrfField.getAttribute('data-csrf-protection-cookie-value');
-
-    if (tokenCheck.test(csrfField.value) && nameCheck.test(csrfCookie)) {
-        event.detail.formSubmission.fetchRequest.headers[csrfCookie] = csrfField.value;
-    }
+    const h = generateCsrfHeaders(event.detail.formSubmission);
+    Object.keys(h).map(function (k) {
+        event.detail.formSubmission.fetchRequest.headers[k] = h[k];
+    });
 });
 
 // When @hotwired/turbo handles form submissions, remove the CSRF cookie once a form has been submitted
 document.addEventListener('turbo:submit-end', function (event) {
-    var csrfField = event.detail.formSubmission.formElement.querySelector('input[data-controller="csrf-protection"], input[name="_csrf_token"]');
+    removeCsrfToken(event.detail.formSubmission.formElement);
+});
+
+export function generateCsrfToken (formElement) {
+    const csrfField = formElement.querySelector('input[data-controller="csrf-protection"], input[name="_csrf_token"]');
 
     if (!csrfField) {
         return;
     }
 
-    var csrfCookie = csrfField.getAttribute('data-csrf-protection-cookie-value');
+    let csrfCookie = csrfField.getAttribute('data-csrf-protection-cookie-value');
+    let csrfToken = csrfField.value;
+
+    if (!csrfCookie && nameCheck.test(csrfToken)) {
+        csrfField.setAttribute('data-csrf-protection-cookie-value', csrfCookie = csrfToken);
+        csrfField.defaultValue = csrfToken = btoa(String.fromCharCode.apply(null, (window.crypto || window.msCrypto).getRandomValues(new Uint8Array(18))));
+        csrfField.dispatchEvent(new Event('change', { bubbles: true }));
+    }
+
+    if (csrfCookie && tokenCheck.test(csrfToken)) {
+        const cookie = csrfCookie + '_' + csrfToken + '=' + csrfCookie + '; path=/; samesite=strict';
+        document.cookie = window.location.protocol === 'https:' ? '__Host-' + cookie + '; secure' : cookie;
+    }
+}
+
+export function generateCsrfHeaders (formElement) {
+    const headers = {};
+    const csrfField = formElement.querySelector('input[data-controller="csrf-protection"], input[name="_csrf_token"]');
+
+    if (!csrfField) {
+        return headers;
+    }
+
+    const csrfCookie = csrfField.getAttribute('data-csrf-protection-cookie-value');
 
     if (tokenCheck.test(csrfField.value) && nameCheck.test(csrfCookie)) {
-        var cookie = csrfCookie + '_' + csrfField.value + '=0; path=/; samesite=strict; max-age=0';
+        headers[csrfCookie] = csrfField.value;
+    }
+
+    return headers;
+}
+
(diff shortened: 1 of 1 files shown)


ajgarlag:
I've tested it with a POST form in a turbo-frame that was failing with the current csrf_protection_controller.js and it worked flawlessly.

Thanks @hlecorche.

smnandre:
Do we really need two distinct files for this ? :|

nicolas-grekas:
Definitely one file please here.

hlecorche:
@nicolas-grekas I have merged everything into a single file.

Here are the modified examples :

This code can be used without Turbo. Here's an example:

```js
import { generateCsrfHeaders, removeCsrfToken } from '../controllers/csrf_protection_controller';

// ...

document.addEventListener('before-event-with-jquery-or-custom-fetch-event', function (event) {
	Object.entries(generateCsrfHeaders(event.detail.form)).forEach(([name, value]) => {
	    // add header
	});
});

document.addEventListener('after-event-with-jquery-or-custom-fetch-event', function (event) {
	removeCsrfToken(event.detail.form);
});
```

It can also be used without relying on the form's submit event :

```js
import { generateCsrfToken, generateCsrfHeaders, removeCsrfToken } from '../controllers/csrf_protection_controller';

// ...

link.addEventListener('click', function (event) {
	const form = document.getElementById('my-form');
	generateCsrfToken(form);
	// Send form with fetch (or jquery) with generateCsrfHeaders and removeCsrfToken functions
});
```

symfony/stimulus-bundle/2.20/assets/controllers/csrf_protection_controller.js
```diff
@@ -1,60 +1,78 @@
...
 // Generate and double-submit a CSRF token in a form field and a cookie, as defined by Symfony's SameOriginCsrfTokenManager
 document.addEventListener('submit', function (event) {
-    var csrfField = event.target.querySelector('input[data-controller="csrf-protection"]');
+    generateCsrfToken(event.target);
+}, true);
+
+// When @hotwired/turbo handles form submissions, send the CSRF token in a header in addition to a cookie
+// The `framework.csrf_protection.check_header` config option needs to be enabled for the header to be checked
+document.addEventListener('turbo:submit-start', function (event) {
+    Object.entries(generateCsrfHeaders(event.detail.formSubmission.formElement)).forEach(([name, value]) => {
```
nicolas-grekas, line 12: Object.entries isn't available with IE11, while I think the previous script was working with it.
`export`  is also a feature not compatible with IE11 so that maybe the script didn't work with it already, but maybe [the eshim](https://github.com/guybedford/es-module-shims) made in work in the end?

What I mean is: if the snippet worked with IE11, let's not break IE11 support. Any idea of the status of this?

Kocal: Turbo and Stimulus do not support IE 11, so I believe it is fine to use more modern API's

smnandre: No IE support for Stimulus, Turbo, import map, nor the importmap polyfill  (see [here](https://github.com/guybedford/es-module-shims/issues/14#issuecomment-493261534))

No consequence / difference with today ☺️

nicolas-grekas: this comment is old, the polyfill does support IE11 (at least the .dist version, I've tried already)

nicolas-grekas:
Thank you @hlecorche

hlecorche:
Thanks @nicolas-grekas

symfony/stimulus-bundle/2.20/assets/controllers/csrf_protection_controller.js
```diff
@@ -1,61 +1,79 @@
-var nameCheck = /^[-_a-zA-Z0-9]{4,22}$/;
-var tokenCheck = /^[-_/+a-zA-Z0-9]{24,}$/;
+const nameCheck = /^[-_a-zA-Z0-9]{4,22}$/;
+const tokenCheck = /^[-_\/+a-zA-Z0-9]{24,}$/;
 
 // Generate and double-submit a CSRF token in a form field and a cookie, as defined by Symfony's SameOriginCsrfTokenManager
 document.addEventListener('submit', function (event) {
-    var csrfField = event.target.querySelector('input[data-controller="csrf-protection"], input[name="_csrf_token"]');
+    generateCsrfToken(event.target);
+}, true);
```
nicolas-grekas, line 7: sorry I missed this during the review: why adding true as second argument here @hlecorche ?

stof: Doing that in the capture phase means that this listener (attached on the root of the document) runs *before* the listeners attached directly on the form. This ensures that listeners attached on a form to perform custom submission logic (using Ajax for instance) will have the value for the generated CSRF token (instead of generating it too late): https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Scripting/Event_bubbling#event_capture

nicolas-grekas: and this fixes https://github.com/symfony/symfony/issues/59571

hlecorche: Here’s an example that illustrates the problem :

```js
document.addEventListener('submit', function (event) {
	if (event.target.matches('form[data-toggle="ajax-form"]')) {
	  event.preventDefault()
	  // Send request with fetch
	}
})
```

Depending on the order of event listener registration, the event that triggers the fetch request can be executed before the csrf_protection_controller.js event listener. This is exactly what happens in my case, as my event listener is registered during the DOMContentLoaded event, which causes it to run before the CSRF protection logic is applied.

To fix this issue, I used the capture phase (true as the third argument in addEventListener) to ensure that the CSRF protection logic is executed before the fetch request is sent.

symfony/stimulus-bundle/2.20/assets/controllers/csrf_protection_controller.js
stof: The `dist` file of `es-module-shims` uses `async`/`await` among other things. So it definitely does not support IE11 without transpiling. And it also uses some modern methods that won't work without polyfiling them.

Projects needing to support IE11 probably have an infrastructure to transpile code to ES5 and load polyfills, which would work for this file as well.

symfony/stimulus-bundle/2.20/assets/controllers/csrf_protection_controller.js
```diff
@@ -1,60 +1,78 @@
...
+        event.detail.formSubmission.fetchRequest.headers[name] = value;
+    });
+});
+
+// When @hotwired/turbo handles form submissions, remove the CSRF cookie once a form has been submitted
+document.addEventListener('turbo:submit-end', function (event) {
+    removeCsrfToken(event.detail.formSubmission.formElement);
+});
+
+export function generateCsrfToken (formElement) {
```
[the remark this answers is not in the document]
hlecorche: It's for this reason that the first version of my pull request included two files (Stimulus and utilities)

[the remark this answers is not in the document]
hlecorche: But yes @stof , the goal is to make them reusable elsewhere :

```js
import { generateCsrfHeaders, removeCsrfToken } from '../controllers/csrf_protection_controller';

// ...

document.addEventListener('before-event-with-jquery-or-custom-fetch-event', function (event) {
	Object.entries(generateCsrfHeaders(event.detail.form)).forEach(([name, value]) => {
	    // add header
	});
});

document.addEventListener('after-event-with-jquery-or-custom-fetch-event', function (event) {
	removeCsrfToken(event.detail.form);
});
```

It can also be used without relying on the form's submit event :

```js
import { generateCsrfToken, generateCsrfHeaders, removeCsrfToken } from '../controllers/csrf_protection_controller';

// ...

link.addEventListener('click', function (event) {
	const form = document.getElementById('my-form');
	generateCsrfToken(form);
	// Send form with fetch (or jquery) with generateCsrfHeaders and removeCsrfToken functions
});
```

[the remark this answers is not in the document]
nicolas-grekas: maybe we can do something about by leveraging the export default entry to register the listeners when stimulus loads the controller?
having one file is desirable to me

symfony/stimulus-bundle/2.20/assets/controllers/csrf_protection_controller.js
```diff
@@ -1,60 +1,78 @@
...
         document.cookie = window.location.protocol === 'https:' ? '__Host-' + cookie + '; secure' : cookie;
     }
-});
+}
 
-// When @hotwired/turbo handles form submissions, send the CSRF token in a header in addition to a cookie
-// The `framework.csrf_protection.check_header` config option needs to be enabled for the header to be checked
-document.addEventListener('turbo:submit-start', function (event) {
-    var csrfField = event.detail.formSubmission.formElement.querySelector('input[data-controller="csrf-protection"]');
+export function generateCsrfHeaders (formElement) {
```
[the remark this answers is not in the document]
smnandre: More an more I'm thinking such  method / exports should not be there. 

We provide  "plug n play" CSRF  here, but we should probably not encourage using this _recipe file_ as  vendor/ lib code.

We can remove it / change it at any time in future versions

symfony/stimulus-bundle/2.20/assets/controllers/csrf_protection_controller.js
```diff
@@ -1,61 +1,79 @@
-var nameCheck = /^[-_a-zA-Z0-9]{4,22}$/;
-var tokenCheck = /^[-_/+a-zA-Z0-9]{24,}$/;
+const nameCheck = /^[-_a-zA-Z0-9]{4,22}$/;
+const tokenCheck = /^[-_\/+a-zA-Z0-9]{24,}$/;
```
whataboutpereira, line 2: I believe escaping a slash inside [] isn't needed and it's triggering no-useless-escape in eslint.

PR #1379, opened 2025-02-11
[symfony/security-bundle] set default algorithm to plaintext

JohJohan:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

I think its nice to have default hashing set to `plaintext` and show the option to have lowest configuration the docs also start with mention of `plaintext` see: https://symfony.com/doc/6.4/security/passwords.html#:~:text=algorithm%3A%20plaintext%20%23%20disable%20hashing%20(only%20do%20this%20in%20tests!)

### DIFF
--- a/symfony/security-bundle/6.4/config/packages/security.yaml
+++ b/symfony/security-bundle/6.4/config/packages/security.yaml
@@ -30,10 +30,11 @@ when@test:
         password_hashers:
             # By default, password hashers are resource intensive and take time. This is
             # important to generate secure password hashes. In tests however, secure hashes
-            # are not important, waste resources and increase test times. The following
-            # reduces the work factor to the lowest possible values.
+            # are not important, waste resources and increase test times.
             Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface:
-                algorithm: auto
-                cost: 4 # Lowest possible value for bcrypt
-                time_cost: 3 # Lowest possible value for argon
-                memory_cost: 10 # Lowest possible value for argon
+                algorithm: plaintext # disable hashing all together
+                # You can also use the following configuration to use the lowest possible values for bcrypt and argon:
+                # algorithm: auto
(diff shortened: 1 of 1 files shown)


94noni (approved):
But for 7.3 instead?

nicolas-grekas:
I'd need a better rationale to be convinced. plaintext is bad practice to me, even in this case, because some ppl will copy/paste and :boom:

94noni:
The doc state this as a « tip »
Perhaps we should then remove completly remove plaintext from the recipe and only let ppl discover through the doc ?

JohJohan:
Yeah interesting thing to think about. I agree with the copy paste and can be damaging, although we have explicitly set it in test env and have the default case set as well.

In our case we are working with Symfony for 8 years and only recently discover this feature and made our pipelines faster.


We could find middle ground and comment out plaintext and show the other options like we had before?

94noni:
Or perhaps we can close here, add link to spécial doc https://symfony.com/doc/current/security/passwords.html
There is a part with Tip detailing test hash

kbond:
See https://github.com/symfony/recipes/issues/1024 and https://github.com/symfony/recipes/pull/1026 for some context as to why we didn't choose `plaintext`. We dug in at the time and saw [virtually no performance difference between plaintext and this method](https://github.com/symfony/recipes/pull/1026#issuecomment-990438058).

JohJohan:
If there is not much benefit i agree we should probably not promote it and have a closer testing experience with hashed passwords.

I did test run myself with `plaintext`:
```
Time: 00:15.404, Memory: 166.41 MB
Time: 00:15.100, Memory: 166.41 MB
Time: 00:15.844, Memory: 166.41 MB
```

with lowest values:
```
Time: 00:15.949, Memory: 166.41 MB
Time: 00:16.183, Memory: 166.41 MB
Time: 00:15.271, Memory: 166.41 MB
```

PR #1380, opened 2025-02-17
Cleaned csrf_protection_controller for very basic eslint rules

lyrixx:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  |


whataboutpereira:
ESLint suggests using `.forEach` instead of `.map` there, so perhaps ignoring the error isn't the best path.

PR #1382, opened 2025-02-26
Update the callable-qa workflow to avoid GHA deprecations

stof:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

PR #1384, opened 2025-03-06
Update "runs-on" to use ubuntu-latest instead of deprecated 20.04

aivus:
| Q             | A
| ------------- | ---
| License       | MIT

Update GitHub runner versions to use `ubuntu-latest`

See https://github.com/symfony/recipes/pull/1382


fabpot (approved):

stof:
@fabpot can you merge this as it requires bypassing the required check ?
Just like #1382, the failure is because our CI job does not handle properly the case where none of the recipes are modified.

PR #1386, opened 2025-03-09
[Webpack-Encore] Remove "ux" alias

smnandre:
| Q             | A
| ------------- | ---
| License       | MIT

Currently if someones types "composer req ux" after a base symfony install ... he ends up with webpack encore.

Worse, the same thing happens if user installed symfony "--webapp", as they will end up with both AssetMapper and Webpack Encore installed.

Finally, and the initial reason I checked this: we will soon need the "ux" alias for the UX Bundle.

cc @Kocal 

PS: not sure if this would require a release to be effective... and hope it does not 😅

### DIFF
--- a/symfony/webpack-encore-bundle/2.0/manifest.json
+++ b/symfony/webpack-encore-bundle/2.0/manifest.json
@@ -8,7 +8,7 @@
         "package.json": "package.json",
         "webpack.config.js": "webpack.config.js"
     },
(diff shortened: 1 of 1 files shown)


Kocal:
For more context, `ux` alias has been added in #891, when Symfony UX was just a few months old.

It makes sense to now remove this alias from Encore, also don't forget about other `manifest.json` files: 
image

fabpot (approved):

PR #1391, opened 2025-03-16
Remove configuration defaults for the container

fabpot:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

Related to https://github.com/symfony/symfony/pull/59987


nicolas-grekas (approved):

bartholdbos:
Maybe this needs some more documentation for messages for example. The documentation for 7.3 currently doesn't show that you need to add the AsMessage attribute. With this change all new projects should use the attribute or their messages get registered as services potentially breaking the container.

PR #1392, opened 2025-03-22
Update security.yaml: Simplifying regex

ThomasLandauer:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | Sister-PR of https://github.com/symfony/symfony-docs/pull/20794#pullrequestreview-2707869766

### DIFF
--- a/symfony/security-bundle/6.4/config/packages/security.yaml
+++ b/symfony/security-bundle/6.4/config/packages/security.yaml
@@ -7,7 +7,7 @@ security:
         users_in_memory: { memory: null }
     firewalls:
         dev:
-            pattern: ^/(_(profiler|wdt)|css|images|js)/
+            pattern: ^/(_profiler|_wdt|assets)/
             security: false
         main:
             lazy: true


symfony/security-bundle/6.4/config/packages/security.yaml
```diff
@@ -7,7 +7,7 @@ security:
         users_in_memory: { memory: null }
     firewalls:
         dev:
-            pattern: ^/(_(profiler|wdt)|css|images|js)/
+            pattern: ^/(_profiler|_wdt|assets)/
```
wouterj, line 10:
```suggestion
            pattern: ^/_profiler|_wdt|assets|build/
```

There is no need for parenthesis if the union is the whole regex pattern.

Also, I think it's good to add `build` from Webpack Encore to the default pattern.

wouterj (changes requested):
This change makes sense to me. The `css|images|js` pattern originates from Assetic (https://github.com/symfony/symfony-standard/pull/109), so replacing it with the Asset mapper and webpack encore variants sound like a good idea.

However, [as per our update policy](https://github.com/symfony/recipes?tab=readme-ov-file#updating-recipes), changes like this should be applied to the upcoming version (7.3) only. To do this, copy the whole 6.4 directory over as 7.3 and edit this file in that directory only.

ThomasLandauer:
OK, there you go: https://github.com/symfony/recipes/pull/1395

PR #1395, opened 2025-03-23
Update security.yaml: Simplifying regex (#2)

ThomasLandauer:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | Sister-PR of https://github.com/symfony/symfony-docs/pull/20794#pullrequestreview-2707869766

As requested at https://github.com/symfony/recipes/pull/1392#pullrequestreview-2708354322

### DIFF
--- a/symfony/security-bundle/7.3/config/packages/security.yaml
+++ b/symfony/security-bundle/7.3/config/packages/security.yaml
@@ -1,0 +1,39 @@
+security:
+    # https://symfony.com/doc/current/security.html#registering-the-user-hashing-passwords
+    password_hashers:
+        Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface: 'auto'
+    # https://symfony.com/doc/current/security.html#loading-the-user-the-user-provider
+    providers:
+        users_in_memory: { memory: null }
+    firewalls:
+        dev:
+           pattern: ^/_profiler|_wdt|assets|build/ # `assets` is for AssetMapper; `build` is for Webpack Encore
+            security: false
+        main:
+            lazy: true
+            provider: users_in_memory
+
+            # activate different ways to authenticate
+            # https://symfony.com/doc/current/security.html#the-firewall
+
+            # https://symfony.com/doc/current/security/impersonating_user.html
+            # switch_user: true
+
+    # Easy way to control access for large sections of your site
+    # Note: Only the *first* access control that matches will be used
+    access_control:
+        # - { path: ^/admin, roles: ROLE_ADMIN }
+        # - { path: ^/profile, roles: ROLE_USER }
+
+when@test:
+    security:
+        password_hashers:
+            # By default, password hashers are resource intensive and take time. This is
+            # important to generate secure password hashes. In tests however, secure hashes
+            # are not important, waste resources and increase test times. The following
+            # reduces the work factor to the lowest possible values.
+            Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface:
+                algorithm: auto
+                cost: 4 # Lowest possible value for bcrypt
+                time_cost: 3 # Lowest possible value for argon
+                memory_cost: 10 # Lowest possible value for argon
--- a/symfony/security-bundle/7.3/config/routes/security.yaml
+++ b/symfony/security-bundle/7.3/config/routes/security.yaml
@@ -1,0 +1,3 @@
+_security_logout:
+    resource: security.route_loader.logout
+    type: service
--- a/symfony/security-bundle/7.3/manifest.json
+++ b/symfony/security-bundle/7.3/manifest.json
@@ -1,0 +1,12 @@
+{
+    "bundles": {
+        "Symfony\\Bundle\\SecurityBundle\\SecurityBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "aliases": ["security"],
+    "conflict": {
+        "symfony/framework-bundle": "<5.3"
+    }
+}


symfony/security-bundle/7.3/config/packages/security.yaml
```diff
@@ -0,0 +1,39 @@
+security:
+    # https://symfony.com/doc/current/security.html#registering-the-user-hashing-passwords
+    password_hashers:
+        Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface: 'auto'
+    # https://symfony.com/doc/current/security.html#loading-the-user-the-user-provider
+    providers:
+        users_in_memory: { memory: null }
+    firewalls:
+        dev:
+            pattern: ^/_profiler|_wdt|assets|build/ # `assets` is for AssetMapper; `build` is for Webpack Encore
```
mtarld, line 10:
```suggestion
            pattern: ^/(_profiler|_wdt|assets|build)/ # `assets` is for AssetMapper; `build` is for Webpack Encore
```
Shouldn't this be with parentheses instead?

wouterj: Not needed if in the union is the full pattern.

mtarld: Yes, but here we have `^/` and `/` that are not part of the union right?

ThomasLandauer: I think @mtarld is right; https://www.php.net/manual/en/regexp.reference.alternation.php doesn't say anything about the alternative branches stopping at special characters. So currently, the 4 alternatives are:
* `^/_profiler`
* `_wdt`
* `assets`
* `build/`

But it "works" nevertheless, since `_wdt` matches anything like `/_wdt/whatever...` ;-)

So the way to go IMO is to bring back the parentheses.

However, one thing is irritating me: While testing this, I commented out that line in my `security.php`, and the web debug toolbar is still showing (and all assets loading) in DEV environment.

xabbuh: You commented out that single line or the whole dev firewall?

wouterj: Oh correct, parenthesis are required because it's not the full pattern. The `/.../` confused me into thinking that they were pattern delimiters 😅

stof: @ThomasLandauer try having an access_control rule requiring some authentication for all pages except the login page. When accessing the login page, the toolbar would not load as it would suddenly require authentication. That's why we have this `dev` firewall disabling authentication for dev URLs.

ThomasLandauer: OK, thanks, so this line is only needed if there's an `access_control`? And you're assuming that most people have one?
So what about adding some comment here, like:
> Exception for the web debug toolbar and assets (only needed if you have a broad `access_control`):

About the (missing) delimiter: This irritated me too. The best solution would probably be to include the delimiter here in the string. But that would be a breaking change, which is probably not worth it. But what about at least adding some comment like:
> The delimiter `xxx` is added automatically

This might be helpful if the delimiter happens to appear in somebody's path (cause they'd have to escape it then).
=> So which delimiter is it?

stof: If you don't have any access control rule enforcing a broad requirement, the fact that the toolbar would be matched by the main firewall may not prevent it from loading.
However, the use case of securing the whole project under login (except a few page like the login and password recovery pages) is a common use case, and we want the toolbar and profiler to work out of the box (otherwise, things would be confusing for users). That's why we have this `dev` firewall disabling any authentication for the dev tooling pages.

> => So which delimiter is it?

the PathRequestMatcher uses `{}` delimiters, which don't require special escaping inside the regex (thanks to the balancing of braces, the engine can distinguish whether a `}` is closing a repetition quantifier or is the end delimiter of the regex)

stof:
@ThomasLandauer could you fix this PR ? We definitely want our `dev` firewall to use a pattern anchored at the beginning, to avoid risking matching URLs of the project that might contain `assets` or `build` somewhere else in the URL (which would disable security check for them).

ThomasLandauer:
Sure, I just dropped in the two comments :-)

symfony/security-bundle/7.3/config/packages/security.yaml
```diff
@@ -0,0 +1,39 @@
+security:
+    # https://symfony.com/doc/current/security.html#registering-the-user-hashing-passwords
+    password_hashers:
+        Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface: 'auto'
+    # https://symfony.com/doc/current/security.html#loading-the-user-the-user-provider
+    providers:
+        users_in_memory: { memory: null }
+    firewalls:
+        dev: # Exception for the web debug toolbar and assets (only needed if you have a broad `access_control`):
```
javiereguiluz, line 9: I'd tweak these comments and its formatting too. Instead of a single long line, we could add the comment above the related config:

```yaml
firewalls:
    # Disable security for dev tools and static assets (only needed if access_control is broad)
    dev:
        # 'assets/' is for AssetMapper, 'build/' for Webpack Encore
        # (note: no regex delimiters needed; Symfony adds them automatically)
        pattern: ^/(_profiler|_wdt|assets|build)
```

symfony/security-bundle/7.3/config/packages/security.yaml
```diff
@@ -0,0 +1,39 @@
...
+    # Easy way to control access for large sections of your site
+    # Note: Only the *first* access control that matches will be used
+    access_control:
+        # - { path: ^/admin, roles: ROLE_ADMIN }
+        # - { path: ^/profile, roles: ROLE_USER }
+
+when@test:
+    security:
+        password_hashers:
+            # By default, password hashers are resource intensive and take time. This is
```
javiereguiluz, line 31: I feel this comment is too much. What about making it more concise?

```yaml
password_hashers:
    # Password hashers are resource-intensive by design to ensure security.
    # In tests, it's safe to reduce their cost to improve performance.
    Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface:
        algorithm: auto
        # ...
```

```

File: symfony/security-bundle/7.3/config/packages/security.yaml [hunk 4]
```diff
@@ -0,0 +1,39 @@
+security:
+    # https://symfony.com/doc/current/security.html#registering-the-user-hashing-passwords
+    password_hashers:
+        Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface: 'auto'
+    # https://symfony.com/doc/current/security.html#loading-the-user-the-user-provider
+    providers:
```
javiereguiluz, line 6: I'd add a `\n` before each of the main sections of this file (`providers`, `firewalls`, etc. ) to make it easier to scan.

ThomasLandauer:
I added some punctuation too. Looks like there are no clear rules (throughout the docs) for comments; but I think omitting any punctuation makes them harder to read.

symfony/security-bundle/7.3/config/packages/security.yaml
```diff
@@ -0,0 +1,42 @@
...
+        users_in_memory: { memory: null }
+
+    firewalls:
+        # Disable security for dev tools and static assets (only needed if access_control is broad):
+        dev:
+            # 'assets/' is for AssetMapper, 'build/' for Webpack Encore.
+            # (Note: no regex delimiters needed; Symfony adds `{}` automatically.)
+            pattern: ^/(_profiler|_wdt|assets|build)/
```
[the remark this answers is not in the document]
stof: in dev, AssetMapper loads the file through the kernel, with a listener resolving them. You don't want the assets to be registered as the "last URL being accessed" under the firewall, saved in the session for the redirection after authentication.
Adding it for Webpack Encore is a safety measure avoiding the same kind of issue in case of a 404 on the built asset (for instance due to versioning), even though non-404 responses for Encore should be managed by the webserver directly (same applies for AssetMapper in production mode btw)

symfony/security-bundle/7.3/config/packages/security.yaml
```diff
@@ -2,36 +2,39 @@ security:
...
     # https://symfony.com/doc/current/security.html#loading-the-user-the-user-provider
     providers:
         users_in_memory: { memory: null }
+
     firewalls:
+        # Disable security for dev tools and static assets (only needed if access_control is broad):
         dev:
-            pattern: ^/(_(profiler|wdt)|css|images|js)/
+            # 'assets/' is for AssetMapper, 'build/' for Webpack Encore.
+            # (Note: no regex delimiters needed; Symfony adds `{}` automatically.)
```
nicolas-grekas:
```suggestion
        dev:
            # Ensure dev tools and static assets are always allowed
```

nicolas-grekas: >there's no need to escape those slashes.

 yet if one does it still works, isn't it?
then we don't need to tell - config files shouldn't be documentation

stof: > Indeed, but there is no documentation for this - so there is no better place ;-)

The better place is contributing to the documentation so that there is documentation for this.

PR #1396, opened 2025-03-26
Add server option for dev & prod in asset_mapper.yaml

cadot-eu:
for remove problem with access files in dev mode

| Q             | A
| ------------- | ---
| License       | MIT


stof:
Why not relying on the default value of this parameter, which is `%kernel.debug%` ?

cadot-eu:
> Why not relying on the default value of this parameter, which is `%kernel.debug%` ?

good idea ;-) it's simple and clean

PR #1398, opened 2025-04-08
Renaming `bootstrap.js` to `stimulus_bootstrap.js`

ThomasLandauer:
| Q             | A
| ------------- | ---
| License       | MIT

Second attempt to finally fix https://github.com/symfony/webpack-encore-bundle/issues/150 and https://github.com/symfony/webpack-encore-bundle/issues/160

### DIFF
--- a/symfony/stimulus-bundle/2.20/manifest.json
+++ b/symfony/stimulus-bundle/2.20/manifest.json
@@ -21,7 +21,7 @@
         },
         {
             "file": "assets/app.js",
-            "content": "import './bootstrap.js';",
+            "content": "import './stimulus_bootstrap.js';",
             "position": "top",
             "warn_if_missing": true
         },


symfony/stimulus-bundle/2.20/manifest.json
```diff
@@ -21,7 +21,7 @@
         },
         {
             "file": "assets/app.js",
-            "content": "import './bootstrap.js';",
```
stof, line 24: This file is mentioned in the comment which is part of the content of the previous block adding lines in `webpack.config.js`, so you need to update the name there as well.

And the next 2 `add-lines` rule which were adding things in this `asset/bootstrap.js` file need to be updated to add them in the new file name instead.

stof (changes requested):

ThomasLandauer:
So I should create version 2.21?

stof:
The new version of the recipe should be based on the latest version of the bundle. so it should be `2.24`

ThomasLandauer:
OK, thanks, then I'm closing here in favor of https://github.com/symfony/recipes/pull/1399

PR #1401, opened 2025-04-12
Add PHPUnit 11 recipe, treat PHPUnitBridge only as a PHPUnit extension

wouterj:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | I'll take care of this once we decide the direction for PHPUnit integration

This is a continuation of the effort to move the community towards PHPunit 11's deprecation features, in favor of the Symfony PHPUnit bridge. This is primarily based on what I submitted to the Symfony Demo a while ago: https://github.com/symfony/demo/pull/1549

This PR makes a couple important changes:

* **Add a recipe for PHPUnit 11.1+**. This version has a full replacement for the deprecation reporting capabilities of PhpunitBridge.
* **Change the recipe for PhpunitBridge 7.3 to no longer set-up testing (i.e. configure PHPUnit)**.

  Continueing https://github.com/symfony/recipes/pull/906 from 4 years ago, the only reason you might want to install PhpunitBridge in 7.2 is to use the Clock and DNS mock provided by the new `SymfonyExtension` for PHPUnit (introduced in 7.2).
  Since 6.2, we have a much better solution for ClockMock: the Clock component. DNS mocking is a very niche feature, unlikely to be commonly used in applications.

  On top of this, keeping the PHPUnit configuration in both the bridge and phpunit recipe creates a conflict: One creates a `phpunit.xml.dist` for PHPUnit 9, while the other creates a `phpunit.dist.xml` for PHPUnit 10+. Functionally, this doesn't create an issue as PHPUnit 10+ favors the new file name, but I think we do better to avoid this confusion of double PHPUnit configuration files.

  This recipe change depends on removing PhpunitBridge from the test-pack (https://github.com/symfony/test-pack/pull/24) for Symfony 7.3.

I'm also open to only doing the 2nd change in Symfony 7.4, where I propose to deprecate the PhpunitBridge's deprecation functionality.

cc @alexander-schranz as you've done great work in this effort as well

### DIFF
--- a/phpunit/phpunit/11.1/.env.test
+++ b/phpunit/phpunit/11.1/.env.test
@@ -1,0 +1,3 @@
+# define your env variables for the test env here
+KERNEL_CLASS='App\Kernel'
+APP_SECRET='$ecretf0rt3st'
--- a/phpunit/phpunit/11.1/manifest.json
+++ b/phpunit/phpunit/11.1/manifest.json
@@ -1,0 +1,21 @@
+{
+    "add-lines": [
+        {
+            "file": "phpunit.dist.xml",
+            "content": "            <method>Doctrine\\Deprecations\\Deprecation::trigger</method>\n            <method>Doctrine\\Deprecations\\Deprecation::delegateTriggerToBackend</method>",
+            "position": "after_target",
+            "target": "<deprecationTrigger>",
+            "requires": ["doctrine/deprecations"]
+        }
+    ],
+    "copy-from-recipe": {
+        ".env.test": ".env.test",
+        "phpunit.dist.xml": "phpunit.dist.xml",
+        "tests/": "tests/",
+        "bin/": "%BIN_DIR%/"
+    },
+    "gitignore": [
+        "/phpunit.xml",
+        "/.phpunit.cache/"
+    ]
+}
--- a/phpunit/phpunit/11.1/phpunit.dist.xml
+++ b/phpunit/phpunit/11.1/phpunit.dist.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!-- https://phpunit.readthedocs.io/en/latest/configuration.html -->
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
+         colors="true"
+         failOnDeprecation="true"
+         failOnNotice="true"
+         failOnWarning="true"
+         bootstrap="tests/bootstrap.php"
+         cacheDirectory=".phpunit.cache"
+>
+    <php>
+        <ini name="display_errors" value="1" />
+        <ini name="error_reporting" value="-1" />
+        <server name="APP_ENV" value="test" force="true" />
+        <server name="SHELL_VERBOSITY" value="-1" />
+    </php>
+
+    <testsuites>
+        <testsuite name="Project Test Suite">
+            <directory>tests</directory>
+        </testsuite>
+    </testsuites>
+
+    <source ignoreSuppressionOfDeprecations="true"
+            ignoreIndirectDeprecations="true"
+            restrictNotices="true"
+            restrictWarnings="true"
+    >
+        <include>
+            <directory>src</directory>
+        </include>
+
+        <deprecationTrigger>
+            <function>trigger_deprecation</function>
+        </deprecationTrigger>
+    </source>
+
+    <extensions>
+    </extensions>
+</phpunit>
--- a/doctrine/deprecations/1.0/manifest.json
+++ b/doctrine/deprecations/1.0/manifest.json
@@ -1,0 +1,11 @@
+{
+    "add-lines": [
+        {
+            "file": "phpunit.dist.xml",
+            "content": "            <method>Doctrine\\Deprecations\\Deprecation::trigger</method>\n            <method>Doctrine\\Deprecations\\Deprecation::delegateTriggerToBackend</method>",
+            "position": "after_target",
+            "target": "<deprecationTrigger>",
+            "warn_if_missing": false
+        }
+    ]
+}
--- a/phpunit/phpunit/11.1/bin/phpunit
+++ b/phpunit/phpunit/11.1/bin/phpunit
@@ -1,0 +1,23 @@
+#!/usr/bin/env php
+<?php
+
+if (!ini_get('date.timezone')) {
+    ini_set('date.timezone', 'UTC');
+}
+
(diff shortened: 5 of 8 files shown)


phpunit/phpunit/11.1/phpunit.dist.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!-- https://phpunit.readthedocs.io/en/latest/configuration.html -->
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
+         backupGlobals="false"
```
Spomky, line 6: `backupGlobals="false"`: is this necessary as it is the default value?

phpunit/phpunit/11.1/.env.test
```diff
@@ -0,0 +1,4 @@
+# define your env variables for the test env here
+KERNEL_CLASS='App\Kernel'
+APP_SECRET='$ecretf0rt3st'
+SYMFONY_DEPRECATIONS_HELPER=999999
```
kbond, line 4: This is still required?

Spomky (dismissed):

nicolas-grekas:
There's still https://github.com/sebastianbergmann/phpunit/pull/6165 on the path for best support of deprecation notices. Unfortunately it's going to be shipped with PHPUnit 12, which requires PHP 8.3 minimum.

I'm wondering: does this recipe support PHPUnit 12 also?

.github/workflows/qa.yml
```diff
@@ -9,6 +9,6 @@ defaults:
 
 jobs:
     call-qa:
-        uses: symfony/recipes/.github/workflows/callable-qa.yml@main
+        uses: symfony/recipes/.github/workflows/callable-qa.yml@5b2dcec
```
fabpot, line 12: Is it needed once merged?

wouterj: No, the test endpoint wasn't updated since we use the git-based recipe update. I was fighting the CI to update it, but can't get it working (maybe requires a separate PR).
I'll manually set-up a test endpoint to test the upgrade and remove these things from the PR.

alexander-schranz: @wouterj `pull_request_target` uses the CI of the main branch so changes of the CI files do not effect the CI run itself. Its kind of security feature but little bit anoying if you need to change the CI files: https://stackoverflow.com/questions/74957218/what-is-the-difference-between-pull-request-and-pull-request-target-event-in-git

wouterj:
Hmm, updating recipe in your application are not the best experience ever after this change unfortunately.

### Installing

If you install `phpunit/phpunit` only, this works perfect. However, if you install `symfony/phpunit-bridge` only, you'll no longer get PHPUnit files and, because the recipe for the bridge is compatible with both PHPUnit 9 and 10+, you also don't get a warning that the extension can't be configured.
Should we respectively update older phpunit bridge recipes to include this as a warning in the post-install instructions?

It gets worse if you install `phpunit/phpunit` and `symfony/phpunit-bridge` in an older Symfony application. This will give you both the PHPUnit 10+ and PHPUnit 9 configuration, with the bridge's listener only enabled in PHPUnit 9.
Is this something we can use recipe `conflict`s for?

### Upgrading

From a 7.2 app with the test-pack to 7.3:
Updating the symfony/phpunit-bridge removes all PHPUnit config, test bootstrap files, etc.
You have to manually update, or even add, the phpunit/phpunit requirement to ^12, then upgrade the recipes to get the test files back (without your custom settings).
Is there anything we can do to improve this experience? Would duplicating the test files in both PHPUnit and PhpunitBridge help?

phpunit/phpunit/11.1/phpunit.dist.xml
```diff
@@ -0,0 +1,44 @@
...
+            restrictWarnings="true"
+    >
+        <include>
+            <directory>src</directory>
+        </include>
+
+        <deprecationTrigger>
+            <function>trigger_deprecation</function>
+            <method>Doctrine\Deprecations\Deprecation::trigger</method>
+            <method>Doctrine\Deprecations\Deprecation::delegateTriggerToBackend</method>
```
wouterj, line 38: This results in PHPUnit warnings if `doctrine/deprecations` is not installed:

```
There were 2 PHPUnit test runner warnings:

1) Method Doctrine\Deprecations\Deprecation::trigger cannot be configured as a deprecation trigger because it is not declared

2) Method Doctrine\Deprecations\Deprecation::delegateTriggerToBackend cannot be configured as a deprecation trigger because it is not declared
```

Do we want to extract this to a new `doctrine/deprecations` recipe, or is it OK given 90% of our users have  doctrine installed (and otherwise, it is not a hard failure)?

wouterj:
I think this PR is now in the best state that I can get it.

If users did `composer req test` before Symfony 7.3, they will have `phpunit/phpunit` locked at 9.x in their `composer.json`. They won't get the new phpunit-bridge recipe thanks to the conflict.

The biggest DX problem is still with upgrading any application with `symfony/phpunit-bridge` to 7.3. The update recipe command will delete the `phpunit.xml.dist` file, as it doesn't realize the file ownership has moved to the `phpunit/phpunit` recipe.

The best upgrade way is now to *manually remove* `symfony/phpunit-bridge` from `symfony.lock`, so you don't have to uninstall it. After this, you can savely remove the `symfony/phpunit-bridge` package and upgrade the `phpunit/phpunit` recipe.
Skipping the manual removal, you'll loose customizations in the PHPUnit config/test bootstrap without warning. Is there anything we can do to improve this?

ro0NL:
> The best upgrade way is now to manually remove symfony/phpunit-bridge 

at this point im wondering if deprecating the bridge as a whole would be better

we run latest phpunit/phpunit native without any hassle

phpunit/phpunit/11.1/manifest.json
```diff
@@ -0,0 +1,21 @@
...
+            "position": "after_target",
+            "target": "<deprecationTrigger>",
+            "requires": ["doctrine/deprecations"]
+        }
+    ],
+    "copy-from-recipe": {
+        ".env.test": ".env.test",
+        "phpunit.dist.xml": "phpunit.dist.xml",
+        "tests/": "tests/",
+        "bin/": "bin/"
```
nicolas-grekas, line 15:
```suggestion
        "bin/": "%BIN_DIR%/"
```

nicolas-grekas (commented):
let's do this

nicolas-grekas:
We should tag a new release of https://github.com/symfony/test-pack after this is merged.

nicolas-grekas (approved):
Flex v2.7.1 released, this should provide the best DX we can have today.

fabpot (approved):

PR #1404, opened 2025-04-24
[doctrine/doctrine-bundle] add environment to sqlite DSN

kbond:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

Since sqlite doesn't benefit from `doctrine.dbal.dbname_suffix`, this is required to not wipe your sqlite db when running tests.

### DIFF
--- a/doctrine/doctrine-bundle/1.12/manifest.json
+++ b/doctrine/doctrine-bundle/1.12/manifest.json
@@ -10,7 +10,7 @@
         "#1": "Format described at https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url",
         "#2": "IMPORTANT: You MUST configure your server version, either here or in config/packages/doctrine.yaml",
         "#3": "",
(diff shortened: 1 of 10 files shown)


weaverryan (commented):
Should we only bother updating the latest version? Keep the older version recipes untouched

kbond:
I'd be fine just changing the latest. I noticed all versions could support so I just did a find/replace.

fabpot (approved):

fabpot:
> Should we only bother updating the latest version? Keep the older version recipes untouched

I think this should be considered as a bug fix.

alexander-schranz:
Was confused by this change as it was already done for sqlite. But after reading:

> Since sqlite doesn't benefit from doctrine.dbal.dbname_suffix, this is required to not wipe your sqlite db when running tests.

It totally make sense. Should we add a comment to dbname_suffix that it has no effects on sqlite: https://github.com/symfony/recipes/blob/137e34765a384120bc53d16148910f35e5fc9673/doctrine/doctrine-bundle/2.13/config/packages/doctrine.yaml#L34

stof:
I think this should better be documented in the documentation than in the recipe.

PR #1413, opened 2025-05-13
[symfony/framework-bundle] add `.editorconfig`

kbond:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

Maybe too _meta_ for the recipe but this is something I add to every project for consistency.

### DIFF
--- a/symfony/framework-bundle/6.4/.editorconfig
+++ b/symfony/framework-bundle/6.4/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
--- a/symfony/framework-bundle/6.4/manifest.json
+++ b/symfony/framework-bundle/6.4/manifest.json
@@ -5,7 +5,8 @@
     "copy-from-recipe": {
         "config/": "%CONFIG_DIR%/",
         "public/": "%PUBLIC_DIR%/",
-        "src/": "%SRC_DIR%/"
+        "src/": "%SRC_DIR%/",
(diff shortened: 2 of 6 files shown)


symfony/framework-bundle/6.4/.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.yml]
+indent_size = 2
```
kbond, line 12: FYI, I chose 2 here because of github actions. I find spacing in these files all over the place but most official sources use 2 spaces.

fabpot: We are using 4 spaces (consistently) in Symfony :)

fabpot (changes requested):

chalasr (approved):

symfony/framework-bundle/6.4/.editorconfig
kbond: I've removed but note that even the [Symfony uses 2 spaces in their GHA .yml files](https://github.com/symfony/symfony/blob/7.3/.github/workflows/package-tests.yml).

Maybe?

```
[.github/**.yml]
indent_size = 2
```

fabpot: All Symfony configuration files (everything in this repo) are using 4 spaces.
GHA files are outside of the scope here.

fabpot (approved):

kbond:
My bad, I didn't know about this policy. I applied to just maintained versions

fabpot:
Indeed, let's remove for 6.4 and 7.2.

PR #1416, opened 2025-05-22
Add env variable for `router.default_uri` configuration option

alexander-schranz:
| Q             | A
| ------------- | ---
| License       | MIT

Like database connection, domains are mostly managed OPS team and not by developers. To skip that the developer team need to keep in mind to add a env variable for the default_uri it would be nice if its out of the box a env variable exists – which can always be set by the ops team without have a developer to change the symfony app before.

### DIFF
--- a/symfony/routing/7.0/manifest.json
+++ b/symfony/routing/7.0/manifest.json
@@ -5,5 +5,10 @@
     "aliases": ["router"],
     "conflict": {
         "symfony/framework-bundle": "<7.0"
+    },
+    "env": {
+        "#1": "Configure how to generate URLs in non-HTTP contexts, such as CLI commands.",
+        "#2": "See https://symfony.com/doc/current/routing.html#generating-urls-in-commands",
+        "DEFAULT_URI": "http://localhost"
     }
 }
--- a/symfony/routing/7.0/config/packages/routing.yaml
+++ b/symfony/routing/7.0/config/packages/routing.yaml
@@ -2,7 +2,7 @@ framework:
     router:
         # Configure how to generate URLs in non-HTTP contexts, such as CLI commands.
         # See https://symfony.com/doc/current/routing.html#generating-urls-in-commands
-        #default_uri: http://localhost
+        default_uri: '%env(DEFAULT_URI)%'
 
 when@prod:
     framework:


94noni (approved):

symfony/routing/7.0/manifest.json
```diff
@@ -5,5 +5,10 @@
     "aliases": ["router"],
     "conflict": {
         "symfony/framework-bundle": "<7.0"
+    },
+    "env": {
+        "#1": "Configure how to generate URLs in non-HTTP contexts, such as CLI commands.",
+        "#2": "See https://symfony.com/doc/current/routing.html#generating-urls-in-commands",
+        "DEFAULT_URI": "http://localhost"
```
Kocal, line 12: What about using the `SYMFONY_PROJECT_DEFAULT_ROUTE_URL` env var defined by the Symfony CLI, and fallback to `http://localhost`?
```suggestion
        "DEFAULT_URI": "${SYMFONY_PROJECT_DEFAULT_ROUTE_URL:-http://localhost}"
```

kbond: @Kocal, ah cool, I didn't know about this fallback syntax. I originally tried to use an Symfony env parser but these do not work in `.env`.

alexander-schranz: Created a PR: https://github.com/symfony-cli/symfony-cli/pull/642

alexander-schranz:
@fabpot as we have released: https://github.com/symfony-cli/symfony-cli/releases/tag/v5.13.0 can we merge this?

melroy89:
After this recipe upgrade I get an error:

```
Executing script cache:clear [KO]
 [KO]
Script cache:clear returned with error code 1
!!  
!!   // Clearing the cache for the prod environment with debug false                
!!  
!!  
!!  In EnvVarProcessor.php line 221:
!!                                                    
!!    Environment variable not found: "DEFAULT_URI".  
!!                                                    
!!  
!!  cache:clear [--no-warmup] [--no-optional-warmers]
!!  
!!  
Script @auto-scripts was called via post-install-cmd
```

While there was no recipe introducing any new en var called `DEFAULT_URI`

alexander-schranz:
@melroy89 you need just run `compsoser recipes:update` that adds the new variable to your `.env` file. I did that in @sulu and it did it without any problem: https://github.com/sulu/skeleton/pull/302

alexander-schranz:
@fmonts it doesn't matter both works the same. I used the example provided by the recipe before in `config/packages/routing.yaml`.

PR #1425, opened 2025-05-31
[symfony/phpunit-bridge] Disable deprecations helper on 7.3

HypeMC:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

As mentioned in #1401, the plan is to replace the deprecation reporting capabilities of PhpunitBridge with PHPUnit 11's built-in deprecation feature. Because the `DeprecationErrorHandler` is registered in a [file required by the bridge's `composer.json`](https://github.com/symfony/symfony/blob/158dff8106151b943fad16f85bcf8b70b7bb1178/src/Symfony/Bridge/PhpUnit/composer.json#L32), it is automatically enabled as soon as the package is required.
This PR proposes disabling the helper for Symfony >= 7.3.

@wouterj Any thoughts on this?

### DIFF
--- a/symfony/phpunit-bridge/7.3/manifest.json
+++ b/symfony/phpunit-bridge/7.3/manifest.json
@@ -2,6 +2,13 @@
     "add-lines": [
         {
             "file": "phpunit.dist.xml",
+            "content": "        <server name=\"SYMFONY_DEPRECATIONS_HELPER\" value=\"disabled\" />",
+            "position": "after_target",
+            "target": "<php>",
+            "warn_if_missing": true
+        },
+        {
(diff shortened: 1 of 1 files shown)


wouterj:
Hi! Rather than putting this code in a recipe (and thus userland), I would prefer if we could add a check in the `bootstrap.php` file in the bridge directly. That can be a bugfix on 7.2 imho.

Do you think such a check would be possible?

HypeMC:
@wouterj Please take a look at https://github.com/symfony/symfony/pull/60645

HypeMC:
Closing as https://github.com/symfony/symfony/pull/60645 was merged.

PR #1429, opened 2025-06-10
[symfony/ux-*] add aliases

kbond:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a


Kocal:
Can we add aliases for other UX packages? 😇

kbond:
> Can we add aliases for other UX packages? 😇

I thought about it but wasn't sure on the others
- "map" felt to short
- "translator" - too close to symfony/translation
- "autocomplete" maybe?

Keep in mind, dropping the `symfony/` works: `composer require ux-map`

fabpot (approved):

PR #1430, opened 2025-06-12
Symfony/framework bundle/merge dummy kernel

Nayte91:
| Q             | A
| ------------- | ---
| License       | MIT

Hello,
During SymfonyCon's Keynote, Fabien showed the Symfony Solo initiative, as the final form of the DX refinement. It makes me wonder if it was possible to merge our tiny Kernel.php with our tiny index.php file; It sounds like a bad idea, but given how dummy-ish their code is, It may be a win.

So please find here 2 commits:
- 1 to create the 7.4 folder in framework bundle dir,
- The other one that delete the src/Kernel.php file, and integrate its logic in public/index.php.

So it works on my projects (no idea of all the side effects), and the question still: why?
1. Uncouple the framework with the src/ folder (if you want to rename it whatever/ or create 3 bases it's easier)
2. Somehow it improves the SRP, as now the public.php script really "boot up" Symfony, knowing that mid-term Kernel class was just a bump to achieve it. As it's now a anonymous class, it doesn't break the "1 class per file" rule neither.
3. Less files, more elegant code; You see only 1 less file? I see a 50% reduction of the number of files in src!
4. Hypothetical but I suppose not having this Kernel.php class around the personal codebase will diminish the bad practice to call it wrongly (unintendedly, or worse.. Intendedly)
5. Improve DX by removing this weird sensation as a beginner "is src/ MY folder to code? But why is there a file that is not mine? And with a very intimidating name??". Don't laugh, I had it for few times.

Sorry if it's a bad idea or if there is a lot of side effects!

### DIFF
--- a/symfony/framework-bundle/7.4/public/index.php
+++ b/symfony/framework-bundle/7.4/public/index.php
@@ -1,0 +1,9 @@
+<?php
+
+use App\Kernel;
+
+require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
+
+return function (array $context) {
+    return new Kernel($context['APP_ENV'], (bool) $context['APP_DEBUG']);
+};
--- a/symfony/framework-bundle/7.4/public/index.php
+++ b/symfony/framework-bundle/7.4/public/index.php
@@ -1,9 +1,12 @@
 <?php
 
-use App\Kernel;
+use Symfony\Bundle\FrameworkBundle\Kernel\MicroKernelTrait;
+use Symfony\Component\HttpKernel\Kernel;
 
 require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
 
 return function (array $context) {
-    return new Kernel($context['APP_ENV'], (bool) $context['APP_DEBUG']);
+    return new class($context['APP_ENV'], (bool) $context['APP_DEBUG']) extends Kernel {
+        use MicroKernelTrait;
+    };
 };
--- a/symfony/console/7.4/bin/console
+++ b/symfony/console/7.4/bin/console
@@ -1,8 +1,9 @@
 #!/usr/bin/env php
 <?php
 
-use App\Kernel;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Bundle\FrameworkBundle\Kernel\MicroKernelTrait;
+use Symfony\Component\HttpKernel\Kernel;
 
 if (!is_dir(dirname(__DIR__).'/vendor')) {
     throw new LogicException('Dependencies are missing. Try running "composer install".');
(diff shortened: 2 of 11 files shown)


symfony/framework-bundle/7.4/public/index.php
```diff
@@ -0,0 +1,12 @@
...
+use Symfony\Bundle\FrameworkBundle\Kernel\MicroKernelTrait;
+use Symfony\Component\HttpKernel\Kernel as BaseKernel;
+
+require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
+
+return function (array $context) {
+    return new class($context['APP_ENV'], (bool) $context['APP_DEBUG']) extends BaseKernel {
+        use MicroKernelTrait;
+    };
+};
```
HypeMC, line 12: There are other recipes such as the one for the console component that [use the `Kernel` class](https://github.com/symfony/recipes/blob/d7254400b0564ac0b34cce6dc6f5f1a5b978f8fe/symfony/console/5.3/bin/console#L17-L20), so removing it would things.

.idea/.gitignore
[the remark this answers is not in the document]
Nayte91: Right, sorry! Removed

symfony/framework-bundle/7.4/public/index.php
Nayte91: Absolutely.
I modifed the bin/console file, as it needed also the Kernel so I replaced with another anonymous class. It worked... Until the MicroKernelTrait "Call to a member function setCurrentDir() on false", as \dirname($file) seems to have trouble.

I guess KernelLoader could need the folder of where the user's code is, in this case it could look at config/services.yaml (but maybe it's too early to be able to check a config value)?
If KernelLoader only need where the current kernel booter is, then I need to find a way for a `$file = (new \ReflectionObject($this))->getFileName();` to retrieve the actual file, as it seems to be a pain with a anonymous class?

stof:
duplicating the kernel class between index.php and console means any customization of the kernel logic has to be duplicated. That's a big -1 from me, as it makes maintenance harder in Symfony projects.

Also, reducing the number of files in `src` by 50% is only true for dummy projects that have no other source files. That argument is void for typical Symfony projects (which are the target of those recipes)

Nayte91:
> duplicating the kernel class between index.php and console means any customization of the kernel logic has to be duplicated. That's a big -1 from me, as it makes maintenance harder in Symfony projects.
> 
> Also, reducing the number of files in `src` by 50% is only true for dummy projects that have no other source files. That argument is void for typical Symfony projects (which are the target of those recipes)

That's the kind of feedback I needed; To keep the spirit, do you see any counter-proposition that can bring more value in the public.php or get rid of the dummy Kernel class, from your POV?

nicolas-grekas:
I don't see any direction where this can go.
Having the Kernel in src/ also improves its wiring (by enabling its autoconfiguration by the DI container)
Let me close. The Hello / Solo variants are not something we'd want to make the default. One could of course create their own skeleton to experiment with the concepts.

PR #1435, opened 2025-07-07
[zenstruck/foundry] Add PHPUnit extension

HypeMC:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | https://github.com/zenstruck/foundry/pull/648

Registers Foundry's PHPUnit extension via recipe.

@nikophil Any thoughts on this?

### DIFF
--- a/zenstruck/foundry/2.5/manifest.json
+++ b/zenstruck/foundry/2.5/manifest.json
@@ -7,7 +7,18 @@
     },
     "conflict": {
         "doctrine/persistence": "<2.0",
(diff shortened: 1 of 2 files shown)


nikophil:
Hi @HypeMC 

thanks for this, it is really nice! I was not aware that recipes can be conditioned to the version of a dependency 

At some point I think we will deprecate **not** using the extension, so this is a first step :)

ping @kbond

nikophil (approved):

Spomky (approved):

fabpot (approved):

PR #1438, opened 2025-07-17
Add symfony/ai-bundle

OskarStark:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | --


symfony/ai-bundle/0.1/config/packages/ai.yaml
```diff
@@ -0,0 +1,23 @@
...
+    platform:
+        openai:
+            api_key: '%env(OPENAI_API_KEY)%'
+
+    agent:
+        default:
+            platform: 'symfony_ai.platform.openai'
+            model:
+                name: 'GPT'
+                version: 'gpt-4o-mini'
```
OskarStark, line 19: I think this changed recently, but need to double check if already merged.

cc @welcoMattic

welcoMattic, line 19:
```suggestion
                class: 'Symfony\AI\Platform\Bridge\OpenAI\GPT'
                name: !php/const Symfony\AI\Platform\Bridge\OpenAI\GPT::GPT_4O_MINI
```

welcoMattic: I've commented a suggestion with the correct config

OskarStark: I would prefer to go with the plain string for name here, WDYT @chr-hertel ?

welcoMattic: Hardcoded one can be set and constant commented to indicate to what the static string refers to?

OskarStark: Sounds good to me

OskarStark:
> can we add the alias ai to enable composer require ai? or is that too big fuss for this? 

yes

welcoMattic (dismissed):
LGTM
Just a quick thought about highlighting OpenAI instead of any other platform 🤔

OskarStark:
alias added

chr-hertel:
> Just a quick thought about highlighting OpenAI instead of any other platform 🤔

Yea, i feel you, and somehow brings up the idea of the `PLATFORM_DSN` in my head again ... :thinking:

chr-hertel (approved):

chalasr (approved):

PR #1440, opened 2025-07-26
[Turbo] Correctly fix `framework.csrf_protection.check_header` configuration

Kocal:
| Q             | A
| ------------- | ---
| License       | MIT

Following https://github.com/symfony/recipes/pull/1439#issuecomment-3122053728 

<details>
<summary>Old alternative with `position: 'bottom'`</summary>

The solution is fragile and kind of ugly (note the extra line before `check_header: true`, as it generates the following file:
```yaml
# Enable stateless CSRF protection for forms and logins/logouts
framework:
    form:
        csrf_protection:
            token_id: submit

    csrf_protection:
        stateless_token_ids:
            - submit
            - authenticate
            - logout

        check_header: true
```

But it works: 
```
➜  app-flex-1440 git:(main) ✗ sfc debug:config framework csrf_protection

Current configuration for "framework.csrf_protection"
=====================================================

stateless_token_ids:
    - submit
    - authenticate
    - logout
check_header: true
enabled: null
cookie_name: csrf-token
```

There are other alternatives, like:
1. introducing position `before_target` for `add-lines` modifier but it requires to update Flex code and people to upgrade Flex
2. or drop UX Turbo recipe and add `# check_header: true` (with a comment) to the `csrf.yaml` from Form recipe

</details>

Applied @xabbuh 's suggestion from https://github.com/symfony/recipes/pull/1440#issuecomment-3123382896

### DIFF
--- a/symfony/ux-turbo/2.20/config/packages/ux_turbo.yaml
+++ b/symfony/ux-turbo/2.20/config/packages/ux_turbo.yaml
@@ -1,0 +1,4 @@
+# Enable stateless CSRF protection for forms and logins/logouts
+framework:
+    csrf_protection:
+        check_header: true
--- a/symfony/ux-turbo/2.20/manifest.json
+++ b/symfony/ux-turbo/2.20/manifest.json
@@ -2,17 +2,12 @@
     "bundles": {
         "Symfony\\UX\\Turbo\\TurboBundle": ["all"]
     },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
     "aliases": ["turbo"],
     "conflict": {
         "symfony/framework-bundle": "<7.2",
         "symfony/security-csrf": "<7.2"
-    },
-    "add-lines": [
-        {
-            "file": "config/packages/csrf.yaml",
-            "position": "after_target",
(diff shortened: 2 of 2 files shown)


grelf111:
I get this error while running $symfony new myApp --webpack
In FileLoader.php line 177:                                                            

>   The file "/home/xx/xx/myApp/config/packages/csrf.yaml" does not contain valid YAML: A colon cannot be used in an unquoted mapping value at line 6 (near "    token_id: submit") in /home/xx/xx/myApp/config/packages/csrf.yaml (which is being  imported from "/home/xx/xx/myApp/src/Kernel.php").

Kocal:
What about dropping the UX Turbo recipe, and directly add `# check_header: true`  to the csrf.yaml from Form recipe?

xabbuh:
The UX Turbo recipe could also create a new file with just the following content instead of trying to patch the existing file:

```yaml
# Enable stateless CSRF protection for forms and logins/logouts
framework:
    csrf_protection:
        check_header: true
```

Kocal:
Indeed, didn't think about this but I feel like this is the best alternative we have without touching Flex.

I'm updating the PR

fabpot (approved):

PR #1446, opened 2025-08-20
fix: [LiveComponent] CSRF token not valid after form second submit

markitosgv:
Fix the “invalid CSRF token” error when working with live-components and submitting a form a second time, for example if a server-side validation fails and the form is re-rendered.

| Q             | A
| ------------- | ---
| License       | MIT


smnandre (approved):
Thank you!

fabpot (approved):

PR #1450, opened 2025-09-11
[AssetMapper] Better CSS starter kit

florentdestremau:
When creating a new project, this helps people having a better first view of a symfony page, inspired by simple.css which I regularly use for test projects. It makes a more welcoming page for people trying a rapid app such as using make:entity, make:crud, etc.

I'm not sure where to point this in terms of branch, I would rather update the 6.4 recipe but it's a noticeable change so this might need a full new recipe.


florentdestremau:
It could be a default example yes ! I'll explore

PR #1453, opened 2025-09-24
Add explicit value for csrf_protection.enable when adding stateless_token_ids

unicolored:
Hello,

I had the recipe adding stateless_token_ids. I thought that csrf_protection will be enabled by default. But debug:config showed me that csrf_protection.enabled is null by default.

The current documentation does not explicitly mention that one should add `enable: true` when using the stateless_token_ids: https://symfony.com/doc/current/security/csrf.html#stateless-csrf-tokens

So I thought to add 'enable: true' by default, to be coherent with the comment of the recipe: `# Enable stateless CSRF protection for forms and logins/logouts`.

I hope this is correct.

### DIFF
--- a/symfony/form/7.2/config/packages/csrf.yaml
+++ b/symfony/form/7.2/config/packages/csrf.yaml
@@ -5,6 +5,7 @@ framework:
             token_id: submit
 
     csrf_protection:
+        enabled: true
         stateless_token_ids:
             - submit
             - authenticate


nicolas-grekas:
This shouldn't be needed, this logic should ensure that defining stateless_token_ids is enough:
https://github.com/symfony/symfony/blob/ce5448818410c279a1274f29bb820fc097150262/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L558-L561

Did you experience any issue?

PR #1510, opened 2026-01-16
Fix exit code handling for Symfony 6 compatibility message

outcomer:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

### Summary
This PR fixes the exit code handling in the "Create-project with skeleton ^6" step so that the informational message is correctly displayed when a package doesn't support Symfony 6.

#### Related issues:
- https://github.com/symfony/recipes-contrib/issues/1172#issuecomment-891233725
- https://github.com/symfony/recipes-contrib/issues/1423

#### Issues Fixed
Missing `$` prefix: The condition `[[ EXIT_CODE -eq 2 ]]` was missing the `$` before `EXIT_CODE`, causing the check to always fail

Script exits before capturing exit code: GitHub Actions runs bash with -e flag, so when composer require fails, the script exits immediately before `EXIT_CODE=$?` can capture the exit code. Fixed by using `EXIT_CODE=0` followed by `composer require ... || EXIT_CODE=$?`

Improved message formatting: Added a green-colored border box for better visibility in the logs

Before
```bash
composer require -W --ansi $PACKAGES
EXIT_CODE=$?                          # Never reached if composer fails
if [[ EXIT_CODE -eq 2 ]]; then        # Missing $ prefix
```

After
```bash
EXIT_CODE=0
composer require -W --ansi $PACKAGES || EXIT_CODE=$?
if [[ $EXIT_CODE -eq 2 ]]; then
```

### DIFF
--- a/.github/workflows/callable-qa.yml
+++ b/.github/workflows/callable-qa.yml
@@ -259,11 +259,17 @@ jobs:
                     composer config extra.symfony.allow-contrib ${{ inputs.contrib && 'true' || 'false' }}
                     composer config minimum-stability dev
                     export SYMFONY_ENDPOINT=https://raw.githubusercontent.com/${{ github.repository }}/flex/pull-${{ github.event.number }}/index.json
-                    composer require -W --ansi $PACKAGES
-                    EXIT_CODE=$?
+                    EXIT_CODE=0
+                    composer require -W --ansi $PACKAGES || EXIT_CODE=$?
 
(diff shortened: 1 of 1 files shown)


nicolas-grekas (approved):

outcomer:
@nicolas-grekas Hey. Thanks for your approve. Could we ask someone else to approve the PR too? Cause at least 2 approving reviews are required (((

stof (approved):

PR #1531, opened 2026-04-15
[stimulus-bundle] CSRF: support input outside the form

DavidPetrasek:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | none

Solves the following situation:
``` html
<form id="foo">
    ...
</form>

<input type="hidden" id="foo__token" name="foo[_token]" data-controller="csrf-protection" form="foo" value="csrf-token">
```


symfony/stimulus-bundle/2.24/assets/controllers/csrf_protection_controller.js
```diff
@@ -77,5 +77,17 @@ export function removeCsrfToken (formElement) {
...
+{
+    // Input element is placed inside the form
+    const qSel = 'input[data-controller="csrf-protection"], input[name="_csrf_token"]';
+    const csrfField = formElement.querySelector(qSel);
+    if (csrfField) {return csrfField;}
+
+    // Input element is placed outside the form
+    return Array.from(document.querySelectorAll(qSel))
+        .find((field) => field.form === formElement) || null;
+}
```
nicolas-grekas:
```suggestion
    const qs = 'input[data-controller="csrf-protection"], input[name="_csrf_token"]';

    return formElement.querySelector(qs) || Array.from(document.querySelectorAll(qs)).find((field) => field.form === formElement) || null;
}
```

nicolas-grekas (commented):
can you please squash while applying this suggestion?

PR #1548, opened 2026-07-22
[symfony/security-bundle] Allow the profiler without a trailing slash in the dev firewall

### LINKED ISSUE #1201: The default security configuration requires a trailing forward slash when accessing the profiler
I created a new Symfony 5.4 project (5.4 specifically because I needed to use PHP 7.4), using `symfony new --webapp`, and the security configuration (`config/packages/security.yml`) generated by default has the following firewall:

```yaml
dev:
    pattern: ^/(_(profiler|wdt)|css|images|js)/
    security: false
```

The problem with the pattern is that it doesn't match the URL `localhost:8000/_profiler`, but it _does_ match the URL `localhost:8000/_profiler/`. This means that when accessing the profiler, a trailing forward slash has to be added in at the end, which might not be added by the browser, otherwise you get redirected to the login screen. 

But, if you navigate to `localhost:8000/_profiler/` (with a trailing `/`), the `dev` firewall is used and the profiler can be accessed without authentication.

The way I came upon this problem is because I have a login screen that isn't working and so I needed to debug it. I navigated to `localhost:8000/_profiler`, but it redirected me to the login page, but since the login page wasn't working, I wasn't sure if it was possible to access the profiler, even though I know it should be accessible from an unauthenticated context.

So the problems here are:

1.The default security configuration leads you to think that you have to be authenticated to be able to access the profile, when it can be accessed by an unauthenticated user.
2. The default security configuration forces developers to add the trailing `/` in the URL when accessing the profiler.

The fix for this was to add a `?` at the end of the regular expression to make the trailing `/` optional:

```
pattern: ^/(_(profiler|wdt)|css|images|js)/?
```

I suggest that this is the change that needs to be made, so that it doesn't need to be added by developers whenever they create new projects.

I had a look at this repo for a CONTRIBUTING.md file, but there is none, so I'm not sure if there's a way I can contribute other tha creating an issue. If there, let me know and I'd be happy to make changes and create a PR.

Amoifr:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Closes #1201.

The `dev` firewall pattern ends with a `/`, so a path like `/_profiler` (without the trailing slash) is not matched and falls through to the `main` firewall. With a catch-all `access_control` rule (e.g. `- { path: ^/, roles: IS_AUTHENTICATED_FULLY }`) accessing the profiler then requires authentication.

This applies @stof's suggestion from the issue: replace the trailing `/` with `(/|$)`, so both `/_profiler` and `/_profiler/...` match. The alternation stays anchored, so it does not over-match (`/_profilersection`, `/assetsfoo`, `/buildx` are still excluded, addressing @weaverryan's concern):

| Path | before | after |
| --- | --- | --- |
| `/_profiler` | no | yes |
| `/_profiler/foo` | yes | yes |
| `/_profilersection` | no | no |
| `/assetsfoo` | no | no |

I only touched the `7.4` recipe. The same pattern exists in the `<= 6.4` recipes (`^/(_(profiler|wdt)|css|images|js)/`); happy to apply the same change there if you'd like.

Thanks @aarong416 for the report and @stof for the suggested fix! 🙏

### DIFF
--- a/symfony/security-bundle/6.4/config/packages/security.yaml
+++ b/symfony/security-bundle/6.4/config/packages/security.yaml
@@ -7,7 +7,7 @@ security:
         users_in_memory: { memory: null }
     firewalls:
         dev:
-            pattern: ^/(_(profiler|wdt)|css|images|js)/
+            pattern: ^/(_(profiler|wdt)|css|images|js)(/|$)
             security: false
         main:
             lazy: true
--- a/symfony/security-bundle/7.4/config/packages/security.yaml
+++ b/symfony/security-bundle/7.4/config/packages/security.yaml
@@ -10,7 +10,7 @@ security:
     firewalls:
         dev:
             # Ensure dev tools and static assets are always allowed
-            pattern: ^/(_profiler|_wdt|assets|build)/
+            pattern: ^/(_profiler|_wdt|assets|build)(/|$)
             security: false
         main:
             lazy: true


stof:
I suggest fixing the 6.4 recipe as well (I would leave others untouched as they are for unmaintained versions)

Amoifr:
Done, applied the same `(/|$)` change to the `6.4` recipe and left the older, unmaintained ones untouched. Thanks @stof!

welcoMattic (dismissed):

nicolas-grekas:
This now needs patching the 8.2 file

Amoifr:
Rebased, and the checks job passes now, thanks! Two notes: the `8.2` recipe added in #1569 still had the old pattern, so I applied the same change there and it now matches `7.4`. The remaining `Run updated recipes` failure looks unrelated to this PR: the job requires `symfony/security-bundle:^8.2` in a skeleton pinned to `symfony/symfony 8.1.*`, so the resolution fails before any recipe is applied.

nicolas-grekas (approved):

PR #1550, opened 2026-07-25
[league/oauth2-server-bundle] Add v1.2 recipe

chalasr:
| Q             | A
| ------------- | ---
| License       | MIT

Reducing secret hashing cost in test mode and disabling deprecated things.

### DIFF
--- a/league/oauth2-server-bundle/1.2/config/packages/league_oauth2_server.yaml
+++ b/league/oauth2-server-bundle/1.2/config/packages/league_oauth2_server.yaml
@@ -1,0 +1,30 @@
+league_oauth2_server:
+    authorization_server:
+        private_key: '%env(resolve:OAUTH_PRIVATE_KEY)%'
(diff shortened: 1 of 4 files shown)


ajgarlag:
I think that we should set the required values to prevent any deprecation message:

- `authorization_server.enable_password_grant: false` (See https://github.com/thephpleague/oauth2-server-bundle/issues/259)
- `authorization_server.enable_implicit_grant: false`
- `client.allow_plaintext_secrets: false`

stof:
@chalasr could you please rebase your branch so that we get a working CI run now that the setup has been fixed ? It would simplify the review

chalasr:
Comments addressed and PR rebased. The CI failure is expected : the bundle doesn't have an hard dependency to doctrine-bundle hence it does not require it despite the default config enables the doctrine integration.

nicolas-grekas (approved):

welcoMattic (approved):

PR #1561, opened 2026-08-19
[DoctrineBundle] Add recipe for 3.4 (`auto_discover_entities`)

Nayte91:
New recipe version for `doctrine/doctrine-bundle` 3.4 — **draft until DoctrineBundle 3.4 is released**, since it relies on the `auto_discover_entities` option proposed in doctrine/DoctrineBundle#2277 (itself the companion of symfony/symfony#65437, targeting Symfony 8.2).

Changes compared to the `3.0` recipe:

- `config/packages/doctrine.yaml`: the explicit `orm.mappings` App block (7 lines) is replaced by `auto_discover_entities: true` — entities are discovered from the `#[Entity]` attribute by the container scan, so new projects no longer need to declare where their entities live;
- `manifest.json`: `conflict` on `symfony/framework-bundle < 8.2`, the version that tags the discovered entity classes;
- everything else (env vars, docker-compose, post-install, `src/Entity` + `src/Repository` scaffolding) is carried over unchanged.

The `src/Entity/` directory remains the conventional home for entities, but becomes a convention rather than a requirement: entities in any scanned namespace are discovered.

### DIFF
--- a/doctrine/doctrine-bundle/3.4/config/packages/doctrine.yaml
+++ b/doctrine/doctrine-bundle/3.4/config/packages/doctrine.yaml
@@ -1,0 +1,40 @@
+doctrine:
+    dbal:
+        url: '%env(resolve:DATABASE_URL)%'
+
+        # IMPORTANT: You MUST configure your server version,
+        # either here or in the DATABASE_URL env var (see .env file)
+        #server_version: '16'
+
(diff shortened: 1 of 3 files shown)


Spomky (approved):
Many thanks!
Waiting for 3.4 to be released.

PR #1562, opened 2026-08-20
[Webhook] Add recipe for 8.2 (standard signature format)

lacatoire:
Suggested by @nicolas-grekas in https://github.com/symfony/symfony-docs/pull/22776#issuecomment-5352378012.

`framework.webhook.signature_format` was added in symfony/symfony#62958 (8.2) and
defaults to `legacy` for BC reasons. New applications have no existing subscribers
to migrate, so this recipe opts them into the [Standard Webhooks](https://www.standardwebhooks.com/)
scheme from the start.

The `8.2` directory is a copy of `7.3` plus the new `config/packages/webhook.yaml`,
following the "do it for the next version of the dependency" rule for new best practices.

### DIFF
--- a/symfony/webhook/8.2/config/packages/webhook.yaml
+++ b/symfony/webhook/8.2/config/packages/webhook.yaml
@@ -1,0 +1,5 @@
+framework:
+    webhook:
+        # see https://www.standardwebhooks.com/
+        # use "transitional" to also emit the legacy signature while migrating existing subscribers
+        signature_format: standard
--- a/symfony/webhook/8.2/manifest.json
+++ b/symfony/webhook/8.2/manifest.json
@@ -1,0 +1,5 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    }
+}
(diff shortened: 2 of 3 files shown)


symfony/webhook/8.2/manifest.json
```diff
@@ -0,0 +1,5 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
```
nicolas-grekas, line 3: we need also a conflict rule for fwb < 8.2

symfony/webhook/8.2/config/packages/webhook.yaml
```diff
@@ -0,0 +1,5 @@
+framework:
+    webhook:
+        # see https://www.standardwebhooks.com/
+        # use "transitional" to also emit the legacy signature while migrating existing subscribers
+        signature_format: standard
```
nicolas-grekas, line 5: lets also add `no_private_network: true` from https://github.com/symfony/symfony/pull/65750

nicolas-grekas:
Thanks, I'm continuing this in https://github.com/symfony/recipes/pull/1556

PR #1563, opened 2026-08-27
[FrameworkBundle] Add AGENTS.md for AI coding agents

wachterjohannes:
**Title:** [FrameworkBundle] Add AGENTS.md for AI coding agents

| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | n/a

AI coding agents on a fresh `symfony new` project don't reliably use idiomatic
Symfony patterns unless told (e.g. hand-rolled locking instead of
`symfony/lock`, `json_decode()` instead of Serializer/Validator). This adds a
lightweight `AGENTS.md` to the framework-bundle recipe covering the common
cases: ask before assuming persistence/interface/auth, use Flex instead of
hand-wiring config, prefer the framework's own components over custom code,
plus testing and code-style conventions. A minimal `CLAUDE.md` (`@AGENTS.md`)
is included too, since Claude Code looks for that filename specifically and
this way there's a single source of truth instead of duplicated content.

Note: `8.1` is already released, so per this repo's versioning rule a new
best-practice would normally target a `8.2` folder instead. Targeting `8.1`
directly here is deliberate (purely additive, already aligned within the core
team) — happy to move it to `8.2` if we decide to keep that rule strict.

### DIFF
--- a/symfony/framework-bundle/8.1/AGENTS.md
+++ b/symfony/framework-bundle/8.1/AGENTS.md
@@ -29,12 +29,16 @@ because it isn't installed yet; installing it is one command.
 Follow https://symfony.com/doc/current/best_practices.html to write idiomatic
 Symfony:
 
-- Use PHP attributes for framework metadata: `#[Route]`, `#[Assert\...]`,
-  `#[MapRequestPayload]`, `#[IsGranted]`, `#[AsCommand]`. No YAML or XML routing.
+- Use PHP attributes for framework metadata, and not only on controllers:
+  `#[Route]`, `#[MapRequestPayload]`, `#[IsGranted]` on actions, `#[Assert\...]`
+  on properties, `#[AsCommand]`, `#[AsEventListener]`, `#[AsMessageHandler]`, and
+  `#[AsAlias]` / `#[AsTaggedItem]` / `#[Autoconfigure]` on services. No YAML or
+  XML routing.
 - Rely on autowiring and autoconfiguration. Type-hint constructor arguments and
-  let the container resolve them; write an explicit service definition only when
-  it genuinely can't be resolved (scalar arguments, several implementations of
-  one interface).
+  let the container resolve them. Where a type-hint can't express it, stay in the
+  class with `#[Autowire]` (parameters, env vars, expressions) or `#[Target]` (one
+  of several implementations of an interface). A YAML service definition is the
+  last resort, not the first.
 - Controllers extend `AbstractController`, stay thin, and delegate to services.
 - Use the framework for what it already does: Form for server-rendered forms,
   Validator for validation, Serializer for JSON, Messenger for async work,
--- a/symfony/framework-bundle/8.1/AGENTS.md
+++ b/symfony/framework-bundle/8.1/AGENTS.md
@@ -1,0 +1,76 @@
+# AGENTS.md
+
+This is a Symfony project. Check `composer.json` for the exact Symfony/PHP version
+in use, and read `symfony.lock` to see which recipes ran — don't assume Doctrine,
+Twig, API Platform, Messenger, or Lock are installed unless one of those says so.
+
+## Ask before generating
+
+If the task doesn't specify, ask rather than guess:
+
+- Persistence: Doctrine ORM, Doctrine ODM, or none?
+- Interface: server-rendered (Twig), API (Serializer, maybe API Platform), or both?
+- Auth: SecurityBundle, and which authenticator?
+
+If you can't ask (no interactive channel), state the assumption you're making and
+pick the smallest option (e.g. no persistence layer) rather than scaffolding a
+full stack nobody asked for.
+
+## Adding features: Flex, not hand-wiring
+
+Install new capabilities with `composer require <package>` (e.g. `symfony/lock`,
+`symfony/messenger`, `orm-pack`) and let the Flex recipe register the bundle and
+generate its config. Don't hand-edit `config/bundles.php` or hand-write a bundle's
+base config — that's what the recipe is for. Don't skip a good-fit component just
+because it isn't installed yet; installing it is one command.
+
+## Conventions
+
+- Routing: PHP attributes on controller methods (`#[Route(...)]`), not YAML/XML.
+- Controllers: extend `AbstractController`; use `#[MapRequestPayload]` /
+  `#[MapQueryString]` on action arguments to get Serializer + Validator wired
+  automatically, instead of hand-calling `json_decode()` or `SerializerInterface`.
+  Requires `symfony/serializer` and `symfony/validator` — if the interface is
+  API-only and neither is installed yet, `composer require` them first rather
+  than falling back to manual parsing.
+- Services: rely on autowiring/autoconfiguration; add an explicit service
+  definition only when it genuinely can't be resolved (scalar args, an interface
+  with multiple implementations).
+- Constructor property promotion + `readonly` for DTOs and plain value objects by
+  default. Skip `readonly` on a service if it might end up `lazy: true` — lazy
+  proxies can't extend a `readonly` class.
+- Validate with `symfony/validator` (`#[Assert\...]` + `ValidatorInterface`), not
+  custom validation code.
+- Locking / mutual exclusion: `symfony/lock` (`LockFactory`), not a hand-built
+  mechanism.
+- Long-running or background work: Messenger, not a custom queue.
+- If Doctrine ORM is installed: schema changes go through
+  `doctrine/doctrine-migrations-bundle` (`bin/console make:migration` +
+  `doctrine:migrations:migrate`), not `doctrine:schema:update` or hand-written SQL.
+- Secrets: `bin/console secrets:set` (the vault), read via `%env(...)%`. `.env` is
+  committed to the repo — real secrets go in `.env.local` or the vault, never in
+  `.env` itself.
+- Scaffolding: if `maker-bundle` is installed, prefer `bin/console make:*` with
+  all arguments passed up front and `--no-interaction` where supported — makers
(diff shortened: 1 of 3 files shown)


Spomky:
Great !
This is really complete.
One remark: in the convention section, attributes are only mentioned on controllers. It is worth indicating they are also used for service declaration, injection, commands, properties and so on.

symfony/framework-bundle/8.1/AGENTS.md
```diff
@@ -0,0 +1,76 @@
...
+until it has a test that exercises it the way a caller would — an HTTP request
+for a controller, a service call for a service — not just "it didn't throw."
+
+## Code style
+
+Symfony's coding standard, the `@Symfony` php-cs-fixer ruleset (a PSR-12-derived
+superset). Run `vendor/bin/php-cs-fixer fix` if `friendsofphp/php-cs-fixer` is
+installed — it isn't part of the skeleton by default.
+
+## When training data might be stale
```
javiereguiluz: We could expand this section as follows:

```md
## Discover, don't guess

Framework APIs change between versions and your training data may be stale.
Look things up in the project instead of relying on memory:

- `bin/console about` — versions, environment, paths.
- `bin/console debug:router`, `debug:container`, `debug:autowiring <name>`,
  `debug:config <bundle>`, `config:dump-reference <bundle>` — what exists and
  how it's configured.
- `bin/console lint:container`, `lint:twig templates/`, `lint:yaml config/` —
  validate before running.
- Read the installed source and docblocks in `vendor/`.
- Docs: https://symfony.com/doc/current/ (pick the version that matches
  `composer.json` if the current one doesn't).
```

symfony/framework-bundle/8.1/AGENTS.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Adding features: Flex, not hand-wiring
+
+Install new capabilities with `composer require <package>` (e.g. `symfony/lock`,
+`symfony/messenger`, `orm-pack`) and let the Flex recipe register the bundle and
+generate its config. Don't hand-edit `config/bundles.php` or hand-write a bundle's
+base config — that's what the recipe is for. Don't skip a good-fit component just
+because it isn't installed yet; installing it is one command.
+
+## Conventions
```
javiereguiluz, line 27: Just a proposal. This "Conventions" section looks like two sections: (1) how to write code and (2) how to work in the Symfony app. We could extract the second part to a new smaller section:

```md
## Everyday workflow

- Run the app with `symfony serve -d`; run commands with `symfony console ...`
  (or `bin/console` if the Symfony CLI is not available).
- When something fails, check `var/log/dev.log` and the web profiler
  (`/_profiler`) before changing code.
- If `maker-bundle` is installed, prefer `bin/console make:*` with all
  arguments passed up front and `--no-interaction`; makers prompt on a terminal
  and hang a non-interactive shell. If a maker still needs input, write the
  code by hand.
- If Doctrine ORM is installed, schema changes go through migrations
  (`make:migration` + `doctrine:migrations:migrate`), never
  `doctrine:schema:update` or hand-written SQL.
- `.env` is committed and holds defaults only. Real secrets go in `.env.local`
  (ignored by git) or the secrets vault (`bin/console secrets:set`), and are
  read via `%env(...)%`.
```

javiereguiluz, line 27: I like most of the contents of the "Conventions" section, but some isntructions seem too specific. Just for your consideration, here's a generalization:

```md
## Conventions

Follow https://symfony.com/doc/current/best_practices.html to write idiomatic Symfony:

- Use PHP attributes for framework metadata: `#[Route]`, `#[Assert\...]`,
  `#[MapRequestPayload]`, `#[IsGranted]`, `#[AsCommand]`, etc. No YAML/XML
  routing or annotations.
- Rely on autowiring and autoconfiguration. Type-hint constructor arguments
  and let the container resolve them; write an explicit service definition
  only when it genuinely can't be resolved (scalar arguments, several
  implementations of one interface).
- Controllers extend `AbstractController`, stay thin, and delegate to services.
- Use the framework for what it already does: Form for server-rendered forms,
  Validator for validation, Serializer for JSON, Messenger for async work,
  Security (voters, authenticators) for access control, Twig `path()`/`url()`
  instead of hardcoded URLs.
- Before hand-writing infrastructure (locks, queues, caches, HTTP clients,
  mailers, schedulers...) or adding a third-party library, check whether a
  Symfony component covers it. It usually does.
```

javiereguiluz (commented):
Thanks Johannes!

I like your proposal ... but I added some comments for your consideration.

symfony/framework-bundle/8.1/AGENTS.md
wachterjohannes: Adopted, thanks. This section gained the most: I told the agent to prefer `vendor/` over recollection but gave it nothing to run. One tweak: hedged the linters as "`lint:container`, plus `lint:twig templates/` and `lint:yaml config/` where those packages are installed", since the intro says not to assume Twig or YAML are present.

symfony/framework-bundle/8.1/AGENTS.md
wachterjohannes: Agreed, adopted. You're right that it was two lists. Makers, migrations and secrets are operation, not convention, and splitting them makes "Conventions" short enough to actually be read.

wachterjohannes: Took your framing and bullets, including the best-practices link the file was missing. Your "before hand-writing infrastructure, check whether a component covers it" is strictly better than my enumeration, since it covers what a list always misses.

I kept three specifics though. An `AGENTS.md` is a prompt, not documentation: a model reading "use the framework for what it already does" agrees and then carries on unchanged, because it already believes it is. What changes behaviour is naming the thing it gets wrong. General rules for breadth, a short list of known failure modes for the actual work.

- `#[MapRequestPayload]` over hand-rolled `json_decode()`, including "install Serializer and Validator rather than fall back to manual parsing".
- No `readonly` on a service that may become `lazy: true`: a lazy proxy can't extend a `readonly` class. Non-obvious, and it fails at compile time long after the code looks right.
- `symfony/lock` for mutual exclusion. A hand-built flag reads fine and is wrong under concurrency.

The first and third are the two examples in the PR description, so generalising them away would drop the cases the file was written for. They sit under a heading saying why they're there, so it's clear they're documented failure modes and not an invitation to grow the list. Happy to trim further if it still reads as too specific.

wachterjohannes:
@Spomky good catch, and it went further than the list. My autowiring bullet said to write an explicit service definition for scalar arguments, which is exactly what `#[Autowire]` exists to avoid, so the section was arguing against itself.

Both bullets are now broader: attributes on properties (`#[Assert\...]`), commands, listeners and message handlers, and on services (`#[AsAlias]`, `#[AsTaggedItem]`, `#[Autoconfigure]`); and for injection, `#[Autowire]` for parameters, env vars and expressions, `#[Target]` to pick between implementations, with a YAML definition as the last resort rather than the first.

nicolas-grekas (approved):

welcoMattic:
I'm 50/50 on this one:

👍 , because it leads AI agent to the right direction to code into a fresh new app
👎 , because pushing `symfony serve -d` assume developers use it locally (not always true: docker stack with nginx/php-fpm or frankenphp image, ddev, etc).

Would love to have a way in recipes to ask users for this kind of optional files if they want them or not.

javiereguiluz (approved):

wachterjohannes:
@nicolas-grekas should we adopt this to all the other versions?

Kocal:
Usually we do not modify older versions

alexisLefebvre:
`composer recipes:update symfony/framework-bundle` replace my existing `CLAUDE.md` with a reference to `AGENTS.md`. This is a very opinionated change.

Will users have to revert this every time ` symfony/framework-bundle` get a new recipe version?

wachterjohannes:
@nicolas-grekas question on the behavior above. `recipes:update` overwrites `CLAUDE.md` unconditionally through `copy-from-recipe`. There's no merge, so a project with its own pre-existing `CLAUDE.md` loses it, as @alexisLefebvre found. Is that expected recipe behavior, or should `CLAUDE.md` come out of `copy-from-recipe` and leave only `AGENTS.md`?

alexisLefebvre:
I also have to delete `AGENTS.md` though, since the project use Claude and not another LLM. But it’s a smaller problem than the `CLAUDE.md` overwriting.

What if we had a list of ignored files so that Flex skip them?

```json
    "extra": {
        "symfony": {
            "allow-contrib": false,
            "ignore": [
                "AGENTS.md",
                "CLAUDE.md"
            ],
            "require": "8.1.*",
            "docker": false
        }
    }
```

nicolas-grekas:
People are supposed to use git and review the changes anyway...

alexisLefebvre:
> People are supposed to use git and review the changes anyway...

Fair enough, I agree for PHP or conf files, but I’m not fond of `symfony/framework-bundle` altering my bot configuration.

On one hand I understand the intent of providing general rules, on the other hand I’m unsure if that belongs to the scope of the `symfony/framework-bundle` recipe.

Could these 2 files be provided as a separate package?

wachterjohannes:
On splitting into a separate package: that only reaches people who already know to look for it. `AGENTS.md` is meant to help the ones who don't, someone running `symfony new` for the first time, with an AI agent that has no idea `symfony/lock` exists or that recipes replace hand-wired config. Bundling it with `framework-bundle` means every new project gets that nudge automatically. A separate opt-in package would miss exactly the newcomers it's for.

The overwrite question above is separate and still open.

Kocal:
Laravel uses a separate package for that: [Laravel Boost](https://laravel.com/ai/boost)

PR #1239, opened 2023-09-11
Add Recipe for PHPUnit 10

alexander-schranz:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | symfony/symfony-docs#...

This will add a recipe for PHPUnit 10:

Following changes:

 - `listeners` do not longer exists in PHPUnit 10  where replaced by extensions
 - the old `extensions` still exists but need to be written differently
    - they require `bootstrap` tag now: https://docs.phpunit.de/en/10.3/configuration.html#the-extensions-element and use the new event system https://docs.phpunit.de/en/10.3/extending-phpunit.html#phpunit-s-event-system (this effects maybe phpunit-bridge and panther)
 - move from `phpunit.xml.dist` to `phpunit.dist.xml` for better IDE support (same already done for phpstan) see also: https://github.com/sebastianbergmann/phpunit/issues/4650 / https://github.com/sebastianbergmann/phpunit/commit/07a022ad0548823b04c8fab073a7bff2fbcf9c8c
 - from cache file `.phpunit.result.cache` to cache directory `/.phpunit.cache/`
 - move unrelated Panther ENV vars to Panther recipe

PR #1566, opened 2026-08-29
[symfony/panther] Set PANTHER_APP_ENV to an environment the Symfony 8 kernel allows

### LINKED ISSUE #1565: [panther] PANTHER_APP_ENV=panther is rejected by the Symfony 8 kernel
**Disclaimer:** I do not use Panther myself and cannot validate the details below from experience. This report was produced by an AI agent (Claude Code) that reproduced each step locally. I am filing it because it surfaced while fixing an unrelated CI failure in symfony/maker-bundle (https://github.com/symfony/maker-bundle/pull/1811), and it looked like it affects real users rather than just that test suite. Please correct anything that is wrong.

### Summary

Two recipes in this repository conflict on Symfony 8. Installing `symfony/panther` into a Symfony 8 skeleton app produces a Panther test that cannot boot the application under test.

* `symfony/panther` (recipe 1.0, `eac1868`) writes into `.env.test`:

  ```
  ###> symfony/panther ###
  PANTHER_APP_ENV=panther
  PANTHER_ERROR_SCREENSHOT_DIR=./var/error-screenshots
  PANTHER_DEVTOOLS=0
  ###< symfony/panther ###
  ```

* `symfony/framework-bundle` (recipe 8.1, `312027a`) generates `src/Kernel.php` containing:

  ```php
  private function getAllowedEnvs(): array
  {
      return ['prod', 'dev', 'test'];
  }
  ```

Panther's `WebServerManager` starts the built-in web server with `APP_ENV` taken from `PANTHER_APP_ENV`, so the server process boots the kernel as `panther`. `Symfony\Component\DependencyInjection\Kernel\KernelTrait` then throws:

```
The environment "panther" is not registered as allowed by "App\Kernel::getAllowedEnvs()".
```

The browser receives Symfony's error page instead of the application, so any assertion against page content fails with something like `Failed asserting that 'Symfony Exception' contains "Hello World"`.

### Reproduction

```console
composer create-project symfony/skeleton:8.* app && cd app
composer require --dev symfony/panther phpunit browser-kit symfony/css-selector
composer require --dev dbrekelmans/bdi && php vendor/dbrekelmans/bdi/bdi detect drivers
```

Add a route returning `<h1>Hello World</h1>`, then a test:

```php
class SmokeTest extends PantherTestCase
{
    public function testHomepage(): void
    {
        $client = static::createPantherClient();
        $client->request('GET', '/');
        $this->assertSelectorTextContains('h1', 'Hello World');
    }
}
```

`vendor/bin/phpunit` fails. Symfony 7.4 is unaffected, since its skeleton kernel has no `getAllowedEnvs()`.

### Workaround

One line in the user's own `src/Kernel.php`: add `'panther'` to `getAllowedEnvs()`. Returning `[]` also works, because the check only applies when the list is non-empty.

### Possible fixes

1. **Set `PANTHER_APP_ENV=test` in the Panther recipe.** One line, fixes every new installation. The cost is losing the dedicated `panther` environment the recipe deliberately introduces. Note that Panther itself does not require the variable: when it is unset, `WebServerManager` passes no env and the server inherits `APP_ENV=test` from the test run.
2. **Document it** in the Panther docs for Symfony 8, and accept that users patch their kernel.

### Wider point

This is not specific to Panther. Any recipe that introduces its own `APP_ENV` value hits the same wall on Symfony 8, because the environment allow-list lives in a file only the user can edit. It may be worth deciding on a convention for that case.

Amoifr:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | -

Fixes #1565.

`symfony/panther` writes `PANTHER_APP_ENV=panther` into `.env.test`, and the `symfony/framework-bundle` 8.1 recipe generates a kernel whose `getAllowedEnvs()` returns `['prod', 'dev', 'test']`. Panther starts the web server with `APP_ENV` taken from `PANTHER_APP_ENV`, so on Symfony 8 the server boots into an environment the kernel rejects, and every Panther test asserts against Symfony's error page instead of the application.

This is the first of the two options the issue lists. I checked each claim against the sources rather than taking the report as read:

- the value is in `symfony/panther/1.0/manifest.json`, and the allow-list in `symfony/framework-bundle/8.1/src/Kernel.php`;
- `KernelTrait::getKernelParameters()` throws `The environment "%s" is not registered as allowed by "%s::getAllowedEnvs()"`, and only when the list is non-empty, which is why returning `[]` is also a workaround;
- `WebServerManager::__construct()` sets `$env['APP_ENV']` from `PANTHER_APP_ENV` when it is present.

Two things that were not in the issue and that make the trade-off smaller than it looks:

- **The `panther` environment configures nothing.** No recipe ships a `config/packages/panther/`, and Panther's own documentation never suggests that value; its CHANGELOG describes `PANTHER_APP_ENV` only as a way to override `APP_ENV`. The value came in through #1239 (a PHPUnit 10 recipe change) and was carried over unchanged when the `dotenv` configurator replaced the `add-lines` entry.
- **`symfony/panther` is the only recipe in this repository that sets a non-standard `APP_ENV` value.** Every other occurrence is the `APP_ENV=dev` of `symfony/framework-bundle`, which the allow-list covers. So the wider convention question raised at the end of the issue currently has exactly one case, this one.

I kept the variable rather than removing it, even though Panther works without it: when it is unset the web server inherits its environment from the test process, which depends on how the application propagates `APP_ENV`, whereas setting it explicitly does not.

Happy to close this if you would rather document the kernel change instead, which is the other option in the issue.

### DIFF
--- a/symfony/panther/1.0/manifest.json
+++ b/symfony/panther/1.0/manifest.json
@@ -20,7 +20,7 @@
     ],
     "dotenv": {
         "test": {
-            "PANTHER_APP_ENV": "panther",
+            "PANTHER_APP_ENV": "test",
             "PANTHER_ERROR_SCREENSHOT_DIR": "./var/error-screenshots",
             "PANTHER_DEVTOOLS": "0"
         }


Amoifr:
The red `Run updated recipes` job is not caused by this change, and no manifest content could make it green.

The job installs the recipe's package from the directory name, so it runs `composer req 'symfony/panther:^1.0'`, and that fails before any recipe is applied:

```
- Root composer.json requires symfony/panther ^1.0 -> satisfiable by symfony/panther[v1.0.0, ..., 1.x-dev].
- symfony/panther[v1.0.0, ..., 1.x-dev] require symfony/browser-kit ^4.4 || ^5.0 -> found symfony/browser-kit[...] but these were not loaded
```

Panther 1.x caps `symfony/browser-kit` at `^5.0`, while the job pins Symfony to `6.4.*` and `7.4.*`. The current major is 2.x (v2.4.0, January 2026) and requires `symfony/browser-kit ^6.4 || ^7.3 || ^8.0`, so the resolution only works there.

The recipe itself is fine: Flex treats the directory as a minimum version, so people installing Panther 2.x do get this recipe. It is only the CI harness that picks the 1.x line and cannot resolve it.

Making that job green would mean a `symfony/panther/2.0` directory, which changes which versions receive the recipe, so it felt like a separate decision rather than something to smuggle into a one-line fix. Happy to open it if you want it.

nicolas-grekas (approved):

welcoMattic (approved):

PR #1569, opened 2026-09-02
[SecurityBundle] Import the OIDC login routes

welcoMattic:
| Q             | A
| ------------- | ---
| License       | MIT
| Doc issue/PR  | symfony/symfony-docs#22881

Symfony 8.2 adds an `oidc_login` firewall authenticator (symfony/symfony#64954). The OIDC provider redirects the user back to a callback path, and that path needs a route: without one the router answers the redirect with a 404 before the firewall ever sees it. Since symfony/symfony#65814, the same loader also declares a start route per firewall, which begins the flow by redirecting to the provider, so a login page can link to it.

Symfony declares these routes through a route loader the application imports, exactly as it does for the logout routes, so this adds the import next to `_security_logout`:

```yaml
# config/routes/security.yaml
_security_oidc_login:
    resource: security.authenticator.oidc_login.route_loader
    type: service
```

The loader is registered unconditionally since symfony/symfony#65800, and declares no route until a firewall uses the OIDC authenticator, so the import is safe in applications that do not use OIDC.

New `8.2` directory, copied from `7.4` with only that block added.

### DIFF
--- a/symfony/security-bundle/8.2/config/packages/security.yaml
+++ b/symfony/security-bundle/8.2/config/packages/security.yaml
@@ -1,0 +1,39 @@
+security:
+    # https://symfony.com/doc/current/security.html#registering-the-user-hashing-passwords
+    password_hashers:
+        Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface: 'auto'
+
+    # https://symfony.com/doc/current/security.html#loading-the-user-the-user-provider
+    providers:
+        users_in_memory: { memory: null }
+
+    firewalls:
+        dev:
+            # Ensure dev tools and static assets are always allowed
+            pattern: ^/(_profiler|_wdt|assets|build)/
+            security: false
+        main:
+            lazy: true
+            provider: users_in_memory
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
Thanks Mathieu, this closes the last wiring gap of the 8.2 OIDC feature. I amended one thing while verifying it (ddc8bc2): the import is now named `_security_oidc_login` instead of `_oidc_login_callbacks`, since symfony/symfony#65814 made the loader declare the start routes too, so the name no longer described what it imports; it also matches the `_security_logout` convention next to it and the name the SecurityBundle test fixtures use. The `8.2` directory is a faithful copy of `7.4` otherwise (manifest and packages config byte-identical), and the import shape is the one the bundle's functional tests exercise. Title and description updated to match.

nicolas-grekas (approved):
Verified against the merged 8.2 loader: faithful 7.4 copy-forward, import shape matches the bundle fixtures.

chalasr (approved):

